PR #1, opened 2024-11-08
Initial encodings

aldenks:
Initial data variable encodings which maintain a close to faithful representation but get compression ratio to 0.27.


aldenks:
Distributions of data variables -- used to determine add_offset, categorical encodings, etc
variable-distributions

Left column: uncompressed; right column: compressed
compression

PR #2, opened 2024-11-18
download only specific bands from source gribs

aldenks:
* Download source gribs band-by-band
* Use rasterio/gdal instead of cfgrib as grib decoder
* Parallelize downloads and file reads
* New structure for data variable encoding and attributes
* Support full forecast horizon


noaa/gefs/forecast/read_data.py
```diff
@@ -27,66 +53,195 @@ def download_file(
...
+    path: os.PathLike[str],
+    data_var: DataVar,
+) -> None:
+    grib_element = data_var.internal_attrs.grib_element
+    if data_var.internal_attrs.include_lead_time_suffix:
+        lead_hours = coords["lead_time"].total_seconds() / (60 * 60)
+        if lead_hours % 6 == 0:
+            grib_element += "06"
+        else:
+            grib_element += "03"
```
aldenks, line 202: lets put this within some sort of afirmative elif check for 03 and have the else raise an AssertionError or similar so we fail in any future cases we aren't thinking about

PR #3, opened 2024-12-04
Add additional data variables

margocrawf:
Add additional variables from `pgrb2sp25`

TODOs before merging:
- [ ] Pare down variables to the most relevant ones for the initial release
- [ ] Update encodings for the new variables
- [ ] Validate the step types and add new ones to the `Literal`


aldenks (approved):
We may want to pare down variables further, but the rest of this is good to go so merging!

PR #4, opened 2024-12-18
Set up ability to deploy to AWS cluster

laurengulland:
Done:
* eks cluster exists

In progress:
* cluster permissioning
* instructions on how other folks can get set up and connected to the cluster

To do:
* actually get our stuff running in aws and test compute/egress costs
* add to & clean up setup instructions, they're very messy at the moment.


aldenks:
Thank you! These are great notes for us in our particular deployment but we may not want to merge them into main as the idea is this code could be deployed by others on their infra too at some point

laurengulland:
Great point Alden, thank you! 

We now have internal documentation about getting set up, and all this is reflected there.

PR #5, opened 2024-12-18
Add expected_forecast_length and ingested_forecast_length coords along init_time dimension

margocrawf:
* `expected_forecast_length` is filled in based on the time of day-- 35 day for 0z forecasts and 10 day for 6z/12z/18z forecasts
* `ingested_forecast_length` is used to indicate when an operational forecast is partially ingested (For 0z, the first 10 days of lead_time are released before the other 25 days)


noaa/gefs/forecast/template.py
```diff
@@ -39,10 +47,24 @@ def get_template(init_time_end: DatetimeLike) -> xr.Dataset:
...
+    # Fill in expected_forecast_lengths based on init_time
+    template_expected_forecast_length = ds["expected_forecast_length"]
+    ds.coords["expected_forecast_length"] = (
+        "init_time",
+        [
+            INIT_TIME_HOURS_TO_FORECAST_LENGTHS[pd.Timestamp(init_time).hour]
+            for init_time in ds.init_time.values
+        ],
+    )
+    ds["expected_forecast_length"].encoding = template_expected_forecast_length.encoding
```
aldenks, line 59: We don't have any attrs yet, but lets copy them over too from the template in case we add them later

noaa/gefs/forecast/template.py
```diff
@@ -25,6 +26,13 @@
 
 TEMPLATE_PATH = "noaa/gefs/forecast/templates/latest.zarr"
 
+INIT_TIME_HOURS_TO_FORECAST_LENGTHS = {
+    0: np.timedelta64(840, "h"),
+    6: np.timedelta64(384, "h"),
+    12: np.timedelta64(384, "h"),
+    18: np.timedelta64(384, "h"),
```
aldenks: We're a little inconsistent between xarray's numpy datetime/timedelta and pandas. We can't change xarray, but pandas UX is nicer..  Specific comment for here is if this would work as pd.Timedelta lets do that instead.

noaa/gefs/forecast/template.py
```diff
@@ -39,10 +47,24 @@ def get_template(init_time_end: DatetimeLike) -> xr.Dataset:
...
+        ],
+    )
+    ds["expected_forecast_length"].encoding = template_expected_forecast_length.encoding
+
     # Coordinates which are dask arrays are not written with
     # to_zarr(store, compute=False) so we ensure all coordinates are loaded.
     for coordinate in ds.coords.values():
-        assert isinstance(coordinate.data, np.ndarray)
+        coordinate.load()
+        assert isinstance(
```
aldenks, line 65: I believe this is guaranteed if we've just called .load() but little harm in leaving it

aldenks:
Approved!

noaa/gefs/forecast/template.py
margocrawf: I changed these lines and ran `update-template`, and there was no diff in the template metadata, so it seems like we can use pandas :)

PR #6, opened 2025-01-07
Operational writes

margocrawf:
Adds a new command (`reformat-operational`) that fetches and saves new data from the end of the current dataset to the present.
This can be run in a cronjob to allow continuous updating of the dataset.

### DIFF
--- a/noaa/gefs/forecast/cli.py
+++ b/noaa/gefs/forecast/cli.py
@@ -36,8 +36,8 @@ def reformat_chunks(
 
 
 @app.command()
-def reformat_operational() -> None:
-    reformat.reformat_operational()
+def reformat_operational_update() -> None:
+    reformat.reformat_operational_update()
 
 
 if __name__ == "__main__":
--- a/noaa/gefs/forecast/cli.py
+++ b/noaa/gefs/forecast/cli.py
@@ -35,5 +35,10 @@ def reformat_chunks(
     )
 
 
+@app.command()
+def reformat_operational() -> None:
+    reformat.reformat_operational()
+
+
 if __name__ == "__main__":
     app()
--- a/noaa/gefs/forecast/reformat.py
+++ b/noaa/gefs/forecast/reformat.py
@@ -1,6 +1,7 @@
 import concurrent.futures
 import os
 import re
+import shutil
 import subprocess
 from collections import defaultdict, deque
 from collections.abc import Iterable
@@ -29,6 +30,57 @@ from noaa.gefs.forecast.read_data import (
 )
 
 _PROCESSING_CHUNK_DIMENSION = "init_time"
+
+
+def reformat_operational() -> None:
+    final_store = get_store()
+    tmp_store = get_local_tmp_store()
+    # Get the dataset, check what data is already present
+    ds = xr.open_dataset(final_store)
+    # TODO base this not only on presence of init_time,
+    # but also forecast_length/expected_forecast_length
+    # coordinates:
+    # https://app.asana.com/0/1207641655330818/1201550874443598/f
+    start = ds.init_time.max()
+    init_time_end = "2024-09-01T18:00"  # pd.Timestamp.utcnow().tz_localize(None)
+    template_ds = template.get_template(init_time_end)
+
+    new_init_times = template_ds.init_time.loc[template_ds.init_time > start]
+    new_init_time_indices = template_ds.get_index("init_time").get_indexer(
+        new_init_times
+    )  # type: ignore
+    i_slices = list(
+        starmap(
+            slice,
+            zip(
+                new_init_time_indices,
+                list(new_init_time_indices[1:]) + [None],
+                strict=False,
+            ),
+        )
+    )
+    template.write_metadata(template_ds, tmp_store, get_mode(final_store))
+    # TODO: Maybe operational doesn't need to fan out? We expect it to only digest
+    # a small amount of data.
+    reformat_init_time_i_slices(i_slices, template_ds, tmp_store)
+
+    for data_var in ds.data_vars:
+        source_path = tmp_store / str(data_var)
+        # Only the new files will exist for tmp_store.
+        files_to_copy = os.listdir(source_path)
+        # TODO this only works locally
+        dest_path = final_store / str(data_var)
+        [shutil.copy2(source_path / file, dest_path / file) for file in files_to_copy]
+
+    template.write_metadata(template_ds, final_store, get_mode(final_store))
+
+    breakpoint()
+
+    # add the new chunks
+
+    # add the metadata
+
+    breakpoint()
 
 
 def reformat_local(init_time_end: DatetimeLike) -> None:
@@ -123,7 +175,12 @@ def reformat_chunks(
     )
 
     print(f"This is {worker_index = }, {workers_total = }, {worker_init_time_i_slices}")
+    reformat_init_time_i_slices(worker_init_time_i_slices, template_ds, store)
 
+
+def reformat_init_time_i_slices(
+    init_time_i_slices: list[slice], template_ds: xr.Dataset, store: StoreLike
+) -> None:
     data_var_groups = group_data_vars_by_noaa_file_type(
         [d for d in template.DATA_VARIABLES if d.name in template_ds]
     )
@@ -137,7 +194,7 @@ def reformat_chunks(
     # # but make sure to read thread safety comment in our `read_data` function.
     # proccess_executor = ProcessPoolExecutor(max_workers=os.cpu_count())
 
-    for init_time_i_slice in worker_init_time_i_slices:
+    for init_time_i_slice in init_time_i_slices:
         chunk_template_ds = template_ds.isel(init_time=init_time_i_slice)
 
         chunk_init_times = pd.to_datetime(chunk_template_ds["init_time"].values)
@@ -280,6 +337,10 @@ def get_store() -> StoreLike:
         "s3://us-west-2.opendata.source.coop/aldenks/noaa-gefs-dev/forecast/dev.zarr"
     )
     return store
+
+
+def get_local_tmp_store() -> Path:
+    return Path("data/output/noaa/gefs/forecast/tmp.zarr").absolute()
 
 
 def get_mode(store: StoreLike) -> Literal["w-", "w"]:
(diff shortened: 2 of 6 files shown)


noaa/gefs/forecast/cli.py
```diff
@@ -35,5 +35,10 @@ def reformat_chunks(
     )
 
 
+@app.command()
+def reformat_operational() -> None:
```
aldenks: reformat_operational_update

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
 _PROCESSING_CHUNK_DIMENSION = "init_time"
 
 
+def reformat_operational() -> None:
+    final_store = get_store()
+    tmp_store = get_local_tmp_store()
+    # Get the dataset, check what data is already present
+    ds = xr.open_dataset(final_store, engine="zarr")  # type: ignore
```
aldenks: prefer xr.open_zarr

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+    new_init_time_indices = template_ds.get_index("init_time").get_indexer(
+        new_init_times
+    )  # type: ignore
+    new_init_time_i_slices = list(
+        starmap(
+            slice,
+            zip(
+                new_init_time_indices,
+                list(new_init_time_indices[1:]) + [None],
+                strict=False,
```
aldenks, line 66: can use itertools.pairwise

also I think there's a ruff check that will remind us this we should make sure is enabled

noaa/gefs/forecast/reformat.py
aldenks: assert chunk sizes match any assumptions in code below (eg init time chunk = 1)

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+                for fn in copy_data_var(data_var, tmp_store, final_store)
+            ]
+        )
+
+    concurrent.futures.wait(futures, return_when="ALL_COMPLETED")
+
+    template.write_metadata(template_ds, final_store, get_mode(final_store))
+
+
+def get_recent_init_times_for_reprocessing(ds: xr.Dataset) -> list[np.datetime64]:
```
aldenks: return type looks like an np.array (Array1D[np.datetime64] from our common.types perhaps?)

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+        ds.init_time > max_init_time - np.timedelta64(7, "D"), drop=True
+    )
+    # Get the recent init_times where we have only partially completed
+    # the ingest.
+    recent_incomplete_init_times = recent_init_times.where(
+        recent_init_times.ingested_forecast_length
+        < recent_init_times.expected_forecast_length
+    )
+    if len(recent_incomplete_init_times) > 0:
+        return recent_incomplete_init_times.min().values  # type: ignore
```
aldenks: remove .min() based on your suggestion as we talked live

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+
+    return max_init_time  # type: ignore
```
aldenks: return empty array in this case

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+
+
+def copy_data_var(
+    data_var: DataVar, tmp_store: Path, final_store: fsspec.FSMap
+) -> list[Callable]:
+    source_path = tmp_store / str(data_var.name)
+    # Only the new files will exist for tmp_store.
+    files_to_copy = [source_path / blob_name for blob_name in os.listdir(source_path)]
+    return [
```
aldenks: can return just the Callable vs a 1 elem list

noaa/gefs/forecast/reformat.py
```diff
@@ -123,7 +224,17 @@ def reformat_chunks(
     )
 
     print(f"This is {worker_index = }, {workers_total = }, {worker_init_time_i_slices}")
+    list(reformat_init_time_i_slices(worker_init_time_i_slices, template_ds, store))
```
aldenks: can use `consume` vs `list` to show intent

noaa/gefs/forecast/reformat.py
```diff
@@ -194,7 +313,13 @@ def reformat_chunks(
                         var_coords_and_paths = coords_and_paths
                     data_array = xr.full_like(chunk_template_ds[data_var.name], np.nan)
                     # valid_time is a coordinate and already written with different chunks
-                    data_array = data_array.drop_vars("valid_time")
+                    data_array = data_array.drop_vars(
+                        [
+                            "valid_time",
+                            "ingested_forecast_length",
+                            "expected_forecast_length",
```
aldenks: so we don't forget to expand this list when we add more coords, lets invert and drop all coords that are not dimension coords

noaa/gefs/forecast/reformat.py
```diff
@@ -210,6 +335,7 @@ def reformat_chunks(
                     print(f"Writing {data_var.name} {chunk_init_times_str}")
                     chunks = template.chunk_args(chunk_template_ds)
                     data_array.chunk(chunks).to_zarr(store, region="auto")
+                    yield (data_var, chunk_init_times[0], max_ingested_lead_time)
```
aldenks: return full chunk_init_times and move indexing into caller where we've asserted 1 init time per chunk

noaa/gefs/forecast/reformat.py
```diff
@@ -282,6 +411,10 @@ def get_store() -> StoreLike:
     return store
 
 
+def get_local_tmp_store() -> Path:
+    return Path("data/output/noaa/gefs/forecast/tmp.zarr").absolute()
```
aldenks: use a random tmp dir to avoid confusing issues testing locally

noaa/gefs/forecast/template.py
```diff
@@ -136,6 +136,7 @@ def write_metadata(
     mode: Literal["w", "w-", "a", "a-", "r+", "r"],
 ) -> None:
     template_ds.to_zarr(store, mode=mode, compute=False)
+    # TODO: this doesn't print anything useful if the store is an FSMap
```
aldenks: ok to remove this to keep our TODO garden spare

uv.lock
```diff
@@ -511,11 +511,11 @@ wheels = [
 
 [[package]]
 name = "fsspec"
```
aldenks, line 513: maybe update xarray + zarr (or everything)

noaa/gefs/forecast/reformat.py
aldenks: oh and lets put it in `data/tmp/<uid>/tmp.zarr` that way it will use any disk space mounted on `data/`

aldenks: aaannd in kubet job yaml, change the mount location from `data/downloads/` to just `data/`

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +32,106 @@
...
+        < recent_init_times.expected_forecast_length
+    )
+    if len(recent_incomplete_init_times) > 0:
+        return recent_incomplete_init_times.min().values  # type: ignore
+
+    return max_init_time  # type: ignore
+
+
+def copy_data_var(
+    data_var: DataVar, tmp_store: Path, final_store: fsspec.FSMap
```
aldenks: we want to update this function to take in the init chunk index and only copy those files

aldenks (changes requested):
WAAAHOOOO!

noaa/gefs/forecast/reformat.py
aldenks: I think you can just do `list(starmap(slice, pairwise(new_init_time_indices)))`

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +33,109 @@
...
+        starmap(
+            slice,
+            zip(
+                recent_incomplete_init_times_indices,
+                [x + 1 for x in recent_incomplete_init_times_indices],
+                strict=False,
+            ),
+        )
+    )
+    # TODO: should this be on the same ThreadPoolExecutor as the downloads?
```
```suggestion
```

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +33,109 @@
...
+    for data_var, max_lead_times in reformat_init_time_i_slices(
+        recent_incomplete_init_time_i_slices + new_init_time_i_slices,
+        template_ds,
+        tmp_store,
+    ):
+        for init_time, max_lead_time in max_lead_times.items():
+            template_ds["ingested_forecast_length"].loc[{"init_time": init_time}] = (
+                max_lead_time
+            )
+            # TODO: this only works because we know that chunks are size 1 in the
```
```suggestion
            This only works because we know that chunks are size 1 in the
```

noaa/gefs/forecast/reformat.py
```diff
@@ -31,6 +33,109 @@
...
+    recent_init_times = ds.init_time.where(
+        ds.init_time > max_init_time - np.timedelta64(7, "D"), drop=True
+    )
+    # Get the recent init_times where we have only partially completed
+    # the ingest.
+    recent_incomplete_init_times = recent_init_times.where(
+        recent_init_times.ingested_forecast_length
+        < recent_init_times.expected_forecast_length
+    ).values
+    return recent_incomplete_init_times[~np.isnat(recent_incomplete_init_times)]  # type: ignore
```
```suggestion
    return recent_incomplete_init_times = recent_init_times.where(
        recent_init_times.ingested_forecast_length
        < recent_init_times.expected_forecast_length
    ).dropna().values  # type: ignore
```

noaa/gefs/forecast/reformat.py
```diff
@@ -210,6 +355,7 @@ def reformat_chunks(
                     print(f"Writing {data_var.name} {chunk_init_times_str}")
                     chunks = template.chunk_args(chunk_template_ds)
                     data_array.chunk(chunks).to_zarr(store, region="auto")
+                    yield (data_var, max_lead_times)
```
aldenks, line 354: Nice, I like yielding the data_var

aldenks (approved):
Approved! lets get testin this in prod and we can refine :)

noaa/gefs/forecast/reformat.py
margocrawf: This ends up not including the final element, since it doesn't include `None`.

uv.lock
margocrawf: I may do this as a separate PR

PR #9, opened 2025-01-13
Configuration for v0.0.1 ingest

aldenks:
- select subset of variables
- start dataset on 2025-01-01
- update template with above changes
- update store location to dynamical org on source coop and call it v0.0.1.zarr


noaa/gefs/forecast/reformat.py
```diff
@@ -201,8 +201,8 @@ def reformat_kubernetes(
             "WORKERS_TOTAL": workers_total,
             "PARALLELISM": parallelism,
             "CPU": 16,
-            "MEMORY": "80G",
-            "EPHEMERAL_STORAGE": "60G",
+            "MEMORY": "110G",
+            "EPHEMERAL_STORAGE": "120G",
```
aldenks: Im not sure either of these additional resources were needed, i just wanted it to work first time

noaa/gefs/forecast/templates/latest.zarr/.zmetadata
```diff
@@ -37,21 +37,21 @@
...
+        "categorical_freezing_rain_surface/.zattrs": {
             "_ARRAY_DIMENSIONS": [
                 "init_time",
                 "ensemble_member",
                 "lead_time",
                 "latitude",
                 "longitude"
             ],
-            "coordinates": "expected_forecast_length ingested_forecast_length valid_time",
+            "coordinates": "valid_time",
```
aldenks: I think we still want the `expected_forecast_length ingested_forecast_length` coordinates, look into this

noaa/gefs/forecast/reformat.py
aldenks: I undid this change

margocrawf (approved):

PR #10, opened 2025-01-14
Operationally update via cronjob

margocrawf:
To apply:
```
$ set IMAGE <your-docker-image>
$ sed -e "s!{{IMAGE}}!$IMAGE!g" deploy/operational_update.yaml | kubectl apply -f -
```

### DIFF
--- a/deploy/operational_update.yaml
+++ b/deploy/operational_update.yaml
@@ -1,5 +1,3 @@
-
-# Once this is validated we can turn this into a cron job
 apiVersion: batch/v1
 kind: CronJob
 metadata:
--- a/deploy/operational_update.yaml
+++ b/deploy/operational_update.yaml
@@ -1,0 +1,63 @@
+
+# Once this is validated we can turn this into a cron job
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: reformat-operational-ingest
+spec:
+  schedule: "0 7 * * *" # At 7:00 every day.
+  jobTemplate:
+    spec:
+      backoffLimit: 1 
+      ttlSecondsAfterFinished: 720 # 12 hours
+      template:
+        spec:
+          restartPolicy: Never
+          containers:
+          - name: worker
+            image: {{IMAGE}}
+            command:
+            - python
+            - main.py
+            - "noaa-gefs-forecast"
+            - reformat-operational-update
+            env:
+            - name: DYNAMICAL_ENV
+              value: prod
+            - name: AWS_ACCESS_KEY_ID
+              valueFrom:
+                secretKeyRef:
+                  name: source-coop-key
+                  key: AWS_ACCESS_KEY_ID
+            - name: AWS_SECRET_ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: source-coop-key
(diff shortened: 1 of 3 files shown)


deploy/operational_update.yaml
```diff
@@ -0,0 +1,63 @@
+
+# Once this is validated we can turn this into a cron job
```
laurengulland:
```suggestion
```

deploy/operational_update.yaml
```diff
@@ -0,0 +1,63 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: reformat-operational-ingest
+spec:
+  schedule: "0 7 * * *" # At 7:00 every day.
+  jobTemplate:
+    spec:
+      backoffLimit: 1
```
laurengulland, line 9: does this make sense in tandem with the `restartPolicy: Never` below? maybe we should delete this line or set it to 0? (also if we do keep it but set it to 0, we should move the restart policy & this to be in the same level of the template?)

deploy/operational_update.yaml
```diff
@@ -0,0 +1,63 @@
...
+            ephemeral:
+              volumeClaimTemplate:
+                metadata:
+                  labels:
+                    type: ephemeral
+                spec:
+                  accessModes: ["ReadWriteOnce"]
+                  resources:
+                    requests:
+                      storage: "500G"
```
laurengulland, line 56: question for my understanding, not blocking review: what is this volume/why is it useful? is it just a big volume where we're directing the data to be downloaded into, but it's useful to mount it separately because...?

noaa/gefs/forecast/reformat.py
```diff
@@ -352,6 +360,10 @@ def reformat_init_time_i_slices(
                     chunks = template.chunk_args(chunk_template_ds)
                     data_array.chunk(chunks).to_zarr(store, region="auto")
                     yield (data_var, max_lead_times)
+                # Reclaim space once done.
+                for coord_and_path in coords_and_paths:
+                    if coord_and_path[1] is not None:
+                        coord_and_path[1].unlink()
```
laurengulland: nit for readability:
```suggestion
                # Reclaim space once done.
                for coord, filepath in coords_and_paths:
                    if filepath is not None:
                        filepath.unlink()
```

laurengulland (commented):
some small suggestions but looks good! 

Did you say there was an issue with the ingested_forecast_length coord getting updated during the ingest, but the matching data not getting updated correctly? Were you intending to add a fix for that in this PR or does that make more sense in a follow-up?

deploy/operational_update.yaml
margocrawf: They can make sense together, because `restartPolicy` and `backoffLimit` are configuring retries at [different levels](https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures). 
I am still trying to decide how many restarts is reasonable.

laurengulland: Thanks for the doc, totally makes sense! :)

deploy/operational_update.yaml
laurengulland: chatted about this live, resolved!

PR #12, opened 2025-01-23
Ensure coordinates are saved in a single chunk

aldenks:
- Consolidate how derived coordinates are created
- When deriving coordinates, make sure encoding and attributes are properly set (this is what fixes the bug)
- Along the way fix a race condition in our updating zarr: if chunk size is `-1` then the array size will change on each update, meaning there will be a moment when the actual array size and metadata array size for the coordinate array differ. This is only an issue for coordinates because with our data var init_time chunk size of 1 we create a whole new chunk each update. Instead of -1, just fix it to a big number of init times (15 years worth).

Evidence the fix works: only one array for all coordinates after writing a new zarr with reformat-local.
```
(coord-chunking-fix)> ls data/output/noaa/gefs/forecast/dev.zarr/*/**
data/output/noaa/gefs/forecast/dev.zarr/ensemble_member/0
data/output/noaa/gefs/forecast/dev.zarr/expected_forecast_length/0
data/output/noaa/gefs/forecast/dev.zarr/ingested_forecast_length/0
data/output/noaa/gefs/forecast/dev.zarr/init_time/0
data/output/noaa/gefs/forecast/dev.zarr/latitude/0
data/output/noaa/gefs/forecast/dev.zarr/lead_time/0
data/output/noaa/gefs/forecast/dev.zarr/longitude/0
data/output/noaa/gefs/forecast/dev.zarr/valid_time/0.0
```


noaa/gefs/forecast/template.py
```diff
@@ -41,45 +35,15 @@ def get_template(init_time_end: DatetimeLike) -> xr.Dataset:
...
-    # Coordinates which are dask arrays are not written with
-    # to_zarr(store, compute=False) so we ensure all coordinates are loaded.
+    # Recompute derived coordinates after reindex
+    ds = add_derived_coordinates(ds)
+
+    # Coordinates which are dask arrays are not written with .to_zarr(store, compute=False)
+    # We want to write all coords when writing metadata, so ensure they are loaded as numpy arrays.
     for coordinate in ds.coords.values():
         coordinate.load()
-        assert isinstance(coordinate.data, np.ndarray), (
```
aldenks, line 78: `.load()` guarantees this will be true

noaa/gefs/forecast/template.py
```diff
@@ -120,41 +83,67 @@ def update_template() -> None:
...
 
     write_metadata(ds, template_path, mode="w")
 
 
+def add_derived_coordinates(ds: xr.Dataset, copy_metadata: bool = True) -> xr.Dataset:
+    new_coords = {
+        "ingested_forecast_length": (
+            "init_time",
+            np.full(ds["init_time"].size, np.timedelta64("NaT", "ns")),
+        ),
```
aldenks, line 107: Biggest question to answer before merging: @margocrawf will calling this from get_template() overwrite the true ingested forecast length data?

noaa/gefs/forecast/template.py
```diff
@@ -41,45 +35,15 @@ def get_template(init_time_end: DatetimeLike) -> xr.Dataset:
...
-    template_expected_forecast_length = ds["expected_forecast_length"]
-    ds.coords["expected_forecast_length"] = (
-        "init_time",
-        [
-            INIT_TIME_HOURS_TO_FORECAST_LENGTHS[pd.Timestamp(init_time).hour]
-            for init_time in ds.init_time.values
-        ],
-    )
-    ds["expected_forecast_length"].encoding = template_expected_forecast_length.encoding
-    ds["expected_forecast_length"].attrs = template_expected_forecast_length.attrs
```
aldenks, line 72: Was it a bug on main that expected_forecast_length is updated twice? was the second one intended to be ingested_forecast_length?

noaa/gefs/forecast/template.py
aldenks: oh no, looks like it'll work great. I see how the template_ds.ingested_forecast_length is filled in from the actual datasetse's values at the top of reformat_operational_update

noaa/gefs/forecast/template.py
```diff
@@ -120,41 +83,67 @@ def update_template() -> None:
...
 
+def add_derived_coordinates(ds: xr.Dataset, copy_metadata: bool = True) -> xr.Dataset:
+    new_coords = {
+        "ingested_forecast_length": (
+            "init_time",
+            np.full(ds["init_time"].size, np.timedelta64("NaT", "ns")),
+        ),
+        "expected_forecast_length": (
+            "init_time",
+            np.where(
```
margocrawf, line 110: Swapping to this inline version removes a check that the `init_time` is in `(0,6,12,18)`. Previously it would fail if you received an unexpected key in the dict.
I'd prefer to keep the check in some capacity, I think it's useful to make sure we don't introduce odd init_times.

aldenks: Thanks! done https://github.com/dynamical-org/reformatters/commit/e91cb881bad8c171379b7a3862649907bc5fb701

PR #14, opened 2025-02-03
Make ingessted_forecast_length coord along ensemble member and init time

margocrawf:
Different ensemble members fill the later lead times with different delays.

### DIFF
--- a/noaa/gefs/forecast/reformat.py
+++ b/noaa/gefs/forecast/reformat.py
@@ -96,10 +96,11 @@ def reformat_operational_update() -> None:
         template_ds,
         tmp_store,
     ):
-        for init_time, max_lead_time in max_lead_times.items():
-            template_ds["ingested_forecast_length"].loc[{"init_time": init_time}] = (
-                max_lead_time
-            )
+        for (init_time, ensemble_member), max_lead_time in max_lead_times.items():
+            if np.issubdtype(type(ensemble_member), np.integer):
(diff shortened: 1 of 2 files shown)


noaa/gefs/forecast/reformat.py
```diff
@@ -342,16 +351,24 @@ def reformat_init_time_i_slices(
...
 
-                max_lead_times: dict[pd.Timestamp, pd.Timedelta] = {}
-                for init_time, init_time_coords_and_paths in groupby(
-                    coords_and_paths, key=lambda v: v[0]["init_time"]
+                def groupbykey(v: tuple[SourceFileCoords, Path | None]):
+                    init_time_portion = v[0]["init_time"]
+                    if "ensemble_member" in v[0]:
+                        ensemble_portion = v[0]["ensemble_member"]
+                    else:
+                        ensemble_portion = v[0]["statistic"]
```
samn, line 359: Fine to leave, but consider using an `else` or  exhaustive `match`   expression here to ensure this remains correct even if additional types of `SourceFileCoords` are added:

```suggestion
from typing import assert_never
# ...
                    match v[0]:
                      case {'ensemble_member': ensemble_member}:
                        ensemble_portion = ensemble_member
                      case {'statistic': statistic}:
                        ensemble_portion = statistic
                      case _ as unreachable:
                        assert_never(unreachable)                    
```

samn (approved):

PR #15, opened 2025-02-03
Fixes to operational update with new avg variables and EKS cluster

margocrawf:


noaa/gefs/forecast/reformat.py
```diff
@@ -111,7 +118,7 @@ def reformat_operational_update() -> None:
 
     concurrent.futures.wait(futures, return_when="ALL_COMPLETED")
 
-    template.write_metadata(template_ds, tmp_store, get_mode(final_store))
+    template.write_metadata(template_ds, tmp_store, "w")
```
margocrawf: For the tmp store, I think we should always have write permission.

noaa/gefs/forecast/reformat.py
```diff
@@ -53,7 +53,14 @@ def reformat_operational_update() -> None:
...
     # If this changes we will need to refactor.
-    assert all(size == 1 for size in ds.chunksizes[_PROCESSING_CHUNK_DIMENSION])
+    ensemble_vars = [var.name for var in template_config._DATA_VARIABLES]
+    stat_vars = [var.name for var in template_config._STATISTIC_DATA_VARIABLES]
+    assert all(
+        size == 1 for size in ds[ensemble_vars].chunksizes[_PROCESSING_CHUNK_DIMENSION]
+    )
+    assert all(
+        size == 1 for size in ds[stat_vars].chunksizes[_PROCESSING_CHUNK_DIMENSION]
+    )
```
```suggestion
    assert all(1 == da.chunksizes[_PROCESSING_CHUNK_DIMENSION] for da in ds.data_vars.values())
```

noaa/gefs/forecast/reformat.py
aldenks: what about updating get_mode to also check if the store name matches the tmp store (like maybe make the tmp store format `.../<uuid>-tmp.zarr` and do `store_root.endswith("dev.zarr") or store_root.endswith("-tmp.zarr"))` that way all mode logic stays in that function

noaa/gefs/forecast/reformat.py
```diff
@@ -139,7 +146,7 @@ def copy_data_var(
     tmp_store: Path,
     final_store: fsspec.FSMap,
 ) -> Callable[[], None]:
-    files_to_copy = list(tmp_store.glob(f"{data_var.name}/{chunk_index}.*.*.*.*"))
+    files_to_copy = list(tmp_store.glob(f"{data_var.name}/{chunk_index}.*.*.*"))
```
```suggestion
    files_to_copy = list(tmp_store.glob(f"{data_var.name}/{chunk_index}.*.*.*"))  # matches any chunk with 4 or more dimensions
```

aldenks (approved):

PR #16, opened 2025-02-03
Optimize reindex by making empty copy

aldenks:


noaa/gefs/forecast/reformat.py
```diff
@@ -390,9 +391,14 @@ def reformat_init_time_i_slices(
...
                     print(f"Writing {data_var.name} {chunk_init_times_str}")
-                    chunks = template.chunk_args(chunk_template_ds[data_var.name])
-                    data_array.chunk(chunks).to_zarr(store, region="auto")
+                    (
+                        data_array.chunk(
+                            data_array.encoding["preferred_chunks"]
+                        ).to_zarr(store, region="auto")
+                    )
                     yield (data_var, max_lead_times)
+                    del data_array
```
aldenks, line 400: this should help peak memory usage by allowing it to free data_array from one iteration before allocating another in the next iteration

PR #17, opened 2025-02-04
Fix mypy warning: `NewGenericSyntax is already enabled by default`

samn:


aldenks:
Is vscode happy with this? The vscode mypy extension used to be a little out of date and so without this we were getting errors in vscode

samn:
As far as I can tell it seems to be working with the mypy plugin, didn't see any errors in any file

aldenks:
great! approved and ill just merge to save ya the time

aldenks:
looks like it hadn't actually run, it is still unhappy so reverting this for now.

PR #22, opened 2025-02-07
Upload one init time at a time

margocrawf:
Rather than waiting until all new init times are processed, upload them one at a time.


aldenks:
Looks great! also nice that if this fails after one init is done but there's another to update the next run doesn't have to start at the beginning

PR #25, opened 2025-02-11
Add a command to validate that GEFS zarrs are well formed

samn:
Right now it only checks that there aren't too many nan values, but it could be extended to run additional checks in the future.

It took ~10 minutes to check 10 sample coordinates locally, but it should be faster when run in prod closer to the data.

### DIFF
--- a/noaa/gefs/forecast/reformat.py
+++ b/noaa/gefs/forecast/reformat.py
@@ -24,7 +24,7 @@ import xarray as xr
 from common import validation
 from common.config import Config  # noqa:F401
 from common.download_directory import cd_into_download_directory
-from common.kubernetes import ReformatCronJob, ReformatJob
+from common.kubernetes import ReformatCronJob, ReformatJob, ValidationCronJob
 from common.types import Array1D, DatetimeLike, StoreLike
 from noaa.gefs.forecast import template, template_config
 from noaa.gefs.forecast.config_models import DataVar, EnsembleStatistic
@@ -38,7 +38,7 @@ from noaa.gefs.forecast.read_data import (
 
 _PROCESSING_CHUNK_DIMENSION = "init_time"
 _CRON_SCHEDULE = "0 7 * * *"  # At 7:00 UTC every day.
-
+_VALIDATION_CRON_SCHEDULE = "0 10 * * *"  # At 10:00 UTC every day.
 logging.basicConfig()
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
@@ -341,23 +341,33 @@ def deploy_operational_updates() -> None:
 
     dataset_id = template_config.DATASET_ID
 
-    cron_job = ReformatCronJob(
+    operational_update_cron_job = ReformatCronJob(
         name=f"{dataset_id}-operational-update",
         schedule=_CRON_SCHEDULE,
         image=image_tag,
         dataset_id=dataset_id,
-        workers_total=1,
-        parallelism=1,
         cpu="6",  # fit on 8 vCPU node
         memory="60G",  # fit on 64GB node
         ephemeral_storage="150G",
-        command=[
-            "reformat-operational-update",
-        ],
+    )
+    validation_cron_job = ValidationCronJob(
+        name=f"{dataset_id}-validation",
+        schedule=_VALIDATION_CRON_SCHEDULE,
+        image=image_tag,
+        dataset_id=dataset_id,
+        cpu="6",  # fit on 8 vCPU node
+        memory="60G",  # fit on 64GB node
+        ephemeral_storage="10G",
+    )
+
+    k8s_resources_str = (
+        json.dumps(operational_update_cron_job.as_kubernetes_object())
+        + "\n-----\n"
+        + json.dumps(validation_cron_job.as_kubernetes_object())
     )
     subprocess.run(  # noqa: S603
         ["/usr/bin/kubectl", "apply", "-f", "-"],
-        input=json.dumps(cron_job.as_kubernetes_object()),
+        input=k8s_resources_str,
         text=True,
         check=True,
     )
@@ -651,7 +661,7 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
 def check_recent_nans(
     ds: xr.Dataset, n_samples: int = 100, max_nan_percentage: float = 70
 ) -> validation.ValidationResult:
-    """Check for NaN values in the most recent day of data. Fails if more than `max_nan_percentage` of sampled data is NaN."""
+    """Check for NaN values in the most recent day of data. Fails if more than 70% of sampled data is NaN."""
     # Get most recent init time
     latest_init = ds.init_time.max()
 
@@ -672,7 +682,7 @@ def check_recent_nans(
 
     problem_vars = []
     for var_name, da in sample_ds.data_vars.items():
-        nan_percentage = (np.isnan(da.values).sum() / da.size) * 100
+        nan_percentage = da.isnull().mean().compute() * 100
         if nan_percentage > max_nan_percentage:
             problem_vars.append((var_name, nan_percentage))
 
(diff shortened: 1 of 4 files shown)


noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
         deque(iterator, maxlen=0)
     else:
         next(islice(iterator, n, n), None)
+
+
+def check_recent_nans(
+    ds: xr.Dataset, n_samples: int = 100, max_nan_percentage: float = 70
+) -> validation.ValidationResult:
+    """Check for NaN values in the most recent day of data. Fails if more than 70% of sampled data is NaN."""
```
margocrawf: The last day's data is probably going to be mostly NaNs past the first 10 days in the `lead_time` dimension, except for a handful of ensemble members which get filled in quicker.
I wonder if a more tailored approach would make more sense here.
1) Change this to `sel` down to 10 days and require a lower `max_nan_percentage` for the `latest_init`
2) Check that the newest `init_time` that's more than 3 days old has a low % of nans in all lead times.

noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
...
+    for dim, size in dims.items():
+        sample_idx = np.random.choice(size, min(n_samples, size), replace=False)
+        sample_coords[dim] = sample_idx
+
+    # First select all coordinates and load the data once
+    sample_ds = ds.sel(init_time=latest_init).isel(sample_coords)
+
+    problem_vars = []
+    for var_name, da in sample_ds.data_vars.items():
+        nan_percentage = (np.isnan(da.values).sum() / da.size) * 100
```
aldenks: you could also do `da.isnull().mean().compute() * 100` to stay in xarray land. no strong pref

noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
...
     else:
         next(islice(iterator, n, n), None)
+
+
+def check_recent_nans(
+    ds: xr.Dataset, n_samples: int = 100, max_nan_percentage: float = 70
+) -> validation.ValidationResult:
+    """Check for NaN values in the most recent day of data. Fails if more than 70% of sampled data is NaN."""
+    # Get most recent init time
+    latest_init = ds.init_time.max()
```
aldenks: This could not alert us if the update just stops happening since the max init will still exist, just be old. I think we want to do something like `.sel(init_time=slice(now - 24h, None))` and then make sure we check for the case that size == 0

noaa/gefs/forecast/reformat.py
```diff
@@ -346,18 +347,28 @@ def deploy_operational_updates() -> None:
...
+        schedule=_VALIDATION_CRON_SCHEDULE,
+        image=image_tag,
+        dataset_id=dataset_id,
+        cpu="6",  # fit on 8 vCPU node
+        memory="60G",  # fit on 64GB node
+        ephemeral_storage="10G",
+    )
+
+    k8s_resources_str = (
+        json.dumps(cron_job.as_kubernetes_object())
```
margocrawf: nit: rename this to `operational_update_cron_job`

noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
         deque(iterator, maxlen=0)
     else:
         next(islice(iterator, n, n), None)
+
+
+def check_recent_nans(
+    ds: xr.Dataset, n_samples: int = 100, max_nan_percentage: float = 70
```
aldenks: 100 samples probably means you're selecting every chunk but also double/tripple selecting a number of chunks. I think we could just load all locations in the lat/lon domain, probably faster. [Or we could do a much smaller sample, like n==5 (there's 20 chunks in lon dim, and 10 in lat)]

aldenks (changes requested):

noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
...
+    monitor_slug=f"{template_config.DATASET_ID}-validation",
+    monitor_config={
+        "schedule": {"type": "crontab", "value": _VALIDATION_CRON_SCHEDULE},
+        "timezone": "UTC",
+        # If an expected check-in doesn't come in `checkin_margin`
+        # minutes, it'll be considered missed
+        "checkin_margin": 10,
+        # The check-in is allowed to run for `max_runtime` minutes
+        # before it's considered failed
+        "max_runtime": int(timedelta(hours=12).total_seconds()),
```
margocrawf: The comment says `max_runtime` is minutes, but the code looks like it's in seconds. Which is it? 
Also, 12 hours seems like a long time for this validation to run.

noaa/gefs/forecast/reformat.py
```diff
@@ -646,3 +657,67 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
...
+    if problem_vars:
+        message = "Excessive NaN values found:\n"
+        for var, pct in problem_vars:
+            message += f"- {var}: {pct:.1f}% NaN\n"
+        message += f"For init_time {latest_init} in sampled locations"
+        return validation.ValidationResult(passed=False, message=message)
+
+    return validation.ValidationResult(
+        passed=True,
+        message="All variables have acceptable NaN percentages (<70%) in sampled locations of latest data",
```
margocrawf:
```suggestion
        message=f"All variables have acceptable NaN percentages (<{max_nan_percentage}%) in sampled locations of latest data",
```
nit

aldenks:
THANK YOU

noaa/gefs/forecast/reformat.py
```diff
@@ -46,20 +48,8 @@
...
-        "checkin_margin": 10,
-        # The check-in is allowed to run for `max_runtime` minutes
-        # before it's considered failed
-        "max_runtime": 120,  # minutes
-        # It'll take `failure_issue_threshold` consecutive failed
-        # check-ins to create an issue
-        "failure_issue_threshold": 1,
-        # It'll take `recovery_threshold` OK check-ins to resolve
-        # an issue
-        "recovery_threshold": 1,
```
samn, line 62: We can configure the alert parameters in Sentry directly, but I think it's good to have the cron schedule line up with the `CronJob` automatically

noaa/gefs/forecast/reformat.py
```diff
@@ -36,8 +38,8 @@
 )
 
 _PROCESSING_CHUNK_DIMENSION = "init_time"
-_CRON_SCHEDULE = "0 7 * * *"  # At 7:00 UTC every day.
-
+_OPERATIONAL_CRON_SCHEDULE = "0 7 * * *"  # At 7:00 UTC every day.
+_VALIDATION_CRON_SCHEDULE = "0 10 * * *"  # At 10:00 UTC every day.
```
```suggestion
_VALIDATION_CRON_SCHEDULE = "0 9 * * *"  # At 9:00 UTC every day.
```
They definitely finish in under two hours (i think its about 75 mins now)

noaa/gefs/forecast/reformat.py
```diff
@@ -645,3 +645,71 @@ def consume[T](iterator: Iterable[T], n: int | None = None) -> None:
...
+    """Check for NaN values in the most recent day of data. Fails if more than 70% of sampled data is NaN."""
+
+    now = pd.Timestamp.now()
+    # We want to show that the latest init time has valid data going out up to 10 days (we may not have forecasts
+    # past that, depending on the ensemble member and init time). To avoid needing to load a rediculous amount of data
+    # we'll choose a random lead_time within that range.
+    lead_time_day = np.random.randint(0, 10)  # [0, 10) since we add 1 below
+    sample_ds = ds.sel(
+        init_time=slice(now - timedelta(days=1), None),
+        lead_time=slice(
```
aldenks, line 677: All lead times are in a single chunk so this actually doesn't speed anything up :) 

We can speed up checks by subsetting in the spatial domain (there's 20 x 10 chunks in lon x lat) but I don't think that's actually necessary

aldenks (approved):

aldenks:
Approved after comments!

noaa/gefs/forecast/reformat.py
samn: Oh hmm, I tried to subset `lead_time` because I was running out of memory loading them all, but I guess we can run this on a vm with more memory than my workstation (32gb)

samn: The validation pod is getting OOMKilled with 60gb of memory and only 4 days of lead time, gonna ship this as is and refine later (maybe it's worth running multiple validation pods to check a wider range)

PR #26, opened 2025-02-11
Run mypy on pre-commit

samn:


.pre-commit-config.yaml
```diff
@@ -7,3 +7,13 @@ repos:
       - id: ruff
       # Run the formatter.
       - id: ruff-format
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: "uv run mypy"
```
aldenks, line 14: doh! nice -- the other ways they offer use a different version of mypy than your environment which i didn't love

aldenks (approved):

PR #27, opened 2025-02-11
Run code quality checks for PRs using GitHub Actions

samn:


samn:
MyPy is failing, will be fixed with https://github.com/dynamical-org/reformatters/pull/26

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Install uv
+      uses: astral-sh/setup-uv@v5
+      with:
+        enable-cache: true
+        cache-dependency-glob: "uv.lock"
+
+    - name: "Set up Python"
```
aldenks, line 21: I think you can have setup-uv do this too https://github.com/astral-sh/setup-uv?tab=readme-ov-file#python-version and that would keep everything in uv

samn: i thought that using `actions/setup-python` might be faster since GitHub will keep it cached https://github.com/astral-sh/setup-uv?tab=readme-ov-file#do-i-still-need-actionssetup-python-alongside-setup-uv

aldenks (approved):

PR #28, opened 2025-02-13
Deploy all operational jobs from CI

samn:


common/deploy.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Protocol
+
+from common import docker, kubernetes
+from noaa.gefs.forecast.reformat import (
+    operational_kubernetes_resources as noaa_gefs_forecast_operational_kubernetes_resources,
+)
+
+
+class OperationalKubernetesResources(Protocol):
+    def __call__(self, image_tag: str) -> list[kubernetes.ReformatJob]: ...
```
aldenks: Should we rename ReformatJob -> Job or K8sJob since they are validation jobs too? Return type could be `Iterable` since we don't actually care its a list.  no need to change in this PR, not biggies

aldenks (approved):
This is sweeeet! Approved

PR #32, opened 2025-02-17
Add tmin & tmax variables to noaa-gefs-forecast

samn:


aldenks (approved):
Let's disable operational updates for this before merging so that those don't add these variables to the current operational zarr and leave missing chunks in the past. We can re-enable after we reprocess the archive

aldenks:
But otherwise, Approved!

samn:
Added https://github.com/dynamical-org/reformatters/pull/32/commits/7bc7ea64f15a0e842ae8b2a09d1494fa6eb84b21 to suspend the `CronJobs` while we work on reprocessing the archive

PR #36, opened 2025-02-17
Move some common pieces around in anticipation of new datasets

samn:


samn:
Paired with @aldenks 

There's some more stuff to move out of reformat.py in a followup

PR #37, opened 2025-02-18
Zarr v3

aldenks:
- Update all dependencies to latest
- Write and update v3 format zarrs and v3 zarr template
- Updates to store and filesystem handling for new zarr Store interface


noaa/gefs/forecast/reformat.py
```diff
@@ -165,13 +169,18 @@ def reformat_operational_update() -> None:
...
 
-        concurrent.futures.wait(futures, return_when="ALL_COMPLETED")
+        concurrent.futures.wait(data_var_upload_futures, return_when="FIRST_EXCEPTION")
+        for future in data_var_upload_futures:
+            if (e := future.exception()) is not None:
+                raise e
 
+        # Write metadata again to update the ingested_forecast_length
         template.write_metadata(
-            truncated_template_ds.isel(init_time=slice(0, init_time_i_slice.stop)),
```
aldenks, line 171: this was superfluous, truncated_template_ds is the result of calling the same isel above

common/zarr.py
```diff
@@ -1,99 +1,147 @@
...
+        #     "file://data/output/noaa/gefs/forecast/dev.zarr"
+        # )
+        # Instead make a zarr LocalStore and attach an fsspec filesystem to it.
+        # Technically that filesystem should be an AsyncFileSystem to match
+        # zarr.storage.FsspecStore but AsyncFileSystem does not support _put.
+        local_path = Path(f"data/output/{dataset_id}/v{version}.zarr").absolute()
+
+        store = zarr.storage.LocalStore(local_path)
+
+        from fsspec.implementations.local import LocalFileSystem  # type: ignore
```
samn: move this up with other imports

common/zarr.py
```diff
@@ -1,99 +1,147 @@
...
-        "s3://us-west-2.opendata.source.coop/dynamical/noaa-gefs-forecast/v0.1.0.zarr"
+    if Config.is_dev:
+        # This should work, but it gives FileNotFoundError when it should be creating a new zarr.
+        # return zarr.storage.FsspecStore.from_url(
+        #     "file://data/output/noaa/gefs/forecast/dev.zarr"
+        # )
+        # Instead make a zarr LocalStore and attach an fsspec filesystem to it.
+        # Technically that filesystem should be an AsyncFileSystem to match
+        # zarr.storage.FsspecStore but AsyncFileSystem does not support _put.
+        local_path = Path(f"data/output/{dataset_id}/v{version}.zarr").absolute()
```
samn, line 31: this used to be `dev.zarr`, not `vdev.zarr`

noaa/gefs/forecast/template.py
```diff
@@ -133,7 +149,7 @@ def add_derived_coordinates(ds: xr.Dataset, copy_metadata: bool = True) -> xr.Da
 
 
 def assign_var_metadata(
-    var: xr.DataArray, var_config: GEFSDataVar | Coordinate
+    var: xr.DataArray, var_config: DataVar[Any] | Coordinate
```
samn, line 152: Why switch to `DataVar[Any]` instead of the more specific type, `GEFSDataVar`?

samn (approved):

noaa/gefs/forecast/template.py
aldenks: To set us up for factoring template stuff out to common, sorry should have been in a different PR

common/zarr.py
aldenks: yeah, i thought for 2 mins about getting fancy re prepending a v based on the version value but decided against it. it also changes the local path from having slashes to dashes but that's more consistent with our output in object stores

PR #40, opened 2025-02-20
Add our own binary rounding implementation

samn:
translated from the Rust reformatters. To allow us to get the space savings from binary rounding without warnings about numcodecs.

### DIFF
--- a/src/reformatters/common/binary_rounding.py
+++ b/src/reformatters/common/binary_rounding.py
@@ -4,7 +4,7 @@ import numpy as np
 
 from reformatters.common.types import ArrayFloat32
 
-# IEEE 754 binary32 constants
+# IEEE 754 binary32 (float32) constants
 MANTISSA_BITS: Final[int] = 23
 MANTISSA_MASK: Final[np.uint32] = np.uint32((1 << MANTISSA_BITS) - 1)
 EXPONENT_MASK: Final[np.uint32] = np.uint32(((1 << 8) - 1) << (32 - 8 - 1))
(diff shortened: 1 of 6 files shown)


aldenks:
Pushed tests and it's looking correct. I think it could potentially be optimized, we don't need to guarantee it returns a copy.

There's a lingering error from the restructure: 
```
src/reformatters/common/binary_rounding.py: error: Source file found twice under different module names: "reformatters.common.binary_rounding" and "src.reformatters.common.binary_rounding"
Found 1 error in 1 file (errors prevented further checking)
```

src/reformatters/common/binary_rounding.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Returns:
+        Rounded float32 array
+    """
+    if keep_mantissa_bits >= MANTISSA_BITS:
+        return value
+
+    # Early return for any NaN or Inf values
+    if np.any(~np.isfinite(value)):
+        return value.copy()
```
samn: This should store a mask of nan values to re-apply later, so we can still round if some values are nan

src/reformatters/common/binary_rounding.py
```diff
@@ -0,0 +1,87 @@
...
+    overflow_mask = mantissa > MANTISSA_MASK
+    if np.any(overflow_mask):
+        # Increment exponent and wrap mantissa
+        exponent[overflow_mask] += np.uint32(1 << MANTISSA_BITS)
+        mantissa[overflow_mask] &= MANTISSA_MASK
+
+        # Check for exponent overflow
+        inf_mask = exponent >= EXPONENT_MASK
+        if np.any(inf_mask):
+            result = value.copy()
```
samn: remove?

PR #41, opened 2025-02-21
Apply own implementation of bitrounding

aldenks:
numba go brrrr -- i confirmed this does no significant allocations and it rounds an 8gb array in about a second on my laptop.

### DIFF
--- a/src/reformatters/common/binary_rounding.py
+++ b/src/reformatters/common/binary_rounding.py
@@ -1,6 +1,7 @@
 from typing import Final
 
 import numpy as np
+from numba import njit, prange  # type: ignore
 
 from reformatters.common.types import ArrayFloat32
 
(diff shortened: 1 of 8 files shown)


src/reformatters/common/binary_rounding.py
```diff
@@ -16,74 +17,78 @@ def round_float32_inplace(value: ArrayFloat32, keep_mantissa_bits: int) -> Array
...
 
-    # Apply rounding rules directly to mantissa
+@njit(parallel=True)  # type: ignore
+def _round_float32_inplace_numba(
+    bits: np.ndarray[tuple[int, ...], np.dtype[np.uint32]],
+    keep_mantissa_bits: int,
+    mantissa_bits: int,
+    mantissa_mask: np.uint32,
+    exponent_mask: np.uint32,
+    sign_mask: np.uint32,
```
samn, line 50: Why do these need to be arguments, instead of using the constants? Is it a `numba` thing?

samn (approved):
Very  cool!

src/reformatters/common/binary_rounding.py
aldenks: yeah, numba functions can't access globals. ill add a comment

PR #43, opened 2025-02-25
Deaccumulate precipitation into rate

aldenks:


aldenks:
Example of raw (top) and deaccumulated values (bottom) for a forecast at a point. What im seeing here all looks correct to me; also the tests are handwritten and cover the key cases.

image

aldenks:
self review complete

src/reformatters/noaa/gefs/forecast/reformat.py
```diff
@@ -458,6 +459,20 @@ def groupbykey(
...
+                            )
+                            try:
+                                deaccumulate_to_rates_inplace(
+                                    data_array, dim="lead_time"
+                                )
+                            except ValueError as e:
+                                # Log exception so we are notified if deaccumulation errors are larger than expected.
+                                logger.exception(
+                                    f"Error deaccumulating {data_var.name}", e
+                                )
```
samn: No need for `e`, `logger.exception` captures it from `sys.exc_info()` automatically:

```suggestion
                            except ValueError:
                                # Log exception so we are notified if deaccumulation errors are larger than expected.
                                logger.exception(
                                    f"Error deaccumulating {data_var.name}"
                                )
```

samn (approved):

PR #44, opened 2025-02-25
Interpolate 0.5 degree data to 0.25 for consistency

samn:


samn:
Seeing some small differences in the interpolated data compared to main beyond resampling:


<details><summary>Main (`maybe_resample`) </summary>

246h

</details> 

<details><summary>Interpolating via rasterio (bilinear)</summary>

246h_branch

</details>

samn:
It's not just from the bilinear resampling, seeing a slight difference with nearest-neighbor too (vs main):

<details><summary>Interpolating via Rasterio NN</summary>

246h_branch_nearest


</details>

aldenks:
> Seeing some small differences in the interpolated data compared to main beyond resampling:

Those differences vs main look expected and good, looking closer at the others. The naive, incorrect previous "resampling" we were doing added a 0.125 degree translation in the south direction by making 4 pixels for each output pixel (which then gave 722 pixels and we cropped off the south pole row -- definitely wrong and not something to emulate)

aldenks:
i just remembered another clear way to check: with the exact same file, we can read it normally and with our resampling and then every other pixel should be very close eg, `raw_data ~= resampled_data[::2, ::2]`
 
 code looks great in general once we get those checks right.

aldenks:
Attaching this very helpful graphic of why the resampling has to be more complex than just "duplicate each pixel 4x" @margocrawf drew up for posterity.

The core issue: every other pixel _center_ in the 0.25 degree data aligns with a pixel center in the 0.5 degree data. The pixels between each 0.5 degree pixel we interpolate with bilinear interp (suggestion number 2).

gefs-zarr-spatial-resolutions

aldenks:
Some debugging info that gives confidence the values are (edit: closely but not quite) aligned:

```
INFO:reformatters.noaa.gefs.forecast.reformat:Reading temperature_2m
Window(col_off=np.float64(0.25), row_off=np.float64(0.25), width=np.float64(720.0), height=np.float64(360.5))
Raw read:
[[-13.603155 -13.273138]
 [-11.443151 -11.753149]
 [-12.613165 -13.933142]]
Average resampled read
:[[-13.603155  -13.449781  -13.273138  -13.00864  ]
 [-12.523153  -12.518501  -12.513144  -12.575795 ]
 [-11.443151  -11.587221  -11.753149  -12.14295  ]
 [-12.028158  -12.4069195 -12.843145  -13.290953 ]
 [-12.613165  -13.226618  -13.933142  -14.438955 ]
 [-13.518164  -13.873687  -14.283148  -14.545338 ]]
np.array_equal(raw, result[::2, ::2])=True
```
<details><summary>Images -- im still seeing a tiny tiny east<>west shift, but maybe that's just the resampling</summary>
<p>

Native 0.25 degree (240 hours)
240h

---
Average resampled 0.5 -> 0.25 (246 hours)
246h-average-resampling

---
Bilinear resampled 0.5 -> 0.25 (246 hours)
246h-bilinear


</p>
</details> 

Yes, bilinear looks better than avg resampled, but I think its valuable to be able to get back to the source values exactly.

aldenks:
After sleeping on it realized the windows row and col offset should be 0.125 not 0.25. The 0.5 deg data goes out 0.25 outside of -180,90 and the 0.25 deg data goes 0.125 so the offset is 0.125. 

That would explain the very subtle shift. We're close!

aldenks:
i may have misunderstood windows and the code on the branch may be as correct as we can get. It appears the window is around the pixel centers, if you ask rasterio for the bounds of the window it does extend out based on the pixel size:

```
Window(col_off=np.float64(0.25), row_off=np.float64(0.25), width=np.float64(720.0), height=np.float64(360.5))

reader_0p5.transform
Affine(0.5, 0.0, -180.25,
       0.0, -0.5, 90.25)

rasterio.windows.bounds(window, reader_0p5.transform)
(np.float64(-180.125), np.float64(-90.125), np.float64(179.875), np.float64(90.125))
```

aldenks:
Ok this is fixed up and working as we want it to! By explicitly doing a rasterio.warp.reproject we get nice bilinear interp, no shift at all, and retain the exact values in the 0.5 degree source data at the pixels where they match the 0.25 data (aka we can `assert np.array_equal(raw, result[::2, ::2])`

In addition, the interpolated values between pixels are exact equal weighted averages of pixels they are between, just how we'd like.

Here's video evidence of the alignment between 0.25 degreee data and the resampled 0.5 degree data: https://upstreamtech.wistia.com/medias/fwypggxlbn

<details><summary>The images in that video:</summary>
<p>

0p5-bilinear


0p25


</p>
</details>

src/reformatters/noaa/gefs/forecast/reformat.py
```diff
@@ -471,6 +462,16 @@ def groupbykey(
                             template_ds[data_var.name].encoding["preferred_chunks"]
                         )
 
+                        # Drop all non-dimension coordinates.
+                        # They are already written by write_metadata.
+                        data_array = data_array.drop_vars(
```
aldenks, line 467: This is moved down so the spatial_ref coordinate is available when calling read_into

aldenks (approved):

PR #45, opened 2025-02-25
Configure chunk sharding

aldenks:
Compared to our v2 zarr chunks:
- The chunks are smaller
- The shards are larger
- We now chunk the lead time dimension into 3 chunks so if you only need the first week or 10 days, that's 1/3 the data
- Spatial chunks are about 20x smaller so point queries should be that much faster
- Spatial shards break the world into 8 shards, so we should be able to maximize network throughput well when making global maps
- encoding for all coordinates uses still puts the whole array in a single chunk (and now shard) so chunks == shards for all our coordinates. xarray loads all dimension coordinates on dataset open and all our coords compress very small regardless so they should be in a single shard/chunk.


aldenks:
This is throwing `ValueError: Stored and computed checksum do not match. Stored: b'\xd4\xaf\x85Z'. 
Computed: b'N\xcci\x88'.` when the shard and chunk shape are not the same. Wondering if it's something about our `.to_zarr(region="auto")` call not being supported well

aldenks:
Resolved the checksum issue! we're living in the sharded zarr future bebe.

Before calling to_zarr we were setting dask chunks to the chunk size to write chunks in parallel. With sharding the shard (not the chunk) is [the smallest unit that can safely be written](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#sharding) so we set dask chunks == shards, avoiding multiple dask chunks trying to write to the same shard. [source: joe https://github.com/pydata/xarray/discussions/9938#discussioncomment-11821657]

src/reformatters/noaa/gefs/forecast/template_config.py
```diff
@@ -122,12 +149,7 @@ def get_init_time_coordinates(
     dtype="float32",
     fill_value=np.nan,
     chunks=ENSEMBLE_VAR_CHUNKS_ORDERED,
-    compressors=[BLOSC_4BYTE_ZSTD_LEVEL3_SHUFFLE],
-)
-ENCODING_CATEGORICAL_WITH_MISSING_DEFAULT = Encoding(
```
aldenks, line 127: Removing this constant isn't really part of this PR, but now that bitrounding is not in the encoding it is the same as the ENCODING_FLOAT32_DEFAULT so this PR removes it and uses ENCODING_FLOAT32_DEFAULT everywhere

samn (approved):

PR #46, opened 2025-02-26
Use the correct entrypoint for deploys

samn:


deploy/Dockerfile
```diff
@@ -58,12 +58,12 @@ EOT
 # LEAVE THIS OUT if your application is NOT a proper Python package.
```
samn: We need to do this to get the `reformatters` module installed (so imports work the same as in dev)

src/reformatters/common/kubernetes.py
```diff
@@ -8,7 +8,7 @@
 class Job(pydantic.BaseModel):
     model_config = pydantic.ConfigDict(arbitrary_types_allowed=True)
 
-    command: Annotated[Iterable[str], pydantic.Field(min_length=1)]
```
samn, line 11: Pydantic was coercing lists into some type that can only be iterated once to satisfy this type hint, causing the `command` to get consumed when generating the `job_name`, resulting in an empty `command` below (when used for the K8S command). Adjusting the type hint let the `list` passed in avoid conversion and be iterated more than once.

aldenks: TIL, thanks for the explanation

deploy/Dockerfile
aldenks: Great! I can't comment on it, but now we do have a real app, we can remove the line around 107 under the comment "f your application is NOT a proper Python package ..."

aldenks (approved):

aldenks:
Noting that my comment about line 107 under the comment "f your application is NOT a proper Python package ..." was wrong, we still need it

PR #48, opened 2025-02-26
Fix deaccumulation with trailing dimensions after lead_time dim of length > 1

aldenks:
This is no longer parallel over _all_ dimensions except lead time, just dimensions proceeding lead time. In practice, with 31 ensemble members before the lead_time dim that's plenty.

### DIFF
--- a/src/reformatters/noaa/gefs/deaccumulation.py
+++ b/src/reformatters/noaa/gefs/deaccumulation.py
@@ -26,9 +26,13 @@ def deaccumulate_to_rates_inplace(
     assert np.isin(np.diff(lead_time_seconds), supported_accum_durations).all()
     is_3h_accum = (lead_time_seconds % SECONDS_PER_6_HOUR) != 0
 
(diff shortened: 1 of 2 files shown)


src/reformatters/noaa/gefs/deaccumulation.py
```diff
@@ -57,55 +61,61 @@ def _deaccumulate_to_rates_numba(
...
     """
-    n_lead_times = values.shape[-1]
+    assert values.ndim == 3
+    assert lead_time_seconds.ndim == 1
+    assert is_3h_accum.ndim == 1
+    assert values.shape[1] == lead_time_seconds.size
+    assert values.shape[1] == is_3h_accum.size
 
-    # Flatten all dimensions except lead_time for parallel processing
-    flat_values = values.reshape((-1, n_lead_times))
```
aldenks, line 69: This reshape, _after_ the moveaxis above, was the issue. in plain numpy this makes a copy (because moveaxis made the array not contiguous), in numba its not allowed.

PR #49, opened 2025-02-27
WIP try to get to_zarr to use cpu well

aldenks:


aldenks:
this didn't work but https://github.com/dynamical-org/reformatters/pull/50 did

PR #51, opened 2025-02-28
Add arm64 support for the production image

samn:


samn:
Image creation was super slow locally (since it had to run the arm64 build in an emulated vm). Hopefully it runs faster in CI, but if not it's easy to use Depot

aldenks:
nice!  [The reason we're doing this is arm cores are generally cheaper on AWS]

aldenks (approved):

PR #55, opened 2025-03-01
Use depot for docker builds if available

aldenks:


aldenks:
Should we just not with the building images locally when kicking off a backfill? It's convenient but maybe too loosy goose and we should require we make PRs to main and then only use images built from CI on main.

aldenks:
wait.. this whole PR can be replaced with just running `depot configure-docker` if you want do use depot.

PR #56, opened 2025-03-01
Remove add_offset and scale_factor from encodings and adjust bit rounding

aldenks:
It appears they are not respected in xarray with zarr v3 and if we adjust bitrounding we can maintain precision and compression without them.

### DIFF
--- a/src/reformatters/noaa/gefs/forecast_35_day/template_config.py
+++ b/src/reformatters/noaa/gefs/forecast_35_day/template_config.py
@@ -328,7 +328,7 @@ COORDINATES: Sequence[Coordinate] = (
 _DATA_VARIABLES = (
     # GEFSDataVar(
     #     name="visibility_surface",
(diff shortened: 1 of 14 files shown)


src/reformatters/noaa/gefs/forecast_35_day/template_config.py
```diff
@@ -945,7 +945,7 @@ def get_init_time_coordinates(
             grib_index_level="mean sea level",
             gefs_file_type="s+a",
             index_position=38,
-            keep_mantissa_bits=GEFS_BITROUND_KEEP_MANTISSA_BITS_DEFAULT,
+            keep_mantissa_bits=8,
```
aldenks, line 948: I later updated this to 10.  https://github.com/dynamical-org/reformatters/commit/e53a68c8fe563cc70ac8e39020ec3565b845478f

Because of the very large values in this variable (around 100,000) and how floating point numbers quantize more the farther from zero you are, we retain more bits to get equivalent precision and compression to other variables.

PR #76, opened 2025-03-12
Refactor reformat code

dan-katz:


src/reformatters/noaa/gefs/read_data.py
```diff
@@ -98,15 +99,15 @@ def download_file(
     local_base_file_name = remote_path.replace("/", "_")
 
     idx_remote_path = f"{remote_path}.idx"
-    idx_local_path = Path(f"{local_base_file_name}.idx")
+    idx_local_path = Path(DOWNLOAD_DIR, f"{local_base_file_name}.idx")
```
dan-katz, line 102: I added `DOWNLOAD_DIR` here because tests were writing the idx files to the working directory since we're no longer cd'ing into the download directory

src/reformatters/noaa/gefs/forecast_35_day/reformat.py
```diff
@@ -390,181 +390,62 @@ def reformat_init_time_i_slices(
...
+                                data_array_template,
+                                shared_buffer.name,
+                                store,
+                            ),
+                            shard_indexers,
                         )
+                    )
 
-                        yield (data_var, max_lead_times)
+                    yield (data_var, max_lead_times)
```
aldenks: I'd pull this for loop body into a function and in that function call one sub-function for each of these logical steps:
- filter coords and paths
- create data array
- read files
- apply data transformations inplace
- write shards

(we could call each of those functions directly in the for loop body if another function for the whole loop body seems excessive)

aldenks (changes requested):
Awesome! this is going to be muuuch easier to work with

aldenks (approved):
:chef kiss:!

PR #79, opened 2025-03-17
[HRRR] Scaffold CLI and Start Template Config

acannistra:
Taking things slow with relatively quick PR that introduces HRRR via the following changes: 
* add `noaa/hrrr/forecast_48_hour` module 
* add `cli.py` with `NotImplementedError` on all of the commands, to be filled in 
* add `template_config.py` with a smattering of real and placeholder values

I think I mentioned this briefly on our call @aldenks , but HRRR has an irregular grid (unlike e.g. the regular lat/lon grid that GEFS has). This means that `latitude` and `longitude` can't be dimensions in the `Dataset` template (unless we re-grid all gridpoints to a common grid, which we probably don't want to do, but we could).

I introduced an `x` and `y` dimension to capture this, but I left out the latitude/longitude coordinates for now since I don't totally grok that part of the update flow yet. I wanted to get something up without complicating that, so I commented those coordinates. Would be happy to have a chat about that. 

Ideally the coordinates look something like: 

```
Coordinates:
    ...
    latitude           (y, x) float64 ...
    longitude          (y, x) float64 ...
    ...
```
but I'm not yet sure how to express that in the template framework. Happy to have a chat about that. 

More to come.

### DIFF
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
@@ -1,0 +1,62 @@
+from typing import Any, Literal
+
+import numpy as np
+import pandas as pd
+
+from reformatters.common.config_models import DatasetAttributes
+from reformatters.common.types import DatetimeLike
+
+DATASET_ID = "noaa-hrrr-forecast-48-hour"
+DATASET_VERSION = "0.0.0"
+
+INIT_TIME_START = pd.Timestamp("2014-07-30T18:00")
+INIT_TIME_FREQUENCY = pd.Timedelta("6h")
+LEAD_TIME_FREQUENCY = pd.Timedelta("1h")
+
+DATASET_ATTRIBUTES = DatasetAttributes(
+    dataset_id=DATASET_ID,
+    dataset_version=DATASET_VERSION,
+    name="NOAA HRRR forecast, 48 hour",
+    description="Weather forecasts from the High Resolution Rapid Refresh (HRRR) model operated by NOAA NWS NCEP.",
+    attribution="NOAA NWS NCEP HRRR data processed by dynamical.org from NOAA Open Data Dissemination archives.",
+    spatial_domain="CONUS",
+    spatial_resolution="3km",
+    time_domain=f"Forecasts initialized {INIT_TIME_START} UTC to Present",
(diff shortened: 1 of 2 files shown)


aldenks:
Heck yea! ill review today

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -0,0 +1,241 @@
...
+        ),
+        attrs=CoordinateAttrs(
+            units="seconds",
+            statistics_approximate=StatisticsApproximate(
+                min=str(_dim_coords["lead_time"].min()),
+                max=str(_dim_coords["lead_time"].max()),
+            ),
+        ),
+    ),
+    # TODO: encode x/y dimensions in latitude/longitude coordinates
```
aldenks, line 130: Re x/y and lon/lat -- agreed on your approach and that we should not regrid.

You can add x and y as additional coordinates in this list and leave the longitude, latitude coordinates in as well in the list. The longitude, latitude coords would have 2 dimensional chunk shapes.

In the `get_template_dimension_coordinates` function you'd only return coordinates for x, y like you're doing in the commented code (since they're the dimension coords) and then in your template.py in a `derive_coordinates` function we can compute longitude and latitude coordinates from the dataset (including its x, y coords and its crs metadata).

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -0,0 +1,241 @@
...
+            ),
+        ),
+    ),
+    Coordinate(
+        name="lead_time",
+        encoding=Encoding(
+            dtype="int64",
+            fill_value=-1,
+            # TODO
+            # compressors=[BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE],
```
aldenks, line 117: All these commented out compressors look like the right ones for the coordinates

aldenks (approved):
Very nice! normally id like you do the honors but I'll merge now so you can get this on main and keep rolling.

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
acannistra: great, that makes sense!

PR #82, opened 2025-03-20
Make shared memory request configurable in k8s job

aldenks:


src/reformatters/common/kubernetes.py
```diff
@@ -15,6 +15,7 @@ class Job(pydantic.BaseModel):
     dataset_id: Annotated[str, pydantic.Field(min_length=1)]
     cpu: Annotated[str, pydantic.Field(min_length=1)]
     memory: Annotated[str, pydantic.Field(min_length=1)]
+    shared_memory: Annotated[str, pydantic.Field(min_length=1)] = "1k"
```
aldenks, line 18: 1k == almost nothing as a default if you don't request

src/reformatters/noaa/gefs/forecast_35_day/reformat.py
```diff
@@ -81,6 +81,7 @@ def reformat_kubernetes(
         parallelism=parallelism,
         cpu="6",  # fit on 8 vCPU node
         memory="60G",  # fit on 64GB node
+        shared_memory="24G",
```
aldenks, line 84: the previous hardcoded value was 22Gi (base 2), the nearest whole G (base 10) that will fit the array is 24G. using G to be consistent with other values passed in. 

gibibytes
```
((31*181*1440*721) * 4) / 2**30 == 21.701919436454773
```

gigabytes
```
((31*181*1440*721) * 4) / 1e9 == 23.30225856
```
Multiply the number of elements by 4 bytes because this is an array of float32s

dan-katz (approved):

Issue #81, opened 2025-03-19
NOAA HRRR forecast, 48 hour


aldenks:
Reformat NOAA's HRRR forecast, for this dataset retrieving only 48 hour long forecasts issued every 6 hours.

- follow up comments on #199
  - [x] update the chunk size for the x and y coordinates in the template config so they each have a single chunk. https://github.com/dynamical-org/reformatters/pull/205
  - [x] move `download_hrrr_file` into region job, simplify https://github.com/dynamical-org/reformatters/pull/210
  - [x] simplify band matching in `read_hrrr_data` https://github.com/dynamical-org/reformatters/pull/219
- [x] tweak compression/dtype of `latitude` and `longitude` coordinates, they are surprisingly large right now (14MB each), max `x`, `y` coords real meters https://github.com/dynamical-org/reformatters/pull/205
- [x] test values returned by dimension_coordinates and derive_coordinates, particularly x, y, longitude, latitude coords https://github.com/dynamical-org/reformatters/pull/244
- [x] add spatial reference system https://github.com/dynamical-org/reformatters/pull/244
- [x] an integration dynamical_dataset test that does a short backfill followed by an update https://github.com/dynamical-org/reformatters/pull/219
- [x] template config close review: select variables to start, review variable metadata for standardization, fine tune chunk sizes (vars & metadata: https://github.com/dynamical-org/reformatters/pull/248 chunks: https://github.com/dynamical-org/reformatters/pull/219)
- [x] validate data after running a backfill
- [x] operational resource definitions and schedule and ship it! (easy)

aldenks:
@acannistra creating an issue here for bigger picture, async communication!

first update: i just merged the initial work for the first of our "derived" datasets (a gefs _analysis_, 3D time x lon x lat dataset) where we run into having multiple "flavors" of dataset from the same model. I'm using a pattern of a common variable/config file for the model that the template_config.py files import and add to. eg https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/noaa/gefs/common_gefs_template_config.py and https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/noaa/gefs/forecast_35_day/template_config.py

acannistra:
Nice, that makes sense! I can take a look and see how you divided up those files to determine if there are any parallels we can use for HRRR.

acannistra:
[...]

PR #90, opened 2025-03-27
[HRRR] Finish template via downloads of actual GRIBs.

acannistra:
Just a quick one here to finish up the template creation. Going to move forward with downloading data vars in the next PR.

#81

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ dependencies = [
     "numpy>=2.1.1",
     "obstore>=0.4.0",
     "pydantic>=2.10.2",
+    "pygrib>=2.1.6",
     "rasterio>=1.4.2",
     "requests>=2.32.3",
     "rioxarray>=0.18.2",
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
@@ -2,6 +2,8 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
+import pandas as pd
+import pygrib  # type: ignore
 import rioxarray  # noqa: F401  Adds .rio accessor to datasets
 import xarray as xr
 
@@ -9,6 +11,7 @@ from reformatters.common.template_utils import assign_var_metadata, make_empty_v
 from reformatters.common.template_utils import (
     write_metadata as write_metadata,  # re-export
 )
+from reformatters.noaa.hrrr.read_data import SourceFileCoords, download_file
 
 from .template_config import (
     COORDINATES,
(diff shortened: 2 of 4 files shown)


src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +60,24 @@ def update_template() -> None:
...
+    # a sample file and extracting the coordinates from the file itself.
+    data_coords = SourceFileCoords(
+        {
+            "init_time": pd.Timestamp.now(),
+            "lead_time": pd.Timedelta("0h"),
+            "domain": "conus",
+        }
+    )
+    # "sfc" is the smallest available file type.
+    _, filepath = download_file(data_coords, "sfc")
```
aldenks, line 73: once you add the option to select a variable lets remember to add that here and download much less data.

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +60,24 @@ def update_template() -> None:
 def derive_coordinates(
     ds: xr.Dataset,
 ) -> dict[str, xr.DataArray | tuple[tuple[str, ...], np.ndarray[Any, Any]]]:
+    # derivation of HRRR latitude / longutide coordinates is simplest by downloading
+    # a sample file and extracting the coordinates from the file itself.
+    data_coords = SourceFileCoords(
+        {
+            "init_time": pd.Timestamp.now(),
```
aldenks, line 67: I think this might be working by chance of formatting logic (since there won't really be an init with some random seconds on the end). Let's hardcode an arbitrary but real init time.

pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
     "numpy>=2.1.1",
     "obstore>=0.4.0",
     "pydantic>=2.10.2",
+    "pygrib>=2.1.6",
```
aldenks, line 16: woah this installs eccodes just via a pip install? miraculous :+1:

aldenks: We might want to use their index filtering functionality to simplify how we parse grib indexes. you could try that out for HRRR and see how the experience is. The one challenge might be that we need to get the matching row _and the next one_ in order to get the byte range end point. The regex we use right now does that in one fell swoop (it's downside is just that its a modestly complex regex)

aldenks (approved):

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
aldenks: I'm going to merge this so you can cut a fresh branch but will you include that change in your next PR?

PR #91, opened 2025-03-27
Revert "[HRRR] Finish template via downloads of actual GRIBs."

aldenks:
Reverts dynamical-org/reformatters#90

pygrib appears to not actually install all of eccodes for you automatically -- twas too good to be true? 

https://github.com/dynamical-org/reformatters/actions/runs/14108313248/job/39520118245

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,6 @@ dependencies = [
     "numpy>=2.1.1",
     "obstore>=0.4.0",
     "pydantic>=2.10.2",
-    "pygrib>=2.1.6",
     "rasterio>=1.4.2",
     "requests>=2.32.3",
     "rioxarray>=0.18.2",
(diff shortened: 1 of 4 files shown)


aldenks:
@acannistra heads up I needed to revert because pygrib doesn't actually seems to be fully installing eccodes ([logs](https://github.com/dynamical-org/reformatters/actions/runs/14108313248/job/39520118245)). Requiring users to manually install eccodes is enough additional complexity I'd like to avoid it if possible. We could attempt to fix the pygrib packaging but my experience installing eccodes is it's not going to be easy to make happen consistently across unix platforms/arches.

What if we copy the logic from pygrib? https://github.com/jswhit/pygrib/blob/master/src/pygrib/_pygrib.pyx#L1623-L1644 I think we can get all the required info (latitudeOfFirstGridPointInDegrees, ..., Nx, ..., DxInMetres) from rasterio under different names.

acannistra:
Strange. I'll take a look!

PR #93, opened 2025-03-28
Skip reading min and max variables for hour zero gribs

dan-katz:
Min and Max variables aren't in hour zero gribs

Also do a little cleanup and escape the string we use to match grib variables to handle the case where there are parentheses in the grib index level, like `entire atmosphere (considered as a single layer)`

### DIFF
--- a/src/reformatters/noaa/gfs/read_data.py
+++ b/src/reformatters/noaa/gfs/read_data.py
@@ -43,7 +43,7 @@ def download_file(
         gfs_idx_data_vars = [
             data_var
             for data_var in gfs_idx_data_vars
-            if data_var.attrs.step_type not in ("accum", "avg")
+            if data_var.attrs.step_type not in ("accum", "avg", "min", "max")
         ]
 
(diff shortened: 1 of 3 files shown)


src/reformatters/noaa/gfs/read_data.py
```diff
@@ -115,19 +115,22 @@ def parse_index_byte_ranges(
...
         if lead_time_hours == 0:
-            hours_str = ""
+            hours_str_prefix = ""
         elif var_info.attrs.step_type == "instant":
-            hours_str = f"{int(lead_time_hours)}"
+            hours_str_prefix = str(lead_time_hours)
         else:
             diff_hours = lead_time_hours % 6
             if diff_hours == 0:
                 reset_hour = lead_time_hours - 6
```
aldenks: Can we put this 6 in a constant ?

aldenks (approved):

src/reformatters/noaa/gfs/read_data.py
dan-katz: Good idea

PR #94, opened 2025-03-28
[HRRR] retry adding latitude + longitude coords by computing them this time

acannistra:
Basically copying the `pygrib` implementation here. Unfortunately I couldn't determine an easy way to get the GRIB parameters that were critical to getting this right (specifically the corner lat/lon coords of the domain) so I hard-coded them. Don't love that, but also quite unlikely to change. I tried a few things (like using the `ds.rio.transform`) but I strangely could only get _close_ to the right values but not quite. I'm not sure why. 

I also added the projected `x` and `y` coordinates. 

I downloaded an HRRR GRIB from the AWS bucket (https://noaa-hrrr-bdp-pds.s3.amazonaws.com/index.html) and (in an environment with pygrib) used this script to validate accuracy: 

```python
import xarray as xr
import pygrib 
import numpy as np

ds = xr.open_zarr("src/reformatters/noaa/hrrr/forecast_48_hour/templates/latest.zarr")
grib = pygrib.open("<path to hrrr_grib")

gribLats, gribLons = grib[1].latlons()

assert np.allclose(ds.latitude.values, gribLats)
assert np.allclose(ds.longitude.values, gribLons)
```

### DIFF
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
@@ -2,6 +2,8 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
+import pandas as pd
+import pygrib  # type: ignore
 import rioxarray  # noqa: F401  Adds .rio accessor to datasets
 import xarray as xr
 
@@ -9,6 +11,7 @@ from reformatters.common.template_utils import assign_var_metadata, make_empty_v
 from reformatters.common.template_utils import (
     write_metadata as write_metadata,  # re-export
 )
+from reformatters.noaa.hrrr.read_data import SourceFileCoords, download_file
 
 from .template_config import (
     COORDINATES,
@@ -57,11 +60,24 @@ def update_template() -> None:
 def derive_coordinates(
     ds: xr.Dataset,
 ) -> dict[str, xr.DataArray | tuple[tuple[str, ...], np.ndarray[Any, Any]]]:
+    # derivation of HRRR latitude / longutide coordinates is simplest by downloading
+    # a sample file and extracting the coordinates from the file itself.
+    data_coords = SourceFileCoords(
+        {
+            "init_time": pd.Timestamp.now(),
+            "lead_time": pd.Timedelta("0h"),
+            "domain": "conus",
+        }
+    )
+    # "sfc" is the smallest available file type.
+    _, filepath = download_file(data_coords, "sfc")
+    grib = pygrib.open(filepath)
+    lats, lons = grib[1].latlons()
+    grib.close()
+
     return {
-        # TODO: Derive these coordinates from an actual model file
-        "latitude": (("y", "x"), np.full((ds["y"].size, ds["x"].size), 0.0)),
-        # TODO: Derive these coordinates from an actual model file
-        "longitude": (("y", "x"), np.full((ds["y"].size, ds["x"].size), 0.0)),
+        "latitude": (("y", "x"), lats),
+        "longitude": (("y", "x"), lons),
         "ingested_forecast_length": (
             ("init_time",),
             np.full((ds["init_time"].size), np.timedelta64("NaT", "ns")),
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
@@ -79,7 +79,13 @@ def derive_coordinates(
     hrrrds = xr.open_dataset(str(filepath), engine="rasterio")
 
     hrrrds_bounds = hrrrds.rio.bounds(recalc=True)
-    dx, dy = 3000, 3000
+    hrrrds_res = hrrrds.rio.resolution(recalc=True)
+    dx, dy = (
+        hrrrds_res[0],
+        hrrrds_res[1] * -1,
+    )  # y offset is positive, but reported as negative by `rio.resolution`
+    # this may be a bug in how rioxarray is handling bounds that cross a central parallel
+    # but also I'd expect that to be the case with lat/lon coordinates too, so I'm not sure.
 
     proj_xcorner, proj_ycorner = hrrrds_bounds[0], hrrrds_bounds[1]
     nx = ds.x.size
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
@@ -2,12 +2,18 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
+import pandas as pd
 import pyproj
+import rioxarray  # noqa: F401  Adds .rio accessor to datasets
 import xarray as xr
 
 from reformatters.common.template_utils import assign_var_metadata, make_empty_variable
 from reformatters.common.template_utils import (
     write_metadata as write_metadata,  # re-export
+)
+from reformatters.noaa.hrrr.read_data import (
+    SourceFileCoords,
+    download_file,
 )
 
 from .template_config import (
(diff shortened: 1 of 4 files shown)


src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +61,29 @@ def update_template() -> None:
...
+            "lead_time": pd.Timedelta("0h"),
+            "domain": "conus",
+        }
+    )
+    # "sfc" is the smallest available file type.
+    _, filepath = download_file(data_coords, "sfc")
+
+    hrrrds = rio.open(str(filepath))
+    height, width = hrrrds.shape
+    yidxs, xidxs = np.meshgrid(np.arange(height), np.arange(width), indexing="ij")
```
acannistra, line 78: pretty sure this is wrong

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +57,27 @@ def update_template() -> None:
...
+    lons, lats = pj(xs, ys, inverse=True)
+    proj_xs, proj_ys = pj(lons, lats)
+
     return {
-        # TODO: Derive these coordinates from an actual model file
-        "latitude": (("y", "x"), np.full((ds["y"].size, ds["x"].size), 0.0)),
-        # TODO: Derive these coordinates from an actual model file
-        "longitude": (("y", "x"), np.full((ds["y"].size, ds["x"].size), 0.0)),
+        "y": (("y",), proj_ys[:, 0]),
+        "x": (("x",), proj_xs[0, :]),
```
acannistra: I was kinda tired when I wrote this - there might be a smoother numpy-fu here.

aldenks: Can the values for these just be `x` and `y`? (or if not, do we understand why not?)

the dim name can also be a single string if there's only one dim, eg

```suggestion
        "y": ("y", y),
        "x": ("x", x),
```

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +57,27 @@ def update_template() -> None:
 def derive_coordinates(
     ds: xr.Dataset,
 ) -> dict[str, xr.DataArray | tuple[tuple[str, ...], np.ndarray[Any, Any]]]:
+    # following pygrib methodology for lat/lon calculation
+    # https://github.com/jswhit/pygrib/blob/a69fb89e62f4ba5fffdfa86aee40948ee26f5c42/src/pygrib/_pygrib.pyx#L1623-L1644
+    lat_corner = 21.138123
+    lon_corner = 237.280472
+    nx = 1799
+    ny = 1059
+    dx, dy = 3000.0, 3000.0
```
aldenks: To reduce hard coding, can we pull any of lat_corner, lon_corner, dx, dy from ds.rio.transform? And can we pull nx, ny from ds.sizes?

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -57,11 +57,27 @@ def update_template() -> None:
...
+
+    pj = pyproj.Proj(ds.rio.crs.to_proj4())
+    x_corner, y_corner = pj(lon_corner, lat_corner)
+    x = x_corner + np.arange(nx) * dx
+    y = y_corner + np.arange(ny) * dy
+    xs, ys = np.meshgrid(x, y)
+    lons, lats = pj(xs, ys, inverse=True)
+    proj_xs, proj_ys = pj(lons, lats)
+
```
aldenks, line 101: after computing these values, lets assert that the relevant mins/maxes match the values in our attributes. eg `ds["x"].attrs["statistics_approximate"]["min"] == x.min()`

aldenks (changes requested):
Nice! a couple questions to see if we can reduce possibilities for hardcoding to get out of sync/ensure it's right

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
acannistra: I'd really love to hard-code less of this. To be honest, I'm a bit rusty on my affine transformations, but I tried the two permutations I knew of to use the transform to compute these values without going back out to lat/lon, but both were slightly incorrect:

I was sure this was going to work, but it produces inaccurate values and the arrays themselves are flipped:

```python
_, filepath = download_file(data_coords, "sfc")
    hrrrds = rio.open(str(filepath))

    hrrr_xform = hrrrds.transform
    proj_ycorner = hrrr_xform.yoff
    proj_xcorner = hrrr_xform.xoff
    nx = 1799
    ny = 1059
    dx, dy = hrrr_xform.a, hrrr_xform.e

    pj = pyproj.Proj(hrrrds.crs.to_proj4())
    x = proj_xcorner + np.arange(nx) * dx
    y = proj_ycorner + np.arange(ny) * dy
    xs, ys = np.meshgrid(x, y)
    lons, lats = pj(xs, ys, inverse=True)
```

The test I posted above in the PR description fails with this method.

acannistra: Yes. I'd really love to hard-code less of this. I had expected to be able to use the affine transform that RIO gives us to convert the `ij` pixel indices into projected coordinates, then use the inverse CRS transformation to get those into lat/lon. That's what was done in this commit (https://github.com/dynamical-org/reformatters/pull/94/commits/63548a7cf9092537df20daf890b757f765901026) but it was incorrect when compared to the pygrib values. 

I'll grab a couple more of these values from the dataset.

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
acannistra: See I knew I was overthinking it

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
acannistra: ok sounds good. I started in on this initially and found that `ds['x'].attrs` was empty in the `derive_coordinates` function, so I'll need to dig into that a bit

aldenks: oh you're right because this function is called before we call assign_var_metadata. you can skip the check for now and lets think about a general way. perhaps our update template pattern also needs a "derive_attributes" function that runs at the end of update_template and has access to all the coordinate values. that would allow us to remove the dance we do calling get_dimension_coordinates in template_config which would be nice.

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
aldenks: that means we're not assigning the right projection with rio.write_crs no?   

you don't necessarily need to use the transform to do the math (what you're doing now seems clearer than the commit you linked to me), but the transform _should_ give us the dx and dy and the corner x and y (which we could inverse project to get the corner lat and lon) right?

acannistra: I thought that too. I double-checked the projection params from the GRIB2 via pygrib and they are the same as what we are using. 

```
>>> import pygrib
>>> gribmsg = pygrib.open("data/download/hrrr.20250321_conus_hrrr.t10z.wrfsfcf00.grib2")[1]
>>> gribmsg.projparams
{'a': 6371229, 'b': 6371229, 'proj': 'lcc', 'lon_0': 262.5, 'lat_0': 38.5, 'lat_1': 38.5, 'lat_2': 38.5}
```

Compared to https://github.com/dynamical-org/reformatters/blob/641f6d50d5835654a7a495870686b995421342c8/src/reformatters/noaa/hrrr/forecast_48_hour/template.py#L38-L40

acannistra: I think I figured this out. 

The issue is that rasterio (for good reason) is using the _domain bounds_ as the transform source, which means the _corners_ of the domain are in the Affine transformation. 

But pygrib, because it's using the `latitudeOfFirstGridPointInDegrees` / `latitudeOfFirstGridPointInDegrees` GRIB metadata, begins the grid at the _center of a gridcell_, rather than the corner of the domain. 

I implemented that here: 65387c7. The only piece that I couldn't use the `.rio` accessor for was the `dx`/`dy` resolution, because `.rio.resolution()` (and therefore `rio.transform()`) both return `dy` as `-3000` , despite the fact that pygrib doesn't report that. I'm not sure where that's coming from.

aldenks: hey! nice sleuthing. I believe the negative is saying the values re decreasing as you go from north to south. final question: are we confident our `y` should be increasing given the transform reports it should be decreasing? 

Either way, lets pull dx, dy from the transform and if we're really sure dy should be positive then we can multiply it by -1. 

Last question then good to merge this, thanks for hanging in on this

aldenks (changes requested):
realized i left the review request hanging, but my one comment above yesterday is the last thing

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
acannistra: Yes, that's how I originally interpreted it, but I am confused. 

In the grib, both `iScansNegatively` and `jScansNegatively` are `0`, I think implying that the `dX` and `dY` values (which are both `+3000` also) don't increment negatively. 

But honestly, it's hard to know exactly without looking at the data. `rioxarray` computes the resolution (which are used for the bounds in the transform) directly by [subtracting](https://github.com/corteva/rioxarray/blob/564c946d07058b8ebdc86b1751ba24169cd6c6ef/rioxarray/rioxarray.py#L1058-L1059) the `top` y coordinate value from the `bottom` y coordinate value, as [reported](https://github.com/corteva/rioxarray/blob/564c946d07058b8ebdc86b1751ba24169cd6c6ef/rioxarray/rioxarray.py#L1009-L1012) from the underlying xarray dataset. 

In the projected coordinates, the `bottom` value (`ds.rio._obj['y'][-1]`, which is negative) is less than the `top` value (`ds.rio._obj['y'][0]`, which is positive), which is how the Y resolution gets reported as negative. This might be a bug in how rasterio is handling conformal projections with standard parallels, I'm not sure. (e.g. should those values be `abs`-ed).

Anyway, I'm pretty confident, especially since this approach is the only one that gives us parity with what `pygrib`/`eccodes` is reporting, which I trust.

aldenks (approved):

aldenks:
Wooot! thanks for sticking with this @acannistra

PR #96, opened 2025-04-02
GEFS analysis back to 2000 by integrating GEFS v12 reforecast 

aldenks:
Use the GEFS v12 retrospective (reforecast) to extend the start of our analysis back to 2000. This reforecast was issued once per day and we use the first 24 hours of each forecast.

Three key changes to make this work
- update generate_chunk_coordinates following the less frequent init times and lack of 0 hour for any variable in the reforecast
- update download_file to know how to make paths matching the reforecast format
- update grib index parsing to also match on lead time so we can pull the right lead time out of the reforecast files which include all lead times for a single variable together in one file


src/reformatters/noaa/gefs/read_data.py
```diff
@@ -296,77 +342,62 @@ def read_rasterio(
...
-                # Every 2nd (0.5 deg) or every 4th (1.0 deg) 0.25 degree pixel's center aligns exactly
-                # with a 1.0/0.5 degree pixel's center.
-                # We use bilinear resampling to retain the exact values from the 1.0/0.5 degree data
-                # at pixels that align, and give a conservative interpolated value for 0.25 degree pixels
-                # that fall between the 1.0/0.5 degree pixels.
-                # Diagram: https://github.com/dynamical-org/reformatters/pull/44#issuecomment-2683799073
-                # Note: having the .read() call interpolate gives very slightly shifted results
-                # so we pay for an extra memory allocation and use reproject to do the interpolation instead.
-                warnings.filterwarnings(
-                    "ignore", category=rasterio.errors.NotGeoreferencedWarning
```
aldenks, line 329: view this diff in ignore whitespace mode. the change in this function is to move this filterwarnings to before the rasterio.open (where the warning was actually coming from) which required indenting the whole body

src/reformatters/noaa/gefs/read_data.py
```diff
@@ -296,77 +342,62 @@ def read_rasterio(
...
-                result = reader.read(
-                    rasterio_band_index,
-                    out_dtype=np.float32,
-                )
-                return result
-            case _ as unreachable:
-                assert_never(unreachable)
-
-
-class DefaultTimeoutHTTPAdapter(requests.adapters.HTTPAdapter):
```
aldenks, line 362: unused

src/reformatters/common/download.py
```diff
@@ -30,14 +31,21 @@ def download_to_disk(
     else:
         response_buffers = obstore.get(store, path).stream()
 
+    # Avoid race by writing to temp file then renaming when complete
+    temp_path = local_path.with_name(f"{local_path.name}.{uuid.uuid4().hex[:8]}")
```
dan-katz, line 35: What's the race condition?

dan-katz (approved):

src/reformatters/common/download.py
aldenks: different if one thread starts making a file another can come right after, the file.exists() check succeeds so it goes to read it but it actually isn't complete yet (I was seeing empty contents read by the second file)

PR #105, opened 2025-04-04
Support running GFS ingests in kubernetes

dan-katz:


src/reformatters/noaa/gfs/forecast/reformat.py
```diff
@@ -19,7 +25,9 @@
     reformat_time_i_slices,
 )
 
-_VARIABLES_PER_BACKFILL_JOB = 3
+_VARIABLES_PER_BACKFILL_JOB = (
+    30  # More variables than we currently have but have a buffer in case we add more
```
aldenks: im kinda partial to putting the comment above so ruff doesn't do the wrap with parens thing

aldenks (approved):
Let's goo!

src/reformatters/noaa/gfs/forecast/reformat.py
dan-katz: Done

PR #110, opened 2025-04-07
Buffer by _2_ time steps so we have both valid endpoints and accumulation resets on first step

aldenks:


aldenks:
Over the shoulder reviewed by @dan-katz

PR #113, opened 2025-04-08
[HRRR] add selective downloading to HRRR

acannistra:
A bit busier around here these days, less progress than I'd hoped. But wanted to get a small iteration up. 

I saw what I thought was a bunch of duplicated downloading code across the NOAA models (GFS/GEFS/HRRR) and so set out to reduce the duplication across the various `read_data` modules, but ended up getting the runaround with some small differences. So I stashed all that into another branch and just went ahead with this duplicative but simple implementation. 

Obviously the important part here is the regex used to parse variables in the IDX file. The only field I'm not using in the HRRR IDX is the last, e.g the `anl` value: 

```
1:0:d=2025010100:REFC:entire atmosphere:anl:
```

I don' think there's any reason to store + match on that since the grib_element and grib_index_level should uniquely define a layer in the GRIB, but I may be wrong. 

Curious what you think next steps should be here @aldenks after we get the template working.

### DIFF
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template.py
@@ -11,6 +11,7 @@ from reformatters.common.template_utils import assign_var_metadata, make_empty_v
 from reformatters.common.template_utils import (
     write_metadata as write_metadata,  # re-export
 )
+from reformatters.noaa.hrrr.hrrr_config_models import HRRRDataVar
 from reformatters.noaa.hrrr.read_data import (
     SourceFileCoords,
     download_file,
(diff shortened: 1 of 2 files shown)


aldenks:
sorry i didn't get to this yesterday, will do today!

aldenks (approved):
Sorry for taking a bit to review this. It looks good, im going to merge!

As for duplicated code and the slight nuances, plow ahead with duplication. Using implementation of multiple datasets to see what _really_ is common and what needs to be parameterized per dataset is exactly my hope. We're getting close to have a good enough idea that i'm nearly ready to do some big refactors.

Re matching on the lead time in the grib, don't worry about it here. The one case we've needed it is for the gefs retrospective which puts all lead times for a single variable in a single file.

As for next steps, I'd whole cloth copy the `gfs/forecast/reformat.py` and `reformat_internals.py` and work on getting a `reformat-local` command working locally. in template.get_template() we often add a filter on lead time hours in dev mode, but better yet would be to create a test like we have in  https://github.com/dynamical-org/reformatters/blob/main/tests/noaa/gefs/forecast_35_day/cli_integration_test.py

PR #120, opened 2025-04-14
GEFS analysis operational updating

aldenks:
Approach
1. update `all_jobs_ordered` to support filtering to chunks in a time range
    - this is also useful for reprocessing specific subsets of the dataset and a pattern already implemented in gefs forecast 35 day -- im expecting to refactor that whole function out into common code
1. filter jobs to shards > the latest already processed time

Unlike forecast datasets where we need to track which forecasts are fully ingested with `ingested_forecast_length` coordinate, in the analysis case things are simpler and we just compare the latest time coord with "now".

### DIFF
--- a/src/reformatters/common/zarr.py
+++ b/src/reformatters/common/zarr.py
@@ -90,11 +90,17 @@ def get_mode(
 
 def copy_data_var(
     data_var_name: str,
-    chunk_index: int,
+    i_slice: slice,
+    template_ds: xr.Dataset,
+    append_dim: str,
     tmp_store: Path,
     final_store: zarr.storage.FsspecStore,
 ) -> None:
-    relative_dir = f"{data_var_name}/c/{chunk_index}/"
+    dim_index = template_ds[data_var_name].dims.index(append_dim)
+    append_dim_shard_size = template_ds[data_var_name].encoding["shards"][dim_index]
+    shard_index = i_slice.start // append_dim_shard_size
+    assert dim_index == 0  # relative_dir format below assumes append dim is first
+    relative_dir = f"{data_var_name}/c/{shard_index}/"
 
     logger.info(
         f"Copying data var chunks to final store ({final_store}) for {relative_dir}."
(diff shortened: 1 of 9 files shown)


src/reformatters/common/zarr.py
```diff
@@ -90,11 +90,17 @@ def get_mode(
...
+    append_dim: str,
     tmp_store: Path,
     final_store: zarr.storage.FsspecStore,
 ) -> None:
-    relative_dir = f"{data_var_name}/c/{chunk_index}/"
+    dim_index = template_ds[data_var_name].dims.index(append_dim)
+    append_dim_shard_size = template_ds[data_var_name].encoding["shards"][dim_index]
+    shard_index = i_slice.start // append_dim_shard_size
+    assert dim_index == 0  # relative_dir format below assumes append dim is first
+    relative_dir = f"{data_var_name}/c/{shard_index}/"
```
aldenks, line 104: This was a bug but we weren't running into it because previous datasets had shard size == 1 in the append dimension so the chunk index == shard index

src/reformatters/common/zarr.py
```diff
@@ -90,11 +90,17 @@ def get_mode(
...
```
aldenks, line 103: This was a bug but we weren't running into it because previous datasets had shard size == 1 in the append dimension so the chunk index == shard index.

src/reformatters/noaa/gefs/forecast_35_day/reformat.py
```diff
@@ -351,7 +340,7 @@ def all_jobs_ordered(
             for time_i_slice in time_i_slices
             if template_ds.isel({chunk_filters.time_dim: time_i_slice})[
                 chunk_filters.time_dim
-            ].min()
+            ].max()
```
aldenks, line 343: this min->max and the below max->min change update the filter to include all chunks which are the time filter includes any amount of rather than all chunks fully contained in the time filter.

PR #127, opened 2025-04-22
[HRRR] Add v1 of reformatting 

acannistra:
As suggested in #113, I started with the GFS `reformat.py` and `reformat_internals.py` and modified them to work with HRRR. 

There were a few speedbumps here. The first was that this revealed a bug in the lat/lon calculations, which were wrong. I was using the wrong corner coordinates (Bottom Left instead of Top Left), which ended up resulting in a bad transform in the zarr template. I fixed that. 

There's also a very suspect `.T` in the `read_into` function which I narrowed down to the fact that the `.rio` accessor seems to be reporting the x,y dimensions swapped. I'm not sure what to make of that but I'm going to continue to dig in. 

This PR results in a zarr on disk that contains the REFC data. I've not yet actually looked at it, but I suspect it might be flipped around incorrectly due to the `.T`. But the data are there. 

Oh, and finally: I _think_ HRRR revised their domain between v1 and v4. I'm trying to narrow that down now. but tl;dr we probably can't put the whole HRRR archive (v1-v4) in a single ZARR without reprocessing the old domain into the new one, which I doubt we want to do. 

Testing: 

```bash
uv run main noaa-hrrr-forecast-48-hour update-template
uv run main noaa-hrrr-forecast-48-hour reformat-local 2018-07-14
```

---

Copilot summary:
This pull request introduces significant functionality to the NOAA HRRR 48-hour forecast reformatter, including the implementation of local and Kubernetes-based reformatting workflows, enhancements to template handling, and improvements to data downloading and processing. Below is a summary of the most important changes, grouped by theme.

### New Reformatting Workflows
* Added `reformat_local` and `reformat_kubernetes` functions in `reformat.py` to support local and Kubernetes-based reformatting of dataset chunks. These functions handle metadata writing, chunk processing, and job submission for parallel execution.
* Integrated the `reformat_local` function into the CLI by replacing the placeholder in the `reformat_local` command with a call to the new implementation.

### Template Enhancements
* Introduced a `get_template` function in `template.py` to dynamically generate and expand dataset templates based on the end time of initialization (`init_time_end`). This ensures proper handling of coordinates and metadata.
* Updated the CRS (coordinate reference system) in `update_template` to correct the longitude origin from `+lon_0=262.5` to `+lon_0=-97.5`.

### Data Download and Processing Improvements
* Refactored `download_file` in `read_data.py` to associate file types with variables and validate consistency within variable groups. This ensures that all variables in a group belong to the same file type. [[1]](diffhunk://#diff-0bad925d0e2eaa204497c7ef32f06b09838228b19a892f18fe0e296e48e504d1R25-R57) [[2]](diffhunk://#diff-0bad925d0e2eaa204497c7ef32f06b09838228b19a892f18fe0e296e48e504d1L41-R66)
* Added logging for byte range starts and ends during data downloading to improve traceability.

### Configuration Updates
* Changed the `INIT_TIME_START` in `template_config.py` from `2014-07-30T18:00` to `2020-01-01T00:00` to reflect a more recent starting point for initialization times.
* Renamed a data variable from `Composite reflectivity` to `composite_reflectivity` and updated its GRIB description for clarity and consistency. [[1]](diffhunk://#diff-adf9cca1e30e58cd42d7c106f0d3a8aabec7d98982adfe93583e743379720612L302-R302) [[2]](diffhunk://#diff-adf9cca1e30e58cd42d7c106f0d3a8aabec7d98982adfe93583e743379720612L318-R318)

### Internal Refactoring
* Added `reformat_internals.py` with the `reformat_time_i_slices` function to handle parallelized chunk processing using shared memory and process pools. This improves CPU utilization during reformatting.

These changes collectively enhance the functionality, accuracy, and maintainability of the NOAA HRRR 48-hour forecast reformatter.

### DIFF
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/reformat_internals.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/reformat_internals.py
@@ -1,0 +1,287 @@
+import concurrent
+import os
+from collections.abc import Generator, Iterable, Sequence
+from concurrent.futures import Future, ProcessPoolExecutor, ThreadPoolExecutor
+from functools import partial
+from itertools import batched, groupby, product
+from pathlib import Path
+from typing import Any
+
+import pandas as pd
+import xarray as xr
+import zarr
+
+from reformatters.common.binary_rounding import round_float32_inplace
+from reformatters.common.config_models import DataVar
+from reformatters.common.iterating import consume
+from reformatters.common.logging import get_logger
+from reformatters.common.reformat_utils import (
+    create_data_array_and_template,
+    create_shared_buffer,
+    write_shards,
+)
+from reformatters.noaa.hrrr.forecast_48_hour import template
+from reformatters.noaa.hrrr.hrrr_config_models import (
+    HRRRDataVar,
+    HRRRDomain,
+    HRRRFileType,
+)
+from reformatters.noaa.hrrr.read_data import SourceFileCoords, download_file, read_into
+from reformatters.noaa.noaa_utils import has_hour_0_values
+
+logger = get_logger(__name__)
+
+type CoordAndPath = tuple[SourceFileCoords, Path | None]
+type CoordsAndPaths = list[CoordAndPath]
+
+
+def reformat_time_i_slices(
+    jobs: Sequence[tuple[slice, list[HRRRDataVar]]],
+    template_ds: xr.Dataset,
+    store: zarr.storage.FsspecStore | Path,
+    var_download_group_size: int,
+) -> Generator[tuple[HRRRDataVar, dict[pd.Timestamp, pd.Timedelta]], None, None]:
+    # The only effective way we've found to fully utilize cpu resources
+    # while writing to zarr is to parallelize across processes (not threads).
+    # Use shared memory to avoid pickling large arrays to share between processes.
+    shared_buffer_size = max(
+        data_var.nbytes
+        for data_var in template_ds.isel(
+            {template.APPEND_DIMENSION: jobs[0][0]}
+        ).values()
+    )
+    with (
+        ThreadPoolExecutor(max_workers=1) as wait_executor,
+        ThreadPoolExecutor(max_workers=(os.cpu_count() or 1) * 2) as io_executor,
+        ThreadPoolExecutor(max_workers=os.cpu_count() or 1) as cpu_executor,
+        ProcessPoolExecutor(max_workers=os.cpu_count() or 1) as cpu_process_executor,
+        create_shared_buffer(shared_buffer_size) as shared_buffer,
+    ):
+        for init_time_i_slice, data_vars in jobs:
+            data_var_names = [data_var.name for data_var in data_vars]
+            chunk_template_ds = template_ds[data_var_names].isel(
+                init_time=init_time_i_slice
+            )
+
+            chunk_init_times = pd.to_datetime(chunk_template_ds["init_time"].values)
+            chunk_lead_times = pd.to_timedelta(chunk_template_ds["lead_time"].values)
+            chunk_domains: Iterable[HRRRDomain] = ["conus"]
+            chunk_file_types = [dv.internal_attrs.hrrr_file_type for dv in data_vars]
+            chunk_coords = generate_chunk_coordinates(
+                chunk_init_times, chunk_lead_times, chunk_domains, chunk_file_types
+            )
+
+            chunk_init_times_str = ", ".join(
+                chunk_init_times.strftime("%Y-%m-%dT%H:%M")
+            )
+            logger.info(f"Starting chunk with init times {chunk_init_times_str}")
+
+            download_var_group_futures = get_download_var_group_futures(
+                chunk_template_ds,
+                chunk_coords,
+                io_executor,
+                wait_executor,
+                var_download_group_size,
+            )
+            for future in concurrent.futures.as_completed(download_var_group_futures):
+                if (e := future.exception()) is not None:
+                    raise e
+
+                coords_and_paths = future.result()
+                completed_data_vars = download_var_group_futures[future]
+
+                max_lead_times = get_max_lead_times(coords_and_paths)
+
+                # Write variable by variable to avoid blowing up memory usage
+                for data_var in completed_data_vars:
+                    data_array, data_array_template = create_data_array_and_template(
+                        chunk_template_ds, data_var.name, shared_buffer
+                    )
+                    var_coords_and_paths = filter_coords_and_paths(
+                        data_var, coords_and_paths
+                    )
+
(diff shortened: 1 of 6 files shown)


src/reformatters/noaa/hrrr/forecast_48_hour/template.py
```diff
@@ -71,23 +109,22 @@ def derive_coordinates(
...
         hrrrds_res[0],
-        hrrrds_res[1] * -1,
-    )  # y offset is positive, but reported as negative by `rio.resolution`
-    # this may be a bug in how rioxarray is handling bounds that cross a central parallel
-    # but also I'd expect that to be the case with lat/lon coordinates too, so I'm not sure.
+        hrrrds_res[1],
+    )
 
-    proj_xcorner, proj_ycorner = hrrrds_bounds[0], hrrrds_bounds[1]
+    proj_xcorner, proj_ycorner = hrrrds_bounds[0], hrrrds_bounds[3]
```
acannistra, line 127: This is the main bug I fixed here. I was using the wrong corner, which explains the need to offset the resolution (which i reverted)

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -22,7 +22,7 @@
 DATASET_ID = "noaa-hrrr-forecast-48-hour"
 DATASET_VERSION = "0.0.0"
 
-INIT_TIME_START = pd.Timestamp("2014-07-30T18:00")
+INIT_TIME_START = pd.Timestamp("2018-07-13T12:00")  # start of HRRR v3
```
acannistra, line 25: I tried starting at v2, but there were lots of missing `.idx` file errors. Didn't dig in much past that.

src/reformatters/noaa/hrrr/read_data.py
```diff
@@ -114,3 +143,76 @@ def parse_index_byte_ranges(
...
+
+    # Source file coords contain the domain and file_type, which are not coords in the data array,
+    # so we drop them from the coords.
+    out_coords = dict(coords)
+    del out_coords["domain"]
+    del out_coords["file_type"]
+
+    # TODO: Why is the shape of the data not the same as the shape of the data array?
+    # out.rio.shape and out.shape are swapped. Need to track this down.
+    out.loc[out_coords] = raw_data.T
```
acannistra, line 180: This is concerning. Need to figure out why I need to transpose this to get it to fit.

aldenks:
hey! im so sorry i missed this, feel free to ping me if its been more than 2 days. I will review by tomorrow, hopefully sooner

src/reformatters/noaa/hrrr/forecast_48_hour/reformat_internals.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+    )
+
+
+def apply_data_transformations_inplace(
+    data_array: xr.DataArray, data_var: DataVar[Any]
+) -> None:
+    if data_var.internal_attrs.deaccumulate_to_rates:
+        raise NotImplementedError("Deaccumulation not implemented for HRRR")
+        # TODO: build deaccumulation?
```
aldenks, line 186: I'm pretty sure our existing deaccumulation code is flexible enough to handle how hrrr does it but good call on pushing that off.

src/reformatters/noaa/hrrr/forecast_48_hour/template.py
aldenks: nice!

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
aldenks: I'm very pro "initial release with shorter archive, then build backwards" so agreed this is good

src/reformatters/noaa/hrrr/read_data.py
```diff
@@ -15,21 +22,39 @@
...
-    hrrr_file_type: HRRRFileType,
     idx_data_vars: Iterable[HRRRDataVar],
 ) -> tuple[SourceFileCoords, Path | None]:
     init_time = coords["init_time"]
     lead_time = coords["lead_time"]
     domain = coords["domain"]
+    file_type = coords["file_type"]
+
+    # Verify that the variables in the group are all from the same file type.
+    # If not, raise (this should be considered a bug).
```
aldenks, line 48: in follow up maybe 
```suggestion
    # If not, raise as this implies a bug earlier in our code.
```

src/reformatters/noaa/hrrr/read_data.py
```diff
@@ -15,21 +22,39 @@
...
+
 
 class SourceFileCoords(TypedDict):
     init_time: pd.Timestamp
     lead_time: pd.Timedelta
     domain: HRRRDomain
+    file_type: HRRRFileType
+
+
+# need to refactor based on the file type being associated with each var
```
aldenks, line 35: Will you say more about this?  As part of a refactor im thinking that data_vars should perhaps be part of SourceFileCoords and i think that aligns with what you're finding here. I ran into this in GEFS reforecast data where they organize source files with a file _per variable_ and forecast. In that case you have one var but multiple lead times in a single file.

What im getting at is the general case may be that SourceFileCoords has a list of init_times, lead_times, etc

src/reformatters/noaa/hrrr/read_data.py
aldenks: Oo I think this might be a dimension ordering thing. if we want north to be up (HRRR's default) and we're representing things in numpy arrays where dim 0 is rows and 1 is cols, then dimension `y` should come before `x` in `DIMS` in template_config.py

aldenks (approved):
Awesome progress @acannistra!

src/reformatters/noaa/hrrr/read_data.py
acannistra: Well, I'm making the fundamental assumption there that we'd like to consider as a part of the scope of this reformatter _all_ variables from the HRRR inventory within the CONUS domain (https://www.nco.ncep.noaa.gov/pmb/products/hrrr/). If so, since these variables are divided among many files (e.g. 50mb geopotential height analysis variable is in a different GRIB from the  whole-atmosphere composite reflectivity analysis variable), so it's necessary to maintain a mapping between the variable and the source file. 

If `data_vars` is a component of `SourceFileCoords`, we would be able to collate those variables which are associated with a given file as input to the request, and we could validate that in each reformatter's downloader.

src/reformatters/noaa/hrrr/read_data.py
acannistra: ok great, that was exactly my hunch too

PR #134, opened 2025-05-15
Refactor pt 3 - DynamicalDataset

aldenks:


src/reformatters/common/dynamical_dataset.py
```diff
@@ -0,0 +1,68 @@
...
+from reformatters.common.types import DatetimeLike
+from reformatters.common.zarr import get_zarr_store
+
+DATA_VAR = TypeVar("DATA_VAR", bound=DataVar[Any])
+SOURCE_FILE_COORD = TypeVar("SOURCE_FILE_COORD", bound=SourceFileCoord)
+
+logger = get_logger(__name__)
+
+
+class DynamicalDataset(pydantic.BaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
```
aldenks: subclass FrozenBaseModel

src/reformatters/common/region_job.py
```diff
@@ -93,9 +93,18 @@ class RegionJob(pydantic.BaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
...
+    # integer slice along append_dim
     region: Annotated[slice, AfterValidator(region_slice)]
     max_vars_per_backfill_job: ClassVar[int]
 
+    @classmethod
+    def group_data_vars(
+        cls,
+        data_vars: Sequence[DATA_VAR],
+    ) -> Sequence[Sequence[DATA_VAR]]:
+        """Return groups of variables, where all variables in a group can be retrieived from the same source file."""
```
```suggestion
        """
        Return groups of variables, where all variables in a group can be retrieived from the same source file.

        This is a class method so it can be called by RegionJob factory methods.
        """
```

src/reformatters/common/region_job.py
```diff
@@ -194,6 +197,96 @@ def summarize_processing_state(
...
+        Parameters
+        ----------
+        store : zarr.abc.store.Store
+            The Zarr store to write into.
+        template_ds : xr.Dataset
+            Dataset template defining structure and metadata.
+        append_dim : AppendDim
+            The dimension along which data is appended (e.g., "time").
+        all_data_vars : Sequence[DATA_VAR]
+            All data variables objects which may be processed in order to access their metadata.
```
aldenks: clarify

PR #137, opened 2025-05-28
Refactor part 4 - Operational updates

aldenks:


src/reformatters/common/region_job.py
```diff
@@ -443,7 +443,12 @@ def _read_into_data_array(
         def _read_and_write_one(coord: SOURCE_FILE_COORD) -> SOURCE_FILE_COORD:
             try:
                 out.loc[coord.out_loc()] = self.read_data(coord, data_var)
-                return replace(coord, status=SourceFileStatus.Succeeded)
+                new_status = (
```
aldenks: I want to undo this change and create a copy of the downloaded source file coords for each variable. That way the read status can correctly represent the read status per variable not per group.

src/reformatters/common/region_job.py
```diff
@@ -443,7 +613,12 @@ def _read_into_data_array(
...
+                    SourceFileStatus.Succeeded
+                    if coord.status == SourceFileStatus.Processing
```
aldenks: I will remove this status logic, because we deepcopy the coords and because we only call _read_and_write_one for successfully downloaded coords, we know the existing status is Processing

src/reformatters/common/dynamical_dataset.py
```diff
@@ -41,6 +47,30 @@ def reformat_kubernetes(self) -> None:
...
+        final_store = self._final_store()
+        tmp_store = self._tmp_store()
+
+        jobs, template_ds = self.region_job_class.operational_update_jobs(
+            final_store=final_store,
+            tmp_store=tmp_store,
+            get_template_fn=self._get_template,
+            append_dim=self.template_config.append_dim,
+            all_data_vars=self.template_config.data_vars,
+            kubernetes_job_name="operational-update",
```
mosegontar, line 61: A bit of a nit, but I wonder if these should be constants or enums  etc since they are checked explicitly and not arbitrary names.

src/reformatters/common/dynamical_dataset.py
```diff
@@ -41,6 +47,30 @@ def reformat_kubernetes(self) -> None:
...
+        template_utils.write_metadata(template_ds, tmp_store, get_mode(tmp_store))
+        for job in jobs:
+            process_results = job.process()
+            updated_template = job.update_template_with_results(process_results)
+            template_utils.write_metadata(
+                updated_template, tmp_store, get_mode(tmp_store)
+            )
+            copy_zarr_metadata(updated_template, tmp_store, final_store)
+
+        logger.info(f"Done operational update writing to {final_store}")
```
mosegontar: nit: this log line is a bit awkward. 

```suggestion
        logger.info(f"Done writing to {final_store}. Operational update complete.")
```

src/reformatters/common/zarr.py
```diff
@@ -118,7 +118,8 @@ def copy_data_var(
     dest = f"{path}/{relative_dir}"
     fsspec_apply(fs, "put", source, dest, recursive=True, auto_mkdir=True)
 
-    progress_tracker.record_completion(data_var_name)
+    if callback is not None:
+        callback()
```
mosegontar: Would it be worth a comment here or in a docstring describing example cases for this callback? If it's always about progress tracking, perhaps just renaming to `progress_tracker_callback`?

src/reformatters/common/dynamical_dataset.py
```diff
@@ -50,24 +80,26 @@ def reformat_local(
...
-        store = self._store()
+        template_ds = self._get_template(append_dim_end)
+        final_store = self._final_store()
 
-        template_utils.write_metadata(template_ds, store, get_mode(store))
+        template_utils.write_metadata(template_ds, final_store, get_mode(final_store))
 
         self.process_region_jobs(
             append_dim_end,
+            kubernetes_job_name="local",
```
mosegontar: Not a huge deal but should we allow `None` here and then set "local" (if an actual name is needed at all) downstream? Even though it's relatively clear from `"local"` I find setting a "kubernetes_job_name" to be a bit awkward, and with less context would even make me wonder a little bit if the string "local" would end up getting used in a kubernetes context. `None` I think is a bit more intuitive that "we're not dealing with kubernetes."

src/reformatters/common/update_progress_tracker.py
```diff
@@ -21,17 +23,19 @@ class UpdateProgressTracker:
 
     def __init__(
         self,
-        store: zarr.storage.FsspecStore,
+        store: zarr.abc.store.Store,
```
mosegontar, line 26: Double checking we don't need `| Path` here

tests/common/region_job_test.py
```diff
@@ -133,19 +137,60 @@ def test_region_job(template_ds: xr.Dataset) -> None:
...
+        append_dim="time",
+        region=slice(0, 18),
+        kubernetes_job_name="test-job",
+    )
+
+    # Mock process results
+    process_results = {
+        "var0": [
+            ExampleSourceFileCoords(
+                time=pd.Timestamp("2025-01-01T12"), status=SourceFileStatus.Succeeded
```
mosegontar, line 158: Nit, worth making these `12:00` and `00:00`?

src/reformatters/common/region_job.py
```diff
@@ -274,7 +384,11 @@ def get_backfill_jobs(
             data_vars = all_data_vars
 
         data_var_groups = cls.source_groups(data_vars)
-        if cls.max_vars_per_backfill_job is not None:
+        # If kind == "operational-update" we need all variables in one job
+        # so we don't extend the dataset before all variables are processed.
+        # If kind == "backfill" it can be useful to make smaller jobs for more parallelism
+        # and granular progress tracking at the kubernetes job level.
+        if kind == "backfill" and cls.max_vars_per_backfill_job is not None:
```
mosegontar, line 393: This line makes me wonder if we want a constant or enum, though it makes working with the `Literal` harder so probably not worth it.

src/reformatters/common/update_progress_tracker.py
aldenks: Yes we only use this in the final store (Paths are just in the tmp stores)

PR #138, opened 2025-05-29
DynamicalDataset cli interface generation

mosegontar:
Draft PR to create a generic method for auto generating a cli interface for a `DynamicalDataset`.

A couple notes/thoughts:

-  To avoid complexity, this doesn't fully try to infer the cli interface from the `DynamicalDataset` interface and it makes explicit which public methods should be exposed to the cli.
- This utility function could arguably be put into `dynamical_dataset.py` itself.
- `Typer` doesn't know what to do with `DatetimeLike`. For the sake of this draft, I switched those to `str`s, but we should probably try to teach typer about this type (specifically that for the cli interface it should expect a str). That said, if the primary ways to interface with the Dataset will be via the `cli`, perhaps this we should just typehint these as `str`s.


src/reformatters/common/cli.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def create_cli(
+    dataset: DynamicalDataset[DATA_VAR, SOURCE_FILE_COORD],
+) -> tuple[str, typer.Typer]:
+    app = typer.Typer()
+    app.command(name="update-template")(dataset.update_template)
+    app.command(name="reformat-local")(dataset.reformat_local)
+    app.command(name="reformat-kubernetes")(dataset.reformat_kubernetes)
+    return dataset.template_config.dataset_id, app
```
mosegontar: Not sure it's actually great to return the id here, but it seems only useful in the cli case

src/reformatters/common/dynamical_dataset.py
```diff
@@ -25,6 +28,11 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
     template_config: TemplateConfig[DATA_VAR]
     region_job_class: type[RegionJob[DATA_VAR, SOURCE_FILE_COORD]]
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def id(self) -> str:
+        return self.template_config.dataset_id
```
mosegontar, line 34: I didn't love having to do `app.add_typer(dataset.create_cli(), name=dataset.template_config.dataset_id)` but it wouldn't be the end of the world if we want to avoid continually promoting this value up the tree.

aldenks (approved):
:taco: !

PR #141, opened 2025-06-09
Refactor pt 5 - reformat_kubernetes, validate_zarr, and operational_kubernetes_resources

aldenks:
The last piece this doesn't implement is sentry cron monitoring on reformat_operational_update and validate_zarr. That'll be a follow up.


aldenks:
I'm going to merge because I want to branch off new work on top and because this only impacts non operational new datasets and deploys (which are easy to see if i've broke em and there's also passing tests for that)

PR #143, opened 2025-06-10
Subclasses override `validators` not `validate_zarr`

aldenks:
This way subclassers don't need to worry about typer annotations nor sentry monitoring. Also a little bit less boilerplate.


mosegontar (approved):
Love it, that's nicer!

PR #144, opened 2025-06-10
Fix fsspec_apply to work with LocalStore and add some tests for UpdateProgressTracker

mosegontar:
This PR makes a fix to `fsspec_apply` to allow it to work with zarr stores that do not have an async implementation.

### DIFF
--- a/src/reformatters/common/fsspec.py
+++ b/src/reformatters/common/fsspec.py
@@ -29,12 +29,14 @@ def fsspec_apply(
     """
     for attempt in range(max_attempts):
         try:
-            if hasattr(fs, f"_{method}"):
+            try:
+                # First try the async method if it exists
                 # Zarr's FsspecStore creates async fsspec filesystems, so use their sync method
                 return zarr.core.sync.sync(
                     getattr(fs, f"_{method}")(*args, **kwargs), timeout=120
                 )
-            else:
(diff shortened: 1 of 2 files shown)


src/reformatters/common/fsspec.py
```diff
@@ -29,12 +29,14 @@ def fsspec_apply(
...
-            if hasattr(fs, f"_{method}"):
+            try:
+                # First try the async method if it exists
                 # Zarr's FsspecStore creates async fsspec filesystems, so use their sync method
                 return zarr.core.sync.sync(
                     getattr(fs, f"_{method}")(*args, **kwargs), timeout=120
                 )
-            else:
+            except (NotImplementedError, AttributeError):
+                # Fall back to the sync method
```
mosegontar: Realized this as I was wrapping up the tests for this: I wonder if we should simplify this to just check if `fs.async_impl` is `True`  and avoid the `try`/`except` here.

aldenks: > The class attribute async_impl can be used to test whether an implementation is async of not.
source: https://filesystem-spec.readthedocs.io/en/latest/async.html#async

That does sound like something we can rely on, that seems maybe a hair nicer, go for it if you want

aldenks (approved):
Approved as is or if you want to change to check async_impl. Thanks!

PR #145, opened 2025-06-10
SWANN dataset integration test fix: Ensure LOCAL_ZARR_STORE_BASE_PATH is monkeypatched to avoid overwriting real final store

mosegontar:
`dataset._final_store()` was resolved to the the "true" final store for my local environment, causing the zarr to be overwritten when the tests run. This PR fixes this by monkeypatching `_LOCAL_ZARR_STORE_BASE_PATH` for all the dataset integration tests.

Broader question: should we configure something at a higher level than this to prevent someone from accidentally overwriting a local zarr?

### DIFF
--- a/tests/u_arizona/swann/dataset_integration_test.py
+++ b/tests/u_arizona/swann/dataset_integration_test.py
@@ -6,10 +6,16 @@ import pytest
 import xarray as xr
 from _pytest.monkeypatch import MonkeyPatch
 
+from reformatters.common import zarr as common_zarr_module
 from reformatters.u_arizona.swann import SWANNDataset
 from reformatters.u_arizona.swann.region_job import SWANNRegionJob, SWANNSourceFileCoord
 
 pytestmark = pytest.mark.slow
+
+
+@pytest.fixture(autouse=True)
+def set_test_final_store(monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
+    monkeypatch.setattr(common_zarr_module, "_LOCAL_ZARR_STORE_BASE_PATH", tmp_path)
 
 
 def test_reformat_local(monkeypatch: MonkeyPatch, tmp_path: Path) -> None:


aldenks (approved):
Looks good!

> Broader question: should we configure something at a higher level than this to prevent someone from accidentally overwriting a local zarr?

These are the two checks we had for this
- [We write to a `dev` version if env != prod](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/zarr.py#L37-L38)
- [We only allow overwrite to a dev.zarr or temp.zarr](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/zarr.py#L86-L89)

Was this a `dev.zarr` or did you somehow get around those checks?

I think it may make sense to add a conditional to get_zarr_store where if env == test we make a tmp dir? then we could avoid that monkeypatch in all tests -- although how would it get cleaned up after the tests unless it was tmp_dir from pytest?

mosegontar:
> Looks good!
> 
> > Broader question: should we configure something at a higher level than this to prevent someone from accidentally overwriting a local zarr?
> 
> These are the two checks we had for this
> 
>    * [We write to a `dev` version if env != prod](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/zarr.py#L37-L38)
> 
>   * [We only allow overwrite to a dev.zarr or temp.zarr](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/zarr.py#L86-L89)
> 
> 
> Was this a `dev.zarr` or did you somehow get around those checks?
> 
> I think it may make sense to add a conditional to get_zarr_store where if env == test we make a tmp dir? then we could avoid that monkeypatch in all tests -- although how would it get cleaned up after the tests unless it was tmp_dir from pytest?

@aldenks  It was the `data/output/u-arizona-swann/vdev.zarr`. I think the the thing that is potentially tricky is that if you run `reformat_local` or `reformat_operational_update` this is the path that is used, and it's also the path in the test env without monkeypatching. 

Some thoughts on possible options:
1. Configure an `autouse=True` fixture like we have in this PR but in `conftest.py` so `_LOCAL_ZARR_STORE_BASE_PATH` is just automatically monkeypatched and will get cleaned up without any further configuration. This is probably my vote.
    - In addition to this, we could add an assert in `get_zarr_store` that if `Config.is_test` the `_LOCAL_ZARR_STORE_BASE_PATH` is not the dev one (or could flip it, it _is_ in `/tmp`). That would probably only end up being a fallback for some weird case where the autouse fixture doesn't take effect or was overwritten.
2. Alternatively, establish a hardcoded path (like we have for the dev env) for `vtest.zarr` and use that when `Config.is_test`. We can configure pytest to automatically clean up between test runs, similar to reset a database between test runs. (I think this works just fine if we're running tests synchronously, but is less fine if we were to ever run them in parallel, so I'm not sure this has any advantages over the first suggestion).

aldenks:
I really like your option 1! great in a follow up or in the PR

mosegontar:
Follow up PR here https://github.com/dynamical-org/reformatters/pull/150

PR #146, opened 2025-06-11
Add SWANN validators

mosegontar:
This PR adds a set of `SWANNDataset.validators`:

- Checks for data within the last 48 hours. (Due to update cadence I've observed, I think a narrower 24 hour window might fail often depending on when we run it -- though we could probably time it: based on the updated at timestamps on https://climate.arizona.edu/data/UA_SWE/DailyData_4km/WY2025/ looks like it is updated at the same time each day).
- Checks that very latest `time` has no more than expected proportion of NaNs. This `time` would likely be the one we've just ingested during an operational run.
- Checks that some other random time within the last year also has no more than the expected amount of NaNs. This extra might be redundant with the first, but given our look back window, this felt reasonable and followed the pattern in the common validators of selecting a random point. 

I manually tested all 3 of these successfully.


src/reformatters/u_arizona/swann/validators.py
```diff
@@ -0,0 +1,71 @@
...
+def check_random_time_within_last_year_nans(
+    ds: xr.Dataset,
+) -> validation.ValidationResult:
+    """
+    Check that the data does not have more than the expected proportion of NaNs
+    for a random time in the last year, as we pull a years worth of data for each operational
+    update of the dataset.
+    """
+    random_time_index = np.random.choice(365)
+    sample_ds = ds.isel(time=slice(random_time_index, random_time_index + 1))
```
aldenks: I think this will only select from the first year of the dataset?
```suggestion
    random_time_index = np.random.choice(365) + 1
    sample_ds = ds.isel(time=-random_time_index)
```

aldenks (approved):
Approved after comment!

src/reformatters/u_arizona/swann/validators.py
mosegontar: Oops nice catch :pray:

PR #148, opened 2025-06-12
Dataset integration guide + template integration

aldenks:
`cp -r` and get rockin


docs/dataset_integration_guide.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Copy template
+From a terminal in the root of this repository, run these commands to copy a template of dataset integration code and tests.
+```bash
+DATASET_ID = "<provider>-<model>-<variant>"
+DATASET_PATH = "<provider>/<model>/<variant>"
+mkdir -p src/reformatters/$DATASET_PATH
+cp -r src/reformatters/example/ src/reformatters/$DATASET_PATH
+mkdir -p tests/$DATASET_PATH
+cp -r tests/example tests/$DATASET_PATH
```
mosegontar, line 32: Nice. It's a little tempting to build this into the cli, e.g., `uv run main init-new-dataset` (kinda like the `rails g` stuff)

For valid `bash` I think you need to close the spaces around `=`. For example:

```
$ bash                                                                                                                                                                             17:04:10
alex@amg:~/repos/dynamical/reformatters$ DATASET_PATH = "<provider>/<model>/<variant>"
DATASET_PATH: command not found
```

But even that won't work with fish and other shells, though they're less common.

I started this comment thinking the little snippet is probably fine. But as long as we're saying "run these commands" I wonder if we should make this a bit easier and handle it in the python via cli.

mosegontar: Or maybe just a lil bash script you invoke with sh? Or make?

mosegontar: Also should this default to copying into the contrib path?

aldenks: good catch! and nice, ya i like that, if not so much around saving time (the frequency here should be high but not _that_ high) but helping us be consistent/not make mistakes. while we're at it it could do the renames in the section below too.

aldenks:
@mosegontar im going to get this on main since its low risk and i don't want it hanging while im OOO tomorrow but i would still love your feedback and will incorporate in a follow up

docs/dataset_integration_guide.md
```diff
@@ -0,0 +1,120 @@
...
+    ...,
+    ProviderModelVariantDataset(),
+]
+```
+
+### 3. Implement `TemplateConfig` subclass
+
+Work through `src/reformatters/$DATASET_PATH/template_config.py`, setting the attributes and method definitions to describe the structure of your dataset.
+
+Hint: providing an AI/LLM with 1) the example template config code to edit, 2) output of running `gdal-info <example source data file>` and 3) any dataset documentation will help it give you a decent first implementation of your `TemplateConfig` subclass.
```
mosegontar, line 65: Typo `gdal-info` should be `gdalinfo`

mosegontar, line 65: Should mention the need to run `ds.rio.write_crs(expected_crs).spatial_ref.attrs` here?

This is a bit of an awkward step in this process. Part of the awkwardness is related to the fact that the `attrs` output is a `dict` but we build `CoordinateAttrs` with kwargs.

Perhaps it'd worth a convention of  doing:

```
            Coordinate(
                name="spatial_ref",
                encoding=Encoding(
                    dtype="int64",
                    fill_value=0,
                    chunks=(),
                    shards=None,
                ),
                attrs=CoordinateAttrs(
                    units=None,
                    statistics_approximate=None,
                    # The following were obtained by running ds.rio.write_crs("EPSG:4269"), see test_get_template_spatial_ref.
                    **{
                          "crs_wkt": ...,
                          "semi_major_axis": ....
                          ...
                     },
                     comment="..."
                ),
            ),
 ```

That way we can instruct to just paste the output in and splat (of course they could do this anyway, but there may be a sense that one ought to follow existing conventions and patterns)

File: docs/dataset_integration_guide.md [hunk 3]
```diff
@@ -0,0 +1,120 @@
...
+Tracking the template in git lets us review diffs of any changes to the structure of our dataset.
+
+Run the tests, making any changes necessary.
+```bash
+uv run pytest tests/$DATASET_PATH/template_config_test.py
+```
+
+### 4. Implement `RegionJob` subclass
+
+Work through `src/reformatters/$DATASET_PATH/region_job.py`, implementing the attributes and method definitions based on the unique structure and processing required for your dataset.
```
mosegontar, line 81: Maybe a small thing, but especially if we are init-ing a new dataset via the cli and there's no instructions to set env variables, maybe we should write this as `<dataset-path>` and `<dataset-id>` to the implication that one should set environment variables somewhere.

src/reformatters/example/dynamical_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+        raise NotImplementedError(
+            f"Implement `operational_kubernetes_resources` on {self.__class__.__name__}"
+        )
+
+    def validators(self) -> Sequence[validation.DataValidator]:
+        """Return a sequence of DataValidators to run on this dataset."""
+        # return (
+        #     validation.check_analysis_current_data,
+        #     validation.check_analysis_recent_nans,
+        # )
```
mosegontar, line 48: I wonder if we should avoid suggesting using these validators or pulling from common vs the SWANN approach of having a dataset specific `validators.py` where these are defined. The validations in `common` didn't really fit for SWANN. Should we add `validators` to our `example/` directory?

mosegontar:
These docs are great. Does make wonder if "tutorial" blog post on `dynamical.org`  would be interesting and helpful? Would probably follow these docs primarily but would be nice to see one proceed through an actual case. Could be a marketing opp too?

src/reformatters/example/region_job.py
```diff
@@ -0,0 +1,284 @@
...
+    #         grouped[data_var.internal_attrs.file_type].append(data_var)
+    #     return list(grouped.values())
+
+    # Implement this method only if specific post processing in this dataset
+    # requires data from outside the region defined by self.region,
+    # e.g. for deaccumulation or interpolation along append_dim in an analysis dataset.
+    #
+    # def get_processing_region(self) -> slice:
+    #     """
+    #     Return a slice of integer offsets into self.template_ds along self.append_dim that identifies
```
mosegontar, line 81: I'm not sure the best place to put, but throughout we give examples and talk about working with `template_ds` but don't explicitly talk about what it is. Maybe it's intuitive? It does make me wonder if we want a paragraph talking about the highlevel design of `reformatters` -- 

-- How we build a template config that describes the dimensions/coordinates/variables/attributes of the dataset (you already touch on this at the beginning of the doc referencing "structure"). We can then load this template as an xarray Dataset (the `template_ds`).
-- How an "update" to the dataset is an modification of that structure (expansion along a particular dimension) 
-- etc

I think we do have references to these things throughout but I wonder if fleshing out the high level summary would help to better contextualize the rest.

docs/dataset_integration_guide.md
```diff
@@ -0,0 +1,120 @@
...
+git add src/reformatters/$DATASET_PATH/templates/latest.zarr
+```
+Tracking the template in git lets us review diffs of any changes to the structure of our dataset.
+
+Run the tests, making any changes necessary.
+```bash
+uv run pytest tests/$DATASET_PATH/template_config_test.py
+```
+
+### 4. Implement `RegionJob` subclass
```
mosegontar, line 79: Might be worth some words here or up in `Words` describing what exactly a `RegionJob` is. The base class doesn't have a docstring either so, I'm not sure it's spelled out anywhere (though probably can be inferred)

src/reformatters/example/dynamical_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+
+from reformatters.common import validation
+from reformatters.common.dynamical_dataset import DynamicalDataset
+from reformatters.common.kubernetes import Job
+
+from .region_job import ExampleRegionJob, ExampleSourceFileCoord
+from .template_config import ExampleDataVar, ExampleTemplateConfig
+
+
+class ExampleDataset(DynamicalDataset[ExampleDataVar, ExampleSourceFileCoord]):
```
mosegontar, line 11: Is this a change from our pattern to put the dataset definition in `__init__.py` (no objections, though if we do this I think we need to correct the example import to add the dataset to DYNAMICAL_DATASETS)

PR #150, opened 2025-06-13
Add autouse fixture to automatically set zarr path to a tmp file in a test env

mosegontar:
This PR:
- Adds a `test` env as possible environment in `config.py`
- Automatically configures tests to run in the `test` env
- Automatically monkeypatches `_LOCAL_ZARR_STORE_BASE_PATH` to the pytest `tmp_path` for the running test
- Removes manual monkeypatching of this constant in individual tests

### DIFF
--- a/src/reformatters/common/zarr.py
+++ b/src/reformatters/common/zarr.py
@@ -35,7 +35,7 @@ BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE = zarr.codecs.BloscCodec(
 
 def get_zarr_store(dataset_id: str, version: str) -> zarr.storage.FsspecStore:
     if not Config.is_prod:
-        version = "dev"
+        version = "dev" if Config.is_dev else version
 
         # This should work, but it gives FileNotFoundError when it should be creating a new zarr.
         # return zarr.storage.FsspecStore.from_url(
--- a/tests/common/test_zarr.py
+++ b/tests/common/test_zarr.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+import pytest
 import zarr
 from pytest import MonkeyPatch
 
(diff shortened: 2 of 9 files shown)


src/reformatters/common/zarr.py
```diff
@@ -35,7 +35,7 @@
 
 def get_zarr_store(dataset_id: str, version: str) -> zarr.storage.FsspecStore:
     if not Config.is_prod:
-        version = "dev"
+        version = "dev" if Config.is_dev else version
```
mosegontar, line 40: I kept the `dev` behavior for consistency. It seemed a bit odd that many of our tests use `get_zarr_path(<dataset-id>, <some-version>)` but then the version is ignored, so I made this change to allow it to pass on through. Alternatively we could just set the version to `Config.env.value` (`dev` or `test`) but again callers passing in a version that is ignored is a bit awkward.

tests/common/test_zarr.py
```diff
@@ -1,12 +1,14 @@
...
 import zarr
 from pytest import MonkeyPatch
 
 from reformatters.common.config import Config, Env
 from reformatters.common.zarr import get_mode, get_zarr_store
 
 
+@pytest.mark.skip_set_test_final_store
 def test_get_zarr_store_dev(monkeypatch: MonkeyPatch) -> None:
     monkeypatch.setattr(Config, "env", Env.dev)
```
mosegontar, line 13: This test really seemed like it should be the exception and not use our monkeypatching, since we want to assert things about the `dev` environment. I didn't find a better way to skip the `autouse` fixture besides this `pytest.mark` trick.

aldenks:
Nice! Will we need to change this given #152 ?

src/reformatters/common/zarr.py
aldenks: this seems to go me

tests/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+
+# This needs to run before any application imports to ensure that
+# Config.env is set to test.
+os.environ["DYNAMICAL_ENV"] = "test"
+
+from reformatters.common import zarr as common_zarr_module
+
+
+@pytest.fixture(autouse=True)
+def set_test_final_store(
```
aldenks: i would name things something about the local store base path to be more reflective of what it technically does

aldenks (approved):
approved after comments!

mosegontar:
> Nice! Will we need to change this given #152 ?

Just needed to resolve some small merge conflicts :)

PR #154, opened 2025-06-17
Run tests on arm too, pin actions

samn:
* Since we deploy arm images we should run tests on it too.
* Pins GitHub Actions to try to prevent vulnerabilities slipping in through the supply chain.
* Improves ruff linting to annotates failures on PRs.
* Checks code format as part of code quality


aldenks (approved):
Awesome thank you

PR #155, opened 2025-06-17
Add contrib direcotry and move UArizona SWANN to contrib/uarizona/swann/analysis

mosegontar:
We want to move the SWANN dataset to a new `contrib/` directory, as it is not a dataset that will be hosted by Dynamical. 

It is also an analysis dataset and, following our new integration guide, should be named as such (and the dataset ID should be updated accordingly).

This PR adds the `contrib` directory, moves SWANN in, and renames accordingly.


aldenks (approved):
Perfecto

PR #156, opened 2025-06-19
GFS Forecast operational updates

aldenks:


src/reformatters/common/region_job.py
```diff
@@ -387,15 +387,23 @@ def get_jobs(
...
         # and granular progress tracking at the kubernetes job level.
         if kind == "backfill" and cls.max_vars_per_backfill_job is not None:
+            # split by source groups first as those are efficient groupings
+            data_var_groups = cls.source_groups(data_vars)
             data_var_groups = cls._maybe_split_groups(
                 data_var_groups, cls.max_vars_per_backfill_job
             )
+        else:
+            # In operational update we must have all variables in one job.
+            # In backfill without max_vars_per_backfill_job set we also want all variables in one job.
```
mosegontar, line 402:
```suggestion
            # In the operational-update case, we must have all variables in one job.
            # In the backfill case when max_vars_per_backfill_job is not set, we also want all variables in one job.
```

tests/noaa/gfs/forecast/dynamical_dataset_test.py
```diff
@@ -5,25 +5,30 @@
...
-    init_time_start = NoaaGfsForecastTemplateConfig().append_dim_start
+
+@pytest.mark.slow
+def test_reformat_local(
+    dataset: NoaaGfsForecastDataset, monkeypatch: MonkeyPatch, tmp_path: Path
+) -> None:
+    init_time_start = dataset.template_config.append_dim_start
     init_time_end = init_time_start + timedelta(hours=12)
 
     monkeypatch.setattr(zarr, "_LOCAL_ZARR_STORE_BASE_PATH", tmp_path)
```
mosegontar: I think this can be removed with the autouse fixture in place?

tests/noaa/gfs/forecast/dynamical_dataset_test.py
```diff
@@ -34,9 +39,9 @@ def test_reformat_local(monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
         filter_variable_names=[
             "temperature_2m",  # instantaneous
             "precipitation_surface",  # accumulation we deaccumulate
-            "precipitable_water_atmosphere",  # average over window
             "maximum_temperature_2m",  # max over window
             "minimum_temperature_2m",  # min over window
+            "categorical_freezing_rain_surface",  # average over window
```
mosegontar: What was the reason for this change in variable name?

aldenks: it was a unique case that (an accumulated variable with average as the accumulation function) that we hadn't covered in the tests and failed in the first test backfill

PR #157, opened 2025-06-20
Set SWANN to write to R2

mosegontar:


src/reformatters/__main__.py
```diff
@@ -30,7 +30,12 @@ class SourceCoopDatasetStorageConfig(DynamicalDatasetStorageConfig):
 # Datasets that have not yet been ported over to the new DynamicalDataset pattern
 # are excluded here until they are refactored.
 DYNAMICAL_DATASETS: Sequence[DynamicalDataset[Any, Any]] = [
-    UarizonaSwannAnalysisDataset(storage_config=SourceCoopDatasetStorageConfig()),
+    UarizonaSwannAnalysisDataset(
+        storage_config=DynamicalDatasetStorageConfig(
+            base_path="s3://upstream-gridded-zarrs",
+            k8s_secret_name="upstream-gridded-zarrs-key",  # noqa: S106
```
aldenks, line 39: I would add a comment that 
- this is actually R2
- the R2 endpoint URL is in the secret and set when it's imported into the env

aldenks (approved):
Approved!

PR #160, opened 2025-06-23
Sentry cron monitoring for DynamicalDatasets

aldenks:
Nice thing about this is it takes all the configuration it needs (monitor slug, check in id aka reformat job name, schedule, timeout) from the already defined operational_kubernetes_resources so there's no extra effort for dataset implementers but it is configured with the unique schedule and run time for each dataset.

TODOs
 - [x] tests
 - [x] confirm it's ok with reformat job name as the check in id or do we need to hash it to make it look like a UUID?

### DIFF
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -1,6 +1,7 @@
 import json
 import subprocess
 from collections.abc import Iterable, Sequence
+from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


aldenks:
I played around with this locally, killing processes mid-way, then re-running them with the same job name and it associates the two together and will successfully complete it. As implemented, it also should not report errors in cases when the pod is terminated by kubernetes, only when there's an exception in our actual code.

We might want to never have it report errors (and just let it time out) because we don't yet have the info to know if this is the last attempt k8s will make to run a pod from this job. But going to see how this works first.

image

PR #163, opened 2025-06-25
Reenable swann operational resources for deployment and stagger validation cron to run an hour later

mosegontar:
Now that https://github.com/dynamical-org/reformatters/pull/156/ has deployed with a fix to ensure k8s secrets are getting passed through to the job, we can reenable this cron.

This PR also pushes back validation to run an hour after the primary worker job runs.

### DIFF
--- a/src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
+++ b/src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
@@ -1,7 +1,9 @@
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
+from datetime import timedelta
 
 from reformatters.common import validation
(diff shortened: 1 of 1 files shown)


src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
```diff
@@ -15,7 +17,7 @@
 )
 
 _OPERATIONAL_CRON_SCHEDULE = "0 0 * * *"
-_VALIDATION_CRON_SCHEDULE = "0 0 * * *"
+_VALIDATION_CRON_SCHEDULE = "0 1 * * *"
```
aldenks: we don't need these in constants anymore, they can be defined inline in the operational_kubernetes_resources. Our sentry integration will pull them from there.

aldenks (approved):

PR #165, opened 2025-06-26
SWANN: Floor Timestamp.now() for latest data validator; run crons at EOD instead of start of day

mosegontar:
Our SWANN latest data validator check failed because no data was found within 48 hours from `pd.Timestamp.now()`. Because all times in this dataset are set to start of day, and the cron was configured to run at `1Z`, it looked for data since `2025-06-24T01:00` and did not count the `2025-06-24T00:00` entry.

This PR floors the `now()` value, so that we are look for data within 48 hours of start-of-day.

The cron is also adjusted to run at the end of the day rather than the start of the day.

This mean's that today's cron (which will run at 2025-06-26T23:00)  will look for entries for the 24th or 25th and fail if either are missing, which is what we want here.

### DIFF
--- a/src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
+++ b/src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
@@ -37,7 +37,7 @@ class UarizonaSwannAnalysisDataset(
     def operational_kubernetes_resources(self, image_tag: str) -> Iterable[CronJob]:
         operational_update_cron_job = ReformatCronJob(
             name=f"{self.dataset_id}-operational-update",
-            schedule="0 0 * * *",
+            schedule="0 22 * * *",
             pod_active_deadline=timedelta(minutes=30),
             image=image_tag,
             dataset_id=self.dataset_id,
@@ -49,7 +49,7 @@ class UarizonaSwannAnalysisDataset(
         )
         validation_cron_job = ValidationCronJob(
             name=f"{self.dataset_id}-validation",
-            schedule="0 1 * * *",
+            schedule="0 23 * * *",
             pod_active_deadline=timedelta(minutes=10),
             image=image_tag,
             dataset_id=self.dataset_id,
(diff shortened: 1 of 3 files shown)


src/reformatters/contrib/uarizona/swann/analysis/validators.py
```diff
@@ -15,7 +15,9 @@ def check_data_is_current(ds: xr.Dataset) -> validation.ValidationResult:
     """
     Check that the data is current within the last 48 hours.
     """
-    now = pd.Timestamp.now()
+    # All times in the dataset are set to start of day, so we need to check
+    # that there is `time` that is within 48 hours from start of day.
+    now = pd.Timestamp.now().floor("D")
     latest_init_time_ds = ds.sel(time=slice(now - pd.Timedelta(hours=48), None))
```
aldenks:
```suggestion
    today_start = pd.Timestamp.now().floor("D")
    latest_init_time_ds = ds.sel(time=slice(today_start - pd.Timedelta(hours=48), None))
```

src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
```diff
@@ -37,7 +37,7 @@ def validators(self) -> Sequence[validation.DataValidator]:
     def operational_kubernetes_resources(self, image_tag: str) -> Iterable[CronJob]:
         operational_update_cron_job = ReformatCronJob(
             name=f"{self.dataset_id}-operational-update",
-            schedule="0 0 * * *",
+            schedule="0 22 * * *",
```
aldenks: Looking at the Last Modified headers when downloading a file (and the html listings, but those are in AZ time) the files are published right before 15:05 UTC every day, with one exception where it took until 16:15 UTC. Giving 15 more mins buffer,

```suggestion
            schedule="30 16 * * *",
```

src/reformatters/contrib/uarizona/swann/analysis/dynamical_dataset.py
```diff
@@ -49,7 +49,7 @@ def operational_kubernetes_resources(self, image_tag: str) -> Iterable[CronJob]:
         )
         validation_cron_job = ValidationCronJob(
             name=f"{self.dataset_id}-validation",
-            schedule="0 1 * * *",
+            schedule="0 23 * * *",
```
aldenks:
```suggestion
            schedule="0 17 * * *",
```

aldenks (approved):
Approved after comments!

PR #166, opened 2025-06-27
CLI Rename

mosegontar:
```
| Existing Name                | New Name                      |
|------------------------------|-------------------------------|
| update_template              | no change                     |
| reformat_operational_update  | update                        |
| reformat_kubernetes          | backfill_kubernetes           |
| reformat_local               | backfill_local                |
| process_region_jobs          | process_backfill_region_jobs  |
| validate_zarr                | validate (method name is validate_dataset)|
```

image


src/reformatters/common/dynamical_dataset.py
```diff
@@ -313,10 +316,10 @@ def get_cli(
...
         app.command()(self.update_template)
-        app.command()(self.reformat_operational_update)
-        app.command()(self.reformat_kubernetes)
-        app.command()(self.reformat_local)
-        app.command()(self.process_region_jobs)
+        app.command()(self.update)
+        app.command()(self.backfill_kubernetes)
+        app.command()(self.backfill_local)
+        app.command()(self.process_backfill_region_jobs)
         app.command()(self.validate_zarr)
```
aldenks: What about keeping the one verb symmetry and renaming the command to `validate` https://typer.tiangolo.com/tutorial/commands/name/

And then making the function name be `validate_dataset` (thinking about an icechunk future) to avoid collision with pydantic's validate.

aldenks (approved):
Approved after last comment!

src/reformatters/common/dynamical_dataset.py
mosegontar: Oh nice yeah that's a good call, will make those changes

mosegontar: updated 
image

PR #172, opened 2025-07-02
Contrib: NOAA NDVI CDR Analysis dataset

mosegontar:
https://www.ncei.noaa.gov/products/climate-data-records/normalized-difference-vegetation-index


src/reformatters/common/config_models.py
```diff
@@ -79,7 +79,7 @@ class Encoding(pydantic.BaseModel):
 
     # Could be any np.typing.DTypeLike but that type is loose and allows any string.
     # It's fine to add any valid dtype string to this literal.
-    dtype: Literal["float32", "float64", "uint16", "int64", "bool"]
```
mosegontar, line 82: This should probably be int16 and we should figure out how to handle the mask. This int32 was left over from a previous attempt and needs to be changed.

src/reformatters/common/region_job.py
```diff
@@ -506,30 +510,33 @@ def process(self) -> Mapping[str, Sequence[SOURCE_FILE_COORD]]:
                 # Process one data variable at a time to ensure a single user of
                 # the shared buffer at a time and to reduce peak memory usage
                 for data_var in data_var_group:
+                    print("Beginning processing", data_var.name)
```
mosegontar, line 513: Will remove debug prints

src/reformatters/contrib/uarizona/swann/analysis/template_config.py
```diff
@@ -182,7 +182,7 @@ def coords(self) -> Sequence[Coordinate]:
     @computed_field  # type: ignore[prop-decorator]
     @property
     def data_vars(self) -> Sequence[UarizonaSwannDataVar]:
-        # Chunking selected to target ~1.5mb compressed chunks (we are assuming ~20% compression)
+        # Chunking selected to target ~1.5mb uncompressed chunks
```
mosegontar, line 185: Typo, ~1.5 was the uncompressed calculation

src/reformatters/common/region_job.py
```diff
@@ -508,11 +513,11 @@ def process(self) -> Mapping[str, Sequence[SOURCE_FILE_COORD]]:
                 for data_var in data_var_group:
                     # Copy so we have a unique status per variable, not per variable group
                     data_var_source_file_coords = deepcopy(source_file_coords)
-
                     data_array, data_array_template = create_data_array_and_template(
                         processing_region_ds,
                         data_var.name,
                         shared_buffer,
+                        fill_value=data_var.encoding.fill_value,
```
mosegontar, line 520: @aldenks this one slipped my mind when you asked what changes I needed to make to deal with the int16 data var, so calling it out here

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,241 @@
...
+            for filepath in available_files_by_year[year]:
+                _, date, _ = filepath.rsplit("_", 2)
+                if year_date_month_str == date:
+                    filename = Path(filepath).name
+                    url = f"{public_base_url}/{year}/{filename}"
+                    return url
+            # TODO: Should we rethink this? This kind of preemptively handles the dataset truncation
+            # which would otherwise happen during operational update. It means we fail if our append_dim end date
+            # is too far in the future.
+            raise ValueError(f"No file found for {time}")
```
mosegontar: Data is available on a ~3 day lag from what i can tell, so I think we definitely want to change this behavior here, otherwise we'll always blow up. I'm still thinking through how to handle this

src/reformatters/common/shared_memory_utils.py
```diff
@@ -96,9 +98,9 @@ def create_data_array_and_template(
         buffer=shared_buffer.buf,
     )
     # Important:
-    # We rely on initializing with nans so failed reads (eg. corrupt source data)
-    # leave nan and to reuse the same shared buffer for each variable.
-    shared_array[:] = np.nan
+    # We rely on initializing with a fill value so failed reads (eg. corrupt source data)
+    # leave the fill value and to reuse the same shared buffer for each variable.
+    shared_array[:] = fill_value
```
mosegontar, line 103: Other half the change i mentioned in the other comment to handle int16s which can't have nans.

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,275 @@
...
+        longitude = origin_lon + (np.arange(7200) + 0.5) * pixel_size_lon
+        latitude = origin_lat + (np.arange(3600) + 0.5) * pixel_size_lat
+
+        return {
+            self.append_dim: self.append_dim_coordinates(
+                self.append_dim_start + self.append_dim_frequency
+            ),
+            "latitude": latitude,
+            "longitude": longitude,
+        }
```
mosegontar: As with SWANN, the `netcdf` files did not have a consistent delta. 

I've noticed, probably due to the high resolution, some mismatches when selecting into the zarr data vs the netcdf: 

```

In [39]: nc.sel(latitude=-37.83, longitude=141.4, method="nearest").NDVI.values
Out[39]: array([6235], dtype=int16)

In [40]: ds.isel(time=1).sel(latitude=-37.83, longitude=141.4, method="nearest").ndvi_raw.values
Out[40]: array(0.9277344, dtype=float32)

In [41]: ds.isel(time=1).sel(latitude=-37.83, longitude=141.4, method="nearest").time.values
Out[41]: np.datetime64('1981-06-25T00:00:00.000000000')

In [42]: nc.sel(latitude=-37.83, longitude=141.4, method="nearest").time.values
Out[42]: array(['1981-06-25T00:00:00.000000000'], dtype='datetime64[ns]')

In [43]: ds.sel(latitude=-37.83, longitude=141.4, method="nearest").longitude.values.tolist()
Out[43]: 141.3750047944719

In [44]: nc.sel(latitude=-37.83, longitude=141.4, method="nearest").longitude.values.tolist()
Out[44]: 141.42498779296875
```

I'm not sure how problematic that is, but want to flag it

src/reformatters/common/region_job.py
```diff
@@ -162,7 +167,7 @@ def read_data(
         self,
         coord: SOURCE_FILE_COORD,
         data_var: DATA_VAR,
-    ) -> ArrayFloat32:
+    ) -> ArrayFloat32 | ArrayInt16:
```
mosegontar: Maybe a bit weird to add these as needed, perhaps we should hint with a more generic type

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
```diff
@@ -0,0 +1,55 @@
...
+        """Return the kubernetes cron job definitions to operationally update and validate this dataset."""
+        # operational_update_cron_job = ReformatCronJob(
+        #    name=f"{self.dataset_id}-operational-update",
+        #    schedule="0 20 * * *",
+        #    pod_active_deadline=timedelta(minutes=30),
+        #    image=image_tag,
+        #    dataset_id=self.dataset_id,
+        #    cpu="2",
+        #    memory="200G",
+        #    shared_memory="190Gi",
```
mosegontar, line 30: `shared_buffer` max size is 189Gi.

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
mosegontar: Handled this with a fake url when none is found with the expectation that this will intentionally blow up as if the URL 404s https://github.com/dynamical-org/reformatters/pull/172/commits/97cd3b920a1dabba5cd2ac9ab8d4d09410ac5c9d. A bit hacky, can walk back (wasn't sure if it made sense to not generate some source file coords after setting an `append_dim_end` value an if that'd lead to downstream problems).

src/reformatters/common/region_job.py
```diff
@@ -123,6 +124,10 @@ class RegionJob(pydantic.BaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
     # This particularly useful of the data source cannot handle a large number of concurrent requests
     download_parallelism: int = (os.cpu_count() or 1) * 2
 
+    # Subclasses can override this to control read parallelism
+    # This is useful in cases where many threads using reading data into shared memory cause deadlocks
```
```suggestion
    # This is useful in cases where many threads reading data
    # into shared memory causes deadlocks or resource contention
```

src/reformatters/common/region_job.py
aldenks: yeah unless we got super fancy and somehow pull out the element type from the DATA_VAR (I don't think we should try), lets do

```suggestion
    ) -> np.ndarray:
```

src/reformatters/common/shared_memory_utils.py
```diff
@@ -47,6 +48,7 @@ def create_data_array_and_template(
     processing_region_ds: xr.Dataset,
     data_var_name: str,
     shared_buffer: SharedMemory,
+    fill_value: float | int | bool = np.nan,
```
aldenks: I think this should not have a default and should be a required arg

src/reformatters/common/types.py
```diff
@@ -15,6 +15,7 @@
     pydantic.PlainValidator(pd.Timedelta),
 ]
 type ArrayFloat32 = np.ndarray[tuple[int, ...], np.dtype[np.float32]]
+type ArrayInt16 = np.ndarray[tuple[int, ...], np.dtype[np.int16]]
```
aldenks, line 18: we might not need this anymore

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
aldenks: you might as well set memory=250G or adjust shard size to fit everything in 128G. The actual instances sizes available in memory optimized families around this are:

r8g.4xlarge - cpu: 16  memory: 128
r8g.8xlarge - cpu: 32  memory: 256

either one is fine. looks like the r8g.8xlarge is ~$0.55/hour -- if this takes < 30 mins to update once a day that's negligible.

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/quality_flags.py
```diff
@@ -0,0 +1,122 @@
...
+AVHRR_RHO3_INVALID = 1 << 13
+AVHRR_BRDF_CORR_PROBLEM = 1 << 14
+AVHRR_POLAR_FLAG = 1 << 15  # Appears as -32768 in int16
+
+# Common values
+FILL_VALUE = -32767  # Common fill value for both AVHRR and VIIRS
+
+
+def get_avhrr_mask(qa_array: Array2D[np.int16]) -> Array2D[np.bool_]:
+    """Generate bad quality mask for AVHRR data.
```
aldenks:
```suggestion
    """
    Generate "bad" quality mask for AVHRR data that is true where values should be ignored.
    
    The flags we use and don't use here come from a practical balance of maintaining 
    enough decent quality data while removing values that clearly disrupt the NDVI signal.
```

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/quality_flags.py
```diff
@@ -0,0 +1,122 @@
...
+    Masks pixels with cloud, shadow, water, sunglint, or invalid channels.
+    Polar flag is NOT considered bad quality.
+
+    True in the return means bad quality.
+    False in the return means good quality.
+    """
+    # Identify fill values before conversion
+    is_fill = qa_array == FILL_VALUE
+
+    # Quality flags are stored as int16 in NetCDF but should be interpreted as unsigned
```
aldenks: I think i sent you wrong here, signed or unsigned works equally well. the one exception i see to that is the polar flag because it has all bits set (including the most significant bit which makes it a negative signed int):
- if you do `np.array(1 << 15, np.int16)` that will fail with `OverflowError: Python integer 32768 out of bounds for int16` because the shift is happening in a python integer that doesn't overflow and then it tries to fit that number thats one too large for an int16 into an int16.
- but if you bitshift a numpy int16 then it works: `np.array(1, np.int16) << 15 == np.int16(-32768)`

I think probably least confusing/simplest change is to change dtype=unit16 to int16 throughout this function and the one for VIIRS, remove the casts, and define the polar flag as `np.int16(1) << 15`

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+    """Coordinates of a single source file to process."""
+
+    time: Timestamp
+    url: str
+
+    def get_url(self) -> str:
+        return self.url
+
+    def out_loc(
+        self,
```
aldenks: might go all on one line if you take the , off

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+    ) -> Mapping[Dim, CoordinateValueOrRange]:
+        return {"time": self.time}
+
+
+class NoaaNdviCdrAnalysisRegionJob(
+    RegionJob[NoaaNdviCdrDataVar, NoaaNdviCdrAnalysisSourceFileCoord]
+):
+    download_parallelism: int = 2
```
aldenks: I think you can probably remove this, the data is on s3 right? might have just been your local internet issues

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+    def generate_source_file_coords(
+        self,
+        processing_region_ds: xr.Dataset,
+        data_var_group: Sequence[NoaaNdviCdrDataVar],
+    ) -> Sequence[NoaaNdviCdrAnalysisSourceFileCoord]:
+        """Return a sequence of coords, one for each source file required to process the data covered by processing_region_ds.
+
+        The names for this dataset include the processing timestamp, which means that we cannot determine the filename and URL
+        ourselves. When generating SourceFileCoords we need to enumerate the available files and match them to each time we are
+        are attempting to process.
```
aldenks:
```suggestion
        The source file names include the processing timestamp, which means that we cannot determine the filename and URL
        just from a timestamp. When generating SourceFileCoords we need to enumerate the available files and match them to each time we are
        are attempting to process.
```

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+        years = {pd.Timestamp(t).year for t in times}
+        available_files_by_year = {
+            year: fs.ls(f"noaa-cdr-ndvi-pds/data/{year}") for year in years
+        }
+
+        def _get_url(time: Timestamp) -> str:
+            timestamp = pd.Timestamp(time)
+            year = timestamp.year
+            year_date_month_str = timestamp.strftime("%Y%m%d")
+            for filepath in available_files_by_year[year]:
```
aldenks: is this parsing all timestamps for each time to match (`O(N^2)`) slow? we could change the algo so it lists and parses the timestamp for all files in all matching years once into a dict of `urls_by_time = {timestamp: url}` and then `[NoaaNdviCdrAnalysisSourceFileCoord(time=t, url=urls_by_time[t]) for t in times if t in urls_by_time]`

that said, the listing the files is probably the slow thing and it doesn't matter

aldenks: oh the other benefit of that is we can skip the "no-url" logic and only return coordinates we know have good urls

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+    def _read_netcdf_data(
+        self,
+        coord: NoaaNdviCdrAnalysisSourceFileCoord,
+        *,
+        var_name: str,
+        dtype: str,
+        netcdf_fill_value: float | int,
+        encoding_fill_value: float | int,
+        scale_factor: float | None = None,
+        add_offset: float | None = None,
```
aldenks: you probably didn't do this for a reason, but why not have this take one DataVar instead of these 6 arguments? it feels ok to me to do something like the following so you always have a data var to pass

```
from reformatters.common.iterating import item
from .template_config import NoaaNdviCdrAnalysisTemplateConfig

QA_DATA_VAR = item(dv for dv in NoaaNdviCdrAnalysisTemplateConfig().data_vars() if dv.name == "qa")
```

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,275 @@
...
+from reformatters.common.zarr import (
+    BLOSC_2BYTE_ZSTD_LEVEL3_SHUFFLE,  # noqa: F401
+    BLOSC_4BYTE_ZSTD_LEVEL3_SHUFFLE,  # noqa: F401
+    BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE,  # noqa: F401
+)
+
+# Constants for QA data variable
+QA_NETCDF_VAR_NAME = "QA"
+QA_FILL_VALUE = -32767
+QA_ENCODING_FILL_VALUE = -32767
```
aldenks, line 33: import from quality_flags.py?

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,275 @@
...
+
+    netcdf_var_name: str
+    fill_value: float
+    scale_factor: float | None = None
+    add_offset: float | None = None
+    valid_range: tuple[float, float] | None = None
+
+
+class NoaaNdviCdrDataVar(DataVar[NoaaNdviCdrInternalAttrs]):
+    attrs: DataVarAttrs
```
aldenks:
```suggestion
    pass
```
attrs is already defined like this on the parent class

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
aldenks: Good catch and yes we do need to look into this. We might be off by 1px: the difference between the longitude values in the last two rows is 0.0499 (basically the dataset pixel resolution). The scaled NDVI value is .62 vs .92 which is definitely different in the NDVI scale of roughly 0 to 1.

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,275 @@
...
+                    comment="EPSG:4326 - WGS 84 Geographic Coordinate System",
+                ),
+            ),
+        ]
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def data_vars(self) -> Sequence[NoaaNdviCdrDataVar]:
+        """Define metadata and encoding for each data variable."""
+        # Chunking selected to target ~1mb uncompressed chunks
```
aldenks:
```suggestion
        # Chunking selected to target ~3mb compressed chunks
```

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,275 @@
...
+        var_chunks: dict[Dim, int] = {
+            "time": 365,
+            "latitude": 100,
+            "longitude": 100,
+        }
+        # Sharding selected to target ~300mb compressed shards (we are assuming ~20% compression)
```
aldenks:
```suggestion
        # Sharding selected to target ~350mb compressed shards (assuming compression to ~20%)
```

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/validators.py
```diff
@@ -0,0 +1,50 @@
+import pandas as pd
+import xarray as xr
+
+from reformatters.common import validation
+
```
aldenks, line 5: With a little bit of generalization of the existing validators in validation.py we could skip this whole file. if you add kwargs for each of the parameters (append_time, window, var_name(s), and nan_fraction) then each dataset can use them like 

```
from functools import partial

def validators(self):
    return (
        partial(check_analysis_current_data, window=pd.Timedelta(days=5), append_dim=...)
        ...
    )
```

src/reformatters/example/dynamical_dataset.py
```diff
@@ -19,7 +19,7 @@ def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
         #     schedule=_OPERATIONAL_CRON_SCHEDULE,
         #     pod_active_deadline=timedelta(minutes=30),
         #     image=image_tag,
-        #     dataset"_id=self.dataset_id,
+        #     dataset_id=self.dataset_id,
```
aldenks, line 22: <3

tests/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset_test.py
```diff
@@ -0,0 +1,90 @@
...
+    # This has the cloud flag set, so it gets masked out in the usable data
+    assert np.isnan(
+        ds.sel(latitude=59.98, longitude=105.61, method="nearest").ndvi_usable.item()
+    )
+
+
+def test_update(monkeypatch: MonkeyPatch, tmp_path: Path) -> None:
+    dataset = NoaaNdviCdrAnalysisDataset(storage_config=noop_storage_config)
+    # Dataset starts at 1981-06-24, test with initial data
+    dataset.backfill_local(append_dim_end=pd.Timestamp("1981-06-25"))
```
aldenks: If this is slower than 5 seconds lets combine these two tests into one longer one so we don't duplicate the backfill_local work

aldenks (commented):
Sweet! The lat lon coordinates not matching the source file is the one key thing to look into.

src/reformatters/common/region_job.py
mosegontar: Type checker is unhappy, with just `np.ndarray`, we'd need to do something like `np.ndarray[Any, Any]` .

I've defined
```
type ArrayND[D: np.generic] = np.ndarray[tuple[int, ...], np.dtype[D]]
```

and the common `read_data` return type looks like:

```
    def read_data(
        self,
        coord: SOURCE_FILE_COORD,
        data_var: DATA_VAR,
    ) -> ArrayND[np.generic]:
 ```

`mypy` seems happy, so I'll run with this but lmk if there might be problems with this

File: src/reformatters/common/types.py (hunk 12 above)
Comment (mosegontar) [reply src/reformatters/common/types.py:18]: We use it in the `ndvi` `read_data` stuff still. I think it makes sense for those functions to be explicit about what they might return (float32 or int16) while leaving the common/parent with the more generic type hint. Inclined to leave it but open to objections!

File: src/reformatters/common/shared_memory_utils.py (hunk 11 above)
Comment (mosegontar) [reply src/reformatters/common/shared_memory_utils.py]: :+1: sounds good, looks like the only place we needed to add it to was the `hrrr` call. The other legacy datasets use their own version of the function.

File: tests/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset_test.py (hunk 26 above)
Comment (mosegontar) [reply tests/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset_test.py]: good call, it's quite slow!

Comment (mosegontar) [reply tests/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset_test.py]: We could remove some of the checks in there, since even parsing this data locally is slow -- though i just removed several (so that we only checked `ndvi_usable` at a specific point) and it only reduced things from 50s to 41s. Not sure if that is significant enough to warrant removing the other checks.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/quality_flags.py (hunk 14 above)
Comment (mosegontar) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/quality_flags.py]: wait, actually we're not using the polar flag -- i'll update the definition of the constant to `np.int16(1) << 15` but just for completeness of including all the AVHRR flags. I'll remove all the unint16 stuff in the body of the function.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py (hunk 19 above)
Comment (mosegontar) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py]: Yeah I had started to do basically this the other day, but backed away from it because technically the template config is configurable, so there is potential drift between how it is instantiated in the dynamical dataset class and how we'd do it here.

That said, we've designed things so that you really shouldn't be overriding any of the defaults via instantiation. And I think the values we'll be plucking out of it are not ones that would be impacted by overrides anyway.
I'll make the change along the lines you sketched out here.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py (hunk 20 above)
Comment (mosegontar) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py:33]: I think probably we want to go in the other direction? These are needed in the internal and encodings for the data var, and quality_flags should use that when building up the mask. I'll replace the `FILL_VALUE` constant in `quality_flags` with `QA_FILL_VALUE`.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/validators.py (hunk 24 above)
Comment (mosegontar) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/validators.py:5]: I began going down this path for SWANN along very similar lines (using partial etc) and backed out of it:

`validation.check_analysis_recent_nans` as currently implemented is actually pretty specific in the `sample_ds` it constructs (random lat/lon across the whole globe, 12 hour time slice). For SWANN and NDVI this doesn't work great for a NaN check because there are many points in the data where we expect to always have a NaN. 

You're right we definitely could generalize the function so that a caller can essentially specify how to construct the `sample_ds`, but what I found was it made the set up and call a bit messy and convoluted; it seems like we already have three distinct approaches for constructing the `sample_ds` for validation which makes me a bit skeptical of trying to use a common function for this. In the end, we're really only making use  of the latter half of the common function (where we calculate the nans and instantiate the `ValidationResult` (and even that we may want to customize, as we did with SWANN to distinguish error messages between two different checks)). 

I ended up thinking it's much clearer to explicitly and simply define the checks directly and independently. 

The recent data check is potentially easier to reuse as a common function, but it's a simple enough function to re-implement (and the logic is modified still for the particular case, so it's not straight duplication) that it seemed preferable to put the particular dataset validators in a single place, rather than having a mix of some pulled from `validation.py` and some custom ones.

Happy to reconsider this if you feel strongly or think it'd be worth making a generic validation function to use across datasets.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py (hunk 18 above)
Comment (mosegontar) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py]: It's pretty fast (as you noted it's the s3 lsing that's slow), [i think it's `O(N*M)`? where M is 365/366])  but agreed this is nicer, good call.

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/validators.py (hunk 24 above)
Comment (aldenks) [reply src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/validators.py:5]: Ahh i see, that works for me. And i don't mind having a dataset us a mix of common and dataset specific validators

File: src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py [hunk 27]
```diff
@@ -0,0 +1,223 @@
...
+                )
+        return coords
+
+    def download_file(self, coord: NoaaNdviCdrAnalysisSourceFileCoord) -> Path:
+        """Download the file for the given coordinate and return the local path."""
+        url = coord.get_url()
+        # We set a special value if, when generating the source file coords, we did not find
+        # an available file.
+        if url == "no-url":
+            raise FileNotFoundError(f"No file found for {coord.time}")
```
aldenks: i think you removed this as a possibility

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,223 @@
...
+        The source data for this dataset is stored and available from a variety of sources. See the "Data Access" section of
+        https://www.ncei.noaa.gov/products/climate-data-records/normalized-difference-vegetation-index. We are currently using
+        the NOAA S3 bucket.
+        """
+        # set anon to True to not require AWS credentials, as this is a public bucket.
+        fs = fsspec.filesystem("s3", anon=True)
+        public_base_url = "https://noaa-cdr-ndvi-pds.s3.amazonaws.com/data"
+
+        times = processing_region_ds["time"].values
+        years = {pd.Timestamp(t).year for t in times}
```
aldenks: this is like a "next time" thing, but you can get pandas fancy with this
```suggestion
        times = pd.to_datetime(processing_region_ds["time"])
        years = set(times.year)
```

that also lets you skip re-parsing the timestamp with pd.Timestamp down on line 108

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -0,0 +1,223 @@
...
+        ndvi_data = self._read_netcdf_data(coord, data_var)
+        qa_data = self._read_netcdf_data(coord, QA_DATA_VAR)
+
+        assert qa_data.dtype == np.int16
+        qa_data = cast(Array2D[np.int16], qa_data)
+
+        timestamp = pd.Timestamp(coord.time)
+
+        # Apply quality filtering based on timestamp
+        if timestamp < VIIRS_START_DATE:
```
aldenks:
```suggestion
        # Apply quality filtering based on timestamp
        if coord.time < VIIRS_START_DATE:
```
was this giving you grief? our `Timestamp` type is already a pd.Timestamp, just annotated to play nicer with pydantic

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
```diff
@@ -0,0 +1,276 @@
...
+            time_resolution="Daily",
+            forecast_domain=None,
+            forecast_resolution=None,
+        )
+
+    def dimension_coordinates(self) -> dict[str, Any]:
+        """
+        Returns a dictionary of dimension names to coordinates for the dataset.
+        """
+        # This file was created from the lat/lon values in one of the source NetCDF files:
```
aldenks, line 80: did you save the midpoint of the lat_bnds / lon_bnd data in the nc files?

mosegontar: Hm, here's what I did: 

```

In [3]: nc = xr.open_dataset("./data/download/noaa-ndvi-cdr-analysis/data/1981/AVHRR-Land_v005_AVH13C1_NOAA-07_19810625_c20170610042839.nc", mask_and_scale=False)

In [4]: 

In [4]: lat = nc.latitude.values

In [5]: lon = nc.longitude.values

In [6]: np.savez('source_latitude_longitude_values.npz', latitude=lat, longitude=lon)

```
was this off?

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
mosegontar: good catch, thank you

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
mosegontar: oh smart, easy enough to do (tried, and tests pass)

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
mosegontar: oh no, just overlooking what the type was, fixed

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/template_config.py
aldenks: ahh that must be right. when i read with gdal/rasterio it surfaces lat and lon as `lat_bnds` and `lon_bnds` which are pairs of values (e.g. a 3000x2 array) that i assume are the min and max value for each pixel. looks like xarray knows what's up and gives you pixel _midpoints_ which is what we want.

In this case of their close to regular but actually irregular spacing your npz file approach seems great

aldenks (approved):
wahoo!

PR #178, opened 2025-07-11
Add boto3 to support lsing S3 buckets anonymously

mosegontar:
We had trouble getting the `anon=True` flag to work with `fsspec` and `s3fs`. While there might be a solution, this PR adds `boto3` which works for lsing with unsigned requests. (Note that we needed to use version `~1.36.0` due a bug with later versions. See https://github.com/boto/boto3/issues/4462#issuecomment-2701723847)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,10 @@ dependencies = [
     "requests>=2.32.3",
     "rioxarray>=0.18.2",
     "s3fs>=2024.9.0",
-    "boto3~=1.36.0",
+    # We ran into an issue when installing boto3 1.39.x, related to https://github.com/boto/boto3/issues/4462#issuecomment-2701723847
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
     "requests>=2.32.3",
     "rioxarray>=0.18.2",
     "s3fs>=2024.9.0",
+    "boto3~=1.36.0",
```
aldenks: add comment noting the range of versions we know have the bug

aldenks (approved):

mosegontar:
We also ran into storage issues and download issues. To resolve the storage issues (no space left on device) I've bumped emphemeral storage to 150G. To resolve the download issues, we've dropped parallelism to 4.

mosegontar:
Commenting out operational resources so this doesn't run over the weekend

PR #179, opened 2025-07-11
NDVI: Refactor to ls and download with obstore and operationalize

mosegontar:
This PR:

- Refactors listing and downloading available source files to use `obstore`
- Reduces our shard size to get us on smaller nodes -- we think this reduces our chance of eviction.
- Removes the `boto3` dependency (no longer needed with `obstore`)
- Tweaks the operational resources definitions.

### DIFF
--- a/deploy/aws/nodepool.yaml
+++ b/deploy/aws/nodepool.yaml
@@ -19,7 +19,7 @@ spec:
           values: ["c", "m", "r"]
         - key: "eks.amazonaws.com/instance-cpu"
           operator: Lt
-          values: ["33"]
+          values: ["17"]
         - key: "eks.amazonaws.com/instance-generation"
           operator: Gt
           values: ["6"]
--- a/src/reformatters/common/download.py
+++ b/src/reformatters/common/download.py
@@ -77,6 +77,33 @@ def http_store(base_url: str) -> obstore.store.HTTPStore:
     )
 
 
+@functools.cache
+def s3_store(
+    bucket_url: str, region: str, skip_signature: bool = True
+) -> obstore.store.S3Store:
+    store = obstore.store.from_url(
+        bucket_url,
+        region=region,
+        skip_signature=skip_signature,
+        client_options={
+            "connect_timeout": "4 seconds",
+            "timeout": "120 seconds",
+        },
+        retry_config={
+            "max_retries": 16,
+            "backoff": {
+                "base": 2,
+                "init_backoff": timedelta(seconds=1),
+                "max_backoff": timedelta(seconds=16),
+            },
+            # A backstop, shouldn't hit this with the above backoff settings
+            "retry_timeout": timedelta(minutes=5),
+        },
+    )
+    assert isinstance(store, obstore.store.S3Store)
+    return store
+
+
 def http_download_to_disk(
     url: str,
     dataset_id: str,
--- a/src/reformatters/common/download.py
+++ b/src/reformatters/common/download.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import contextlib
 import functools
 import os
@@ -5,15 +7,19 @@ import uuid
 from collections.abc import Sequence
 from datetime import timedelta
 from pathlib import Path
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 import obstore
+
+if TYPE_CHECKING:
+    from obstore.store import ObjectStore
 
 DOWNLOAD_DIR = Path("data/download/")
 
 
 def download_to_disk(
-    store: obstore.store.ObjectStore,
+    store: ObjectStore,
     path: str,
     local_path: Path,
     *,
--- a/src/reformatters/common/download.py
+++ b/src/reformatters/common/download.py
@@ -13,7 +13,7 @@ DOWNLOAD_DIR = Path("data/download/")
 
 
 def download_to_disk(
-    store: obstore.store.HTTPStore | obstore.store.S3Store,
+    store: obstore.store.ObjectStore,
     path: str,
     local_path: Path,
     *,
--- a/src/reformatters/common/download.py
+++ b/src/reformatters/common/download.py
@@ -13,7 +13,7 @@ DOWNLOAD_DIR = Path("data/download/")
 
 
 def download_to_disk(
-    store: obstore.store.HTTPStore,
+    store: obstore.store.HTTPStore | obstore.store.S3Store,
     path: str,
     local_path: Path,
     *,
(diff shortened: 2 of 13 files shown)


src/reformatters/common/download.py
```diff
@@ -13,7 +13,7 @@
 
 
 def download_to_disk(
-    store: obstore.store.HTTPStore,
+    store: obstore.store.HTTPStore | obstore.store.S3Store,
```
mosegontar: Is there a more generic type we can use here?

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -62,11 +60,14 @@ def out_loc(self) -> Mapping[Dim, CoordinateValueOrRange]:
 class NoaaNdviCdrAnalysisRegionJob(
     RegionJob[NoaaNdviCdrDataVar, NoaaNdviCdrAnalysisSourceFileCoord]
 ):
-    download_parallelism: int = 4
+    download_parallelism: int = 12
```
mosegontar, line 63: might need to go down to 8 (which is what succeeded for me locally) or configure timeout stuff. Or maybe it'll be fine on k8s.

mosegontar:
might be able to drop the boto3 dependency if we can anon list and download with obstore alone

deploy/aws/nodepool.yaml
```diff
@@ -19,7 +19,7 @@ spec:
           values: ["c", "m", "r"]
         - key: "eks.amazonaws.com/instance-cpu"
           operator: Lt
-          values: ["33"]
```
mosegontar, line 22: I'll kubectl apply after this PR merges

src/reformatters/common/fsspec.py
```diff
@@ -31,7 +31,7 @@ def fsspec_apply(
         try:
             if fs.async_impl:
                 return zarr.core.sync.sync(
-                    getattr(fs, f"_{method}")(*args, **kwargs), timeout=120
+                    getattr(fs, f"_{method}")(*args, **kwargs), timeout=500
```
mosegontar, line 34: We got a time out here which was a bit strange. I bumped it up to 500 which might have been excessive but I wanted to avoid multiple attempts to figure out the minimum. We can drop this down but it's possible we'll run into the timeout issue again (though we could always bump it up again)

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
```diff
@@ -16,32 +18,32 @@ class NoaaNdviCdrAnalysisDataset(
...
-    #        cpu="1.3",
-    #        memory="7G",
-    #        secret_names=self.storage_config.k8s_secret_names,
-    #    )
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        """Return the kubernetes cron job definitions to operationally update and validate this dataset."""
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{self.dataset_id}-operational-update",
+            schedule="0 20 * * *",
+            pod_active_deadline=timedelta(minutes=60),
```
mosegontar, line 26: From the backfill we observed shards taking 30 minutes. Giving it a 30 min buffer here (maybe excessive) but not sure if there's a downside.

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
```diff
@@ -16,32 +18,32 @@ class NoaaNdviCdrAnalysisDataset(
...
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="2",
+            memory="100G",
+            shared_memory="76Gi",
+            ephemeral_storage="150G",
```
mosegontar, line 32: This is another one which may be more than we need -- I had tried 50, ran out disk and just reverted back to 150 which had previously been successful, but we could probably calculate this more precisely if this is problematically high (i think it's about 44 GB for the downloaded files alone for a 2 year period).

src/reformatters/common/download.py
mosegontar: These classes don't seem to have a parent, so I didn't find a built in obstore type that covers all the various stores, but I might have overlooked.

src/reformatters/common/download.py
```diff
@@ -77,6 +77,33 @@ def http_store(base_url: str) -> obstore.store.HTTPStore:
     )
 
 
+@functools.cache
+def s3_store(
+    bucket_url: str, region: str, skip_signature: bool = True
```
mosegontar, line 88: This defaults to assuming you are working with public buckets and it would be opt in to use provisioned credentials, but we could invert.

src/reformatters/common/download.py
aldenks: there is, obstore.store.ObjectStore https://github.com/developmentseed/obstore/blob/main/obstore/python/obstore/store.py#L666

src/reformatters/common/fsspec.py
aldenks: yeah i think this is fine. i mostly want to avoid an indefinite hang situation

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
aldenks: this seems good. if it were to hang for any reason this will cause it to get restarted without needing human intervention

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/dynamical_dataset.py
aldenks: disk is cheap, lets not use any more of your time figuring out since back of the envelope this is close

src/reformatters/contrib/noaa/ndvi_cdr/ndvi_cdr/analysis/region_job.py
```diff
@@ -120,8 +117,16 @@ def generate_source_file_coords(
...
-        url = coord.get_url()
-        return http_download_to_disk(url, self.dataset_id)
+        store = s3_store(self.s3_bucket_url, self.s3_region, skip_signature=True)
+
+        s3_url = coord.get_url()
+        object_key = urlparse(s3_url).path.removeprefix("/")
+        local_path = get_local_path(self.dataset_id, object_key)
+
+        download_to_disk(store, object_key, local_path, overwrite_existing=True)
+        log.info(f"Downloaded {object_key} to {local_path}")
```
aldenks: maybe make this a .debug, looks noisy

aldenks (approved):
Beaut!

src/reformatters/common/download.py
mosegontar: Oh perfect, thanks!

PR #180, opened 2025-07-15
Tweak documentation

JackKelly:
This PR:

- Asks users to use lower case for provider and model.
- Fixes a spelling mistake.
- Uses code formatting in more places.


docs/dataset_integration_guide.md
```diff
@@ -27,6 +27,8 @@ Before getting started, follow the brief setup steps in README.md > Local develo
 uv run main initialize-new-integration <provider> <model> <variant>
 ```
 This will add a number of files within `src/reformatters/<provider>/<model>/<variant>` and `tests/<provider>/<model>/<variant>`.
+Please use lower case for the provider, model, and variant. Hyphens are acceptable for the provider
```
JackKelly, line 30: _Are_ hyphens acceptable?!

JackKelly: I assume we want people to use lower case because the NOAA datasets on Dynamical are lower case

CONVENTIONS.md
```diff
@@ -31,10 +31,10 @@ Common utilities and conventions seek to reduce the amount of unique code requir
...
-* Add only extremely minimal code commends and no docstrings unless I ask for them
+* Add only extremely minimal code comments and no docstrings unless I ask for them
   * Add comments only when doing things out of the ordinary, to highlight gotchas, or if less clear code is required due to an optimization
 * Use Python 3.12+ features
 * Follow mypy strict mode. If you need to add an ignore, ignore a specific check like `# type: ignore[specific]`. Always annotate types on all function arguments and return types.
 * Try to follow ruff format
 * Don't write error handing code unless I ask for it, nor smooth over exceptions/errors unless they are expected as part of control flow. In general, write code that will raise an exception early if something isn't expected. Enforce important expectations with asserts.
-* Test each module with pytest
\ No newline at end of file
+* Test each module with pytest
```
JackKelly, line 40: neovim automatically adds a newline at the end of markdown. This wasn't me! But maybe it's nice to have a newline at the end of the file, so the output from `cat` looks good?

docs/dataset_integration_guide.md
aldenks: We normalize capitalization in the init script so class names are PascalCase and file names and paths are snake_case but let me double check the exact requirements and then I'll merge this in! Thanks for this great polish

JackKelly: To give a bit of background: When I first stepped through the dataset integration guide, I ran `initialize-new-integration DWD ICON-EU forecast` which resulted in the creation of the directory `src/reformatters/DWD/ICON_EU/forecast`. I then noticed that the NOAA directory is all lower case. So, to be consistent with the NOAA directory, I removed `src/reformatters/DWD` and re-ran `initialize-new-integration` with `dwd icon-eu forecast`. (If that makes sense?!)

aldenks:
Thank you @JackKelly! I'm going to actually remove the lowercase requirement (we can do that automatically) and ill re-update the docs when i do.

PR #182, opened 2025-07-16
Append newline char when writing __init__.py

JackKelly:
This PR just appends `\n` to the text written to `src/reformatters/{provider}/{model}/{variant}/__init__.py` so `ruff` doesn't complain immediately after running `uv run main initialize-new-integration <provider> <model> <variant>`.

Fixes the "ruff: Failed" issue shown below:

image

### DIFF
--- a/src/reformatters/example/new_dataset.py
+++ b/src/reformatters/example/new_dataset.py
@@ -74,7 +74,7 @@ def initialize_new_integration(
 
     dataset_class_name = example_to_actual_mappings["ExampleDataset"]
     (src_path / "__init__.py").write_text(
-        f"from .dynamical_dataset import {dataset_class_name} as {dataset_class_name}"
+        f"from .dynamical_dataset import {dataset_class_name} as {dataset_class_name}\n"
     )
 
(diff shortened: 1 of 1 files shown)


aldenks (approved):
Thank you!

JackKelly:
Thanks for approving this PR!

Hmm.. the `CodeQL` check is taking a long time to run...

aldenks:
@JackKelly do you mind trying this? https://stackoverflow.com/a/58190576 if that doesn't work, I'll turn that check off entirely.

JackKelly:
ooh, interesting... I've given that a go...

JackKelly:
Hmm... the `CodeQL` check still says "waiting for stutus to be reported" :cry: 

Sorry if I'm doing something wrong?!

aldenks:
@JackKelly you were doing everything right, thanks for being patient! Something about how that check was configured in github wasn't jiving with the setting where someone with write access to the repo needs to approve pipeline runs. That check was not actually useful anyhow (the other two checks are what run the test suite on amd64 and arm64) and they work and were successful as expected.

Issue #183, opened 2025-07-21
How to handle the slightly different shapes of OCF's ICON-EU archive on Hugging Face?


JackKelly:
The full set of [ICON-EU](https://www.dwd.de/EN/ourservices/nwp_forecast_data/nwp_forecast_data.html)'s coordinates in the GRIB files on [DWD's HTTP server](https://opendata.dwd.de/weather/nwp/icon-eu/grib/) are:

|           | length |  start |    end    |
|----------:|-------:|:------:|:---------:|
|      step |     93 |    0   | 120 hours (hourly to 78 hours, then 3-hourly from 81 to 120 hours) |
|  latitude |    657 |  29.5° |   70.5°   |
| longitude |   1377 | -23.5° |   62.5°   |

Open Climate Fix's [ICON-EU archive](https://huggingface.co/datasets/openclimatefix/dwd-icon-eu) only contains a subset of these coordinates 😢.

Open Climate Fix's ICON-EU archive on Hugging Face is really two archives glued together:

- 2020 until March 2023: Derived from an archive created by Dexter Energy. This has the correct `step` and `latitude` coordinates. But only has 1097 `longitude` coordinates (instead of 1377). It is missing the eastern-most 280 `longitude` coords: The archive ends at 45° longitude (instead of 62.5°). The archive chops off Turkmenistan.
- March 2023 until the present day: Created by OCF's script which pulls GRIBs directly from DWD. This has the correct `latitude` and `longitude` coordinates, but only has 78 `step` values (instead of 93 steps). The forecast horizon ends at 3 days and 6 hours, instead of 5 days. The archive contains all the hourly steps. But none of the 3-hourly steps (from 81 to 120 hours).

([Here's a Google Sheet](https://docs.google.com/spreadsheets/d/1FEmQL95Xm8OTLkcxkhxb7ZRjWmzeV1E6I618Hp_hg7c/edit?usp=sharing) with the same information in tabular form, with the discrepancies highlighted in red.)

How should we handle this in the ICON-EU Zarr on dynamical? A few options for your consideration:

1. Start the archive in 2020. Use the full set of ICON-EU coordinates
    - Pros:
        - Doesn't throw away information. 
        - If we come across an archive which _does_ include the missing data then we can back-fill without changing the shape of the dataset.
    - Cons: NaNs.
2. Start the archive in 2020. Only use the _intersection_ of the available coordinates (i.e. only 79 time steps, and only 1097 longitude values).
[...]

Issue #176, opened 2025-07-11
Which NWP variables and vertical levels to select for ICON-EU?


JackKelly:
We plan to put a reduced set of the variables into the ICON-EU Zarr on Dynamical.

And we currently plan to _not_ have a separate dimension for vertical levels. Instead, we'll use a suffix in the variable name to denote different vertical levels (e.g. `u_10m` for the 10 metre U wind component). And we don't plan to store many different vertical levels. (Update: But see [the discussion below](https://github.com/dynamical-org/reformatters/issues/176#issuecomment-3066947166) about the importance of vertical levels for the AI NWP community!)

Note that OCF's ICON-EU archive from 2020 to 2023 includes only 27 of the 63 ICON-EU variables. (See [here](https://huggingface.co/datasets/openclimatefix/dwd-icon-eu#nwp-variables-archived-for-2020-to-march-2023) for a list of the 27 variables included in the 2020 to 2023 archive; and see [here](https://huggingface.co/datasets/openclimatefix/dwd-icon-eu#nwp-variables-archived-for-march-2023-onwards) for a list of all 63 variables). We'd like the ICON-EU dataset on Dynamical to be as easy to use as possible, and to have as little missing data as possible. So, ideally, the ICON-EU variables exposed on Dynamical would be selected from the 27 variables included in OCF's ICON-EU archive from 2020 to 2023.

jacobbieker:
I do know that the ICON-EU/ICON-Global archives are/have been used for ML weather forecasting training datasets. For those, including the pressure level variables would be necessary to have. Pretty much all the pressure level variables are used by the main ML weather models, so including geopotential, U/V wind, vertical velocity, relative/specific humidity and temperature would be really helpful. 

Other variables used, but not as uniformly, would include the total column water vapor, and surface pressure.

JackKelly:
Very good point, thank you Jacob! At the moment, I think we're working on the assumption that we might not have the capacity to expose all the pressure levels in the Zarr.

But I'd be keen to support the AI NWP community.

Jacob, would it be sufficient if we only put single-level data into the Zarr, but also archived DWD's original GRIB2 files on cloud object storage (so the AI NWP community could use those GRIBs?)

What do you think, @aldenks?

jacobbieker:
Yeah, if its not feasible to have them included, archiving the gribs would work, just having a continuing archive somewhere would be super useful.

JackKelly:
[...]

Issue #173, opened 2025-07-11
DWD ICON-EU forecast dataset (time series optimized)


JackKelly:
Goal: Add DWD's [ICON-EU](https://www.dwd.de/EN/ourservices/nwp_forecast_data/nwp_forecast_data.html) NWP to Dynamical as a single sharded Zarr v3. Update the Zarr directly from the GRIB files published on [DWD's HTTP server](https://opendata.dwd.de/weather/nwp/). DWD deletes the GRIB files after 24 hours. Backfill from Open Climate Fix's [ICON-EU archive on Hugging Face](https://huggingface.co/datasets/openclimatefix/dwd-icon-eu).

My tasks:

- [x] Email everyone who downloaded ICON-EU from OCF's archive on Hugging Face
  - [x] Draft email. Ask them which NWP variables they'd like. Invite their feedback on #176. Say that the ICON-EU archive on HF will be turned off 6 months after the backfilled ICON-EU dataset appears on Dynamical. Explain that the Dynamical dataset will make their lives much easier 🙂: ARCO FTW. 
  - [x] Share draft with @aldenks & @devsjc & Emily (at OCF).
  - [x] Send email after Alden & Sol & Emily have approved the draft.
  - [x] Update the README on the ICON-EU Hugging Face archive with text from the email.
  - [x] [Post to LinkedIn](https://www.linkedin.com/posts/jackkelly0_dynamicalorg-activity-7353352542971432961-UYC1?utm_source=share&utm_medium=member_desktop&rcm=ACoAAAEnercBa7aVz-fCikGRwNaOXk-AoHlrPNc).
- [x] Email DWD
  - [x] Draft email. (Note to self: Check notes on meeting minutes for what to include in this email).
  - [x] Share draft with @aldenks & @devsjc.
  - [x] Send email.
- [x] #183 
- [x] Decide on a chunking & sharding scheme for the ICON-EU Zarr.
- [x] #184
- [x] #226
- [x] #360 (which would replace PRs #331 & #318)
- [x] #258 
- [x] #390
- [x] #381
- [x] #392
- [x] #420
- [x] Test transfer of gribs from DWD's HTTP to Source Co-Op in production
- [x] PR to instantiate and return `ReformatCronJob` and `ValidationCronJob` from `DwdIconEuForecastDataset.operational_kubernetes_resources` and test
- [x] Update the start date in `template_config.py` just before launching the live updates
- [x] Review var requests and add to dataset #176
- [x] Backfill
[...]

PR #184, opened 2025-07-21
ICON-EU step 1: Implement `template_config.py`

JackKelly:
This PR implements live-updating DWD's ICON-EU NWP directly from DWD.

A subsequent PR will implement back-filling from OCF's ICON-EU archive on Hugging Face.

## Related issues:

- #173 
- #176 
- #183

## Related external links

- DWD's [Database Reference for the Global and Regional ICON and ICON-EPS Forecasting System](https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf) (PDF). Detailed description of the variables included in ICON-EU.
- [DWD's HTTP server](https://opendata.dwd.de/weather/nwp/icon-eu/grib/).
- [OCF's ICON-EU archive on Hugging Face](https://huggingface.co/datasets/openclimatefix/dwd-icon-eu).
- [A Google Sheet](https://docs.google.com/spreadsheets/d/1FEmQL95Xm8OTLkcxkhxb7ZRjWmzeV1E6I618Hp_hg7c/edit?usp=sharing) showing the slightly different shapes of the two parts of OCF's archive, with the discrepancies highlighted in red.

## Task list

I'll mark this PR as "ready for review" when all these tasks are done

- [x] Register ICON-EU in `src/reformatters/__main__.py`
- [x] `template_config.py`
- [x] Check the NWP params in `template_config.py` against [the ICON docs](https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf). Copy-and-paste the description of each variable from the PDF into `template_config.py`. Link to the PDF from the code.
- [x] Check if surface albedo is exactly the same at a given time of day, every forecast
- [x] `template_config_test.py`

## For a subsequent PR:
- `region_job.py`
- `region_job_test.py`
- `dynamical_dataset.py`
- `dynamical_dataset_test.py`


src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,525 @@
...
+        )
+
+        default_keep_mantissa_bits = 7
+
+        return [
+            DwdIconEuDataVar(
+                name="downward_diffuse_short_wave_radiation_flux_surface",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="msdfswrf",  # From ECMWF parameter database.
```
JackKelly: ~~@aldenks would you like `short_name` to be set to exactly the short name used by ICON-EU? Or should I try to find the most similar "short name" from ECMWF or NOAA?~~

~~In the current version of the file, I've used `grib_element` to store the code used in ICON-EU. And I've set `short_name` to the most similar short name from the [ECMWF parameter database](https://codes.ecmwf.int/grib/param-db/). My motivation was to try to make it as easy as possible for folks to use NWPs interchangeably. But maybe I've gone down the wrong path, and I should set `short_name` to the keys in the ICON-EU GRIBs (and described in the [ICON documentation PDF](https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf)).~~

(I must admit I'm a bit confused by GRIB naming conventions. It feels like the WMO _tried_ to standardise on a common vocab of short_names. But then the providers did their own thing anyway! Which I assume is because there are subtle variations in the way different providers define each concept.)

aldenks:
@JackKelly I have a slight preference to review in more incremental PRs (e.g. one for the template config + tests, one for region job, one for dynamical dataset). So given that, this looks like it's ready for a review? I'll give it a first pass this afternoon

JackKelly:
Hey. I like small PRs too! Please hold fire on reviewing. There are some issues with the NWP variable names that I'd like to fix first, if that's ok.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Actually, please ignore my comment above! I've just found that the ICON-EU GRIB files direct from DWD contain all the info we need (including the "official" `short_name`). So I'll update the `template_config` with the info from the GRIB files. Hopefully tomorrow.

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+    def derive_coordinates(
+        self, ds: xr.Dataset
+    ) -> dict[str, xr.DataArray | tuple[tuple[str, ...], np.ndarray[Any, Any]]]:
+        """Return a dictionary of non-dimension coordinates for the dataset.
+
+        Called whenever len(ds.append_dim) changes.
+        """
+        return {
+            "valid_time": ds["init_time"] + ds["lead_time"],
+            "ingested_forecast_length": (
```
JackKelly, line 98: should I include `expected_forecast_length`?

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+
+class DwdIconEuDataVar(DataVar[DwdIconEuInternalAttrs]):
+    pass
+
+
+class DwdIconEuForecastTemplateConfig(TemplateConfig[DwdIconEuDataVar]):
+    dims: tuple[Dim, ...] = ("init_time", "lead_time", "latitude", "longitude")
+    append_dim: AppendDim = "init_time"
+    append_dim_start: Timestamp = pd.Timestamp(
+        "2020-01-01T00:00"  # The start of OCF's ICON-EU archive on Hugging Face.
```
JackKelly: should this be changed to the date that we start pulling GRIB files from DWD?

JackKelly:
OK, I think this is ready for review! I've removed the files that I haven't yet updated. Although I've just noticed that the code quality check is failing...

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+    StatisticsApproximate,  # noqa: F401
+)
+from reformatters.common.template_config import (
+    SPATIAL_REF_COORDS,  # noqa: F401
+    TemplateConfig,
+)
+from reformatters.common.types import AppendDim, Dim, Timedelta, Timestamp
+from reformatters.common.zarr import (
+    BLOSC_4BYTE_ZSTD_LEVEL3_SHUFFLE,  # noqa: F401
+    BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE,  # noqa: F401
```
aldenks: you can remove all these noqa's about unused imports.  thinking about the cleanest way to handle this in the example file..

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: yes, lets move up to the start date we will initially publish. realizing there's a rolling archive, you might need to update this again as we're close to deploying operationally and that's fine. We will extend it back as we expand the archive (that will involve doing a backfill to a new version of the zarr)

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+    append_dim_start: Timestamp = pd.Timestamp(
+        "2020-01-01T00:00"  # The start of OCF's ICON-EU archive on Hugging Face.
+    )
+    append_dim_frequency: Timedelta = pd.Timedelta("6h")
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def dataset_attributes(self) -> DatasetAttributes:
+        return DatasetAttributes(
+            dataset_id="dwd-icon_eu-forecast",
```
aldenks:
```suggestion
            dataset_id="dwd-icon-eu-forecast",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+            self.append_dim: self.append_dim_coordinates(
+                self.append_dim_start + self.append_dim_frequency
+            ),
+            "lead_time": (  # Called "step" in the ICON-EU GRIB files.
+                pd.timedelta_range("0h", "78h", freq="1h").union(
+                    pd.timedelta_range("81h", "120h", freq="3h")
+                )
+            ),
+            "latitude": np.linspace(29.5, 70.5, 657),
+            "longitude": np.linspace(-23.5, 62.5, 1377),
```
aldenks, line 86: confirming these are the pixel centers (not corners)? if they match what you get from `xr.load_dataset(ICON_EU_GRIB_FILENAME_FROM_DWD, engine='cfgrib')` they are centers. Another way to confirm is the dataset bounds in gdalinfo should be `0.0625 / 2` wider than the min/max values here

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: yes please even though i'll be all the same value here (unless we start doing the hourly inits). that way user code can do things like this `ds.sel(init_time=ds["ingested_forecast_length"] == ds["expected_forecast_length"])` and the intent is really clear.

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    semi_minor_axis=6371229.0,
+                    inverse_flattening=0.0,
+                    reference_ellipsoid_name="Sphere",
+                    longitude_of_prime_meridian=0.0,
+                    prime_meridian_name="Greenwich",
+                    geographic_crs_name="Coordinate System imported from GRIB file",
+                    horizontal_datum_name="unnamed",
+                    grid_mapping_name="latitude_longitude",
+                    spatial_ref='GEOGCS["Coordinate System imported from GRIB file",DATUM["unnamed",SPHEROID["Sphere",6371229,0]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST]]',
+                    comment="From the WKT string output by gdalinfo.",
```
aldenks:
```suggestion
                    comment="A perfect sphere geographic CRS with a radius of 6,371,229m, extracted from grib.",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+            ),
+        ]
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def data_vars(self) -> Sequence[DwdIconEuDataVar]:
+        """Define metadata and encoding for each data variable."""
+        var_chunks: dict[Dim, int] = {
+            "init_time": 1,
+            "lead_time": 120,
```
aldenks: I think it's actually 93 steps?

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+            "latitude": 73,
```
aldenks:
```suggestion
            "latitude": 165,
```

The exactly evenly divisible property of 73 is nice, but assuming compression to about 20% of original size of the 4 byte float32s that will give us chunks of 830KB (`(93*73*153) * 4 * 0.2`) which is a bit small. It's also nice to have roughly square chunks geographically. 165 would give us ~1.9MB chunks and the last chunk would have data in all but the last 3 pixels.

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    comment="From the WKT string output by gdalinfo.",
+                ),
+            ),
+        ]
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def data_vars(self) -> Sequence[DwdIconEuDataVar]:
+        """Define metadata and encoding for each data variable."""
+        var_chunks: dict[Dim, int] = {
```
aldenks, line 286:
```suggestion
        # Roughly 3.4MB uncompressed, 1.9MB compressed
        var_chunks: dict[Dim, int] = {
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+            "init_time": 1,
+            "lead_time": 120,
+            "latitude": 73,
+            "longitude": 153,
+        }
+        var_shards: dict[Dim, int] = {
```
aldenks, line 293:
```suggestion
        # Roughly 337MB uncompressed, 67MB compressed
        var_shards: dict[Dim, int] = {
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+            # diffuse radiation) in the Zarr because, to quote the DWD
+            # Database Reference: "Values over snow-free land points are based
+            # on a monthly mean MODIS climatology." It's much more data-efficient
+            # to just download those monthly means from DWD.
+            DwdIconEuDataVar(
+                name="downward_diffuse_short_wave_radiation_flux_surface",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="aswdifd_s",
+                    long_name="Downward diffusive short wave radiation flux at surface (mean over forecast time)",
```
aldenks: what does "mean over forecast time" mean here?

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    ),
+                    standard_name="air_temperature",
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="t_2m",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="total_precipitation",
```
aldenks:
```suggestion
                name="precipitation_surface",
```
to match our other datasets

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    ),
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="tot_prec",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                    deaccumulate_to_rate=True,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="wind_u_10",
```
aldenks:
```suggestion
                name="wind_u_10m",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    standard_name="eastward_wind",
+                    comment="Zonal wind at 10m above ground",
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="u_10m",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="wind_v_10",
```
aldenks:
```suggestion
                name="wind_v_10m",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    step_type="instant",
+                    comment="Surface pressure reduced to MSL",
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="pmsl",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="relative_humidity",
```
aldenks:
```suggestion
                name="relative_humidity_2m",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                        " H_SNOW = W_SNOW / RHO_SNOW and is limited to H_SNOW <= 40 m."
+                    ),
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="h_snow",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="pressure_reduced_to_msl",
```
aldenks:
```suggestion
                name="pressure_reduced_to_mean_sea_level",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                    step_type="instant",
+                    comment="Cloud Cover (400 - 800 hPa). Different agencies use different short_names for this same parameter: ECMWF: MCC; WMO GRIB table: MCDC.",
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="clcm",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="total_cloud_cover",
```
aldenks, line 415:
```suggestion
                name="total_cloud_cover_atmosphere",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,593 @@
...
+                        " shortwave radiation flux."
+                    ),
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="aswdir_s",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="convective_available_potential_energy",
```
aldenks, line 355: is this entire atmosphere?
```suggestion
                name="convective_available_potential_energy_atmosphere",
```

aldenks (changes requested):
Awesome! Just a few tiny tweaks but this is basically good to go.

Don't worry too much about the units and short/long names right now. We need to do a detail sweep and get fully aligned to cf conventions and we can do that before we release this.

re the mypy failure in the pipeline
- i think you need to create empty `dwd/__init__.py` and `dwd/icon_eu/__init__.py` files. I should update the init script to create those
- if you install pre-commit ([instructions](https://github.com/dynamical-org/reformatters/blob/main/README.md#setup) it will check for you as you commit. you can also run `uv run mypy` locally.

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: I also just added a ruff check rule for this https://github.com/dynamical-org/reformatters/pull/185

`uv run ruff check --fix` (once you get the new rules on main) will remove all the unused ones for you

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: SGTM! I've added `expected_forecast_length`.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: I agree, using hyphens looks much nicer :slightly_smiling_face:! Committed.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Good question! These coordinates pass the two tests you mentioned. So I'm confident that these coords are for the pixel _centers_.

For completeness, here are the corner coords from `gdalinfo`:

```
Origin = (-23.531250000000000,70.531250000000000)
Pixel Size = (0.062500000000000,-0.062500000000000)
Corner Coordinates:
Upper Left  ( -23.5312500,  70.5312500) ( 23d31'52.50"W, 70d31'52.50"N)
Lower Left  ( -23.5312500,  29.4687500) ( 23d31'52.50"W, 29d28' 7.50"N)
Upper Right (  62.5312500,  70.5312500) ( 62d31'52.50"E, 70d31'52.50"N)
Lower Right (  62.5312500,  29.4687500) ( 62d31'52.50"E, 29d28' 7.50"N)
Center      (  19.5000000,  50.0000000) ( 19d30' 0.00"E, 50d 0' 0.00"N)
```

JackKelly: Should we add your two tests to `dataset_integration_guide.md`? (I'm happy to create a new PR to add your tests to the `dataset_integration_guide` if you'd like?)

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Oops! Great catch! Thank you!

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: The text "mean over forecast time" is from the ICON-EU GRIB files.

I'm 99% sure that "mean over forecast time" means that the value in the GRIB files is the mean over each 1-hour timestep. In other words, I think the text "mean over forecast time" repeats what `step_type="avg"` already tells us! So I've removed "mean over forecast time" from the `long_name`.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Surprisingly, the ICON-EU GRIBs and the [ICON-EU manual](https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf) claim that `GRIB_typeOfLevel` is set to `surface` for the `total_cloud_cover` variable.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Surprisingly, the ICON-EU GRIBs and the [ICON-EU manual](https://www.dwd.de/DWD/forschung/nwv/fepub/icon_database_main.pdf) claim that `GRIB_typeOfLevel` is set to `surface` for the `cape` variable. Although I agree that `entire atmosphere` would make more sense!

JackKelly:
> re the mypy failure in the pipeline
> 
> * i think you need to create empty `dwd/__init__.py` and `dwd/icon_eu/__init__.py` files. I should update the init script to create those

So... this `mypy` error has been an interesting rabbit hole to go down!

I committed the missing Python files that `initialize-new-integration` had created, but I had deliberately not committed yet because I was gonna add those in a subsequent PR. This is why the mypy error wasn't appearing on my local machine. 

And, as you suggested, I created `__init__.py` files in all the relevant directories under `test/`. 

These fixed the mypy error. 

But then I got an import error in pytest, which I think is caused by [a bug in `pytest 8.3.4`](https://github.com/pytest-dev/pytest/pull/13053). After upgrading `pytest` and creating the `__init__.py` files, the bug has gone away :slightly_smiling_face:

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: This sounds good to me but unfortunately both tests fail in `tests/dwd/icon_eu/forecast/template_config_test.py` because the Pydantic model defined in `reformatters.common.config_models.Encoding` [checks](https://github.com/JackKelly/reformatters/blob/416748a8be61acf4b2bb4bf85779694645b8520b/src/reformatters/common/config_models.py#L102-L108) that each shard size is an exact multiple of each chunk size! So I'll change the chunk latitude size to 219 (657 / 3) if that's OK?

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: I'm sorry for being slow but please could you walk me through where the "3.4MB uncompressed" comes from? (I think I understand where the "1.9MB compressed" comes from!)

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: oh good idea. maybe actually the outlines of a test in `src/reformatters/example/template_config_test.py`, that way we keep the dataset integration guide not too complex. but that can be a follow up

JackKelly: SGTM. I agree, let's implement these tests in a follow-up PR. I've started issue #193 to keep track of the implementation of these tests.

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: So, if it's OK, I'd propose that I don't change `total_cloud_cover` to `total_cloud_cover_atmosphere` (because the ICON-EU manual claims that `total_cloud_cover` is a _surface_ variable). Unless I've misunderstood?!

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: So, if it's OK, I'd propose that I don't change `convective_available_potential_energy` to `convective_available_potential_energy_atmosphere` (because the ICON-EU manual claims that `convective_available_potential_energy` is a _surface_ variable). Unless I've misunderstood?!

JackKelly:
OK, I _think_ this PR is ready to merge :slightly_smiling_face:. `mypy` and `pytest` pass on my local machine. I think CI should pass once PR #192 is merged. Please shout if you have any other tweaks you'd like me to make! The only two suggestions I haven't committed are to append `_atmosphere` to the end of `total_cloud_cover` and `convective_available_potential_energy`. I haven't committed these suggestions because the ICON-EU manual says those two variable are surface-level. But please shout if you'd like me to append `_atmosphere` anyway!

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dev-dependencies = [
     "pandas-stubs>=2.2.2.240909",
     "pre-commit>=3.8.0",
     "pyqt6>=6.7.1",
-    "pytest>=8.3.4",
+    "pytest>=8.4.1",
```
JackKelly, line 41: Please see this comment for more info on why I updated `pytest`: https://github.com/dynamical-org/reformatters/pull/184#issuecomment-3128306904

tests/dwd/icon_eu/forecast/region_job_test.py
JackKelly, line 1: I haven't modified this file yet. This is just the output of `initialize-new-integration`. I'll modify this file in a subsequent PR.

uv.lock
JackKelly, line 1: See this comment for more info on why I updated `pytest`: https://github.com/dynamical-org/reformatters/pull/184#issuecomment-3128306904

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
JackKelly, line 1: I haven't modified this file yet. This is just the output of `initialize-new-integration`. I'll modify this file in a subsequent PR.

src/reformatters/dwd/icon_eu/forecast/region_job.py
JackKelly, line 1: I haven't modified this file yet. This is just the output of `initialize-new-integration`. I'll modify this file in a subsequent PR.

tests/dwd/icon_eu/forecast/region_job_test.py
JackKelly, line 1: I haven't modified this file yet. This is just the output of `initialize-new-integration`. I'll modify this file in a subsequent PR.

tests/dwd/icon_eu/forecast/dynamical_dataset_test.py
JackKelly, line 1: I haven't modified this file yet. This is just the output of `initialize-new-integration`. I'll modify this file in a subsequent PR.

uv.lock
```diff
@@ -1174,17 +1174,18 @@ wheels = [
 
 [[package]]
 name = "pytest"
-version = "8.3.4"
+version = "8.4.1"
```
JackKelly, line 1177: See this comment for more info on why I updated pytest: https://github.com/dynamical-org/reformatters/pull/184#issuecomment-3128306904

aldenks:
awesome! will give this a final review and merge this evening

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,622 @@
...
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def dataset_attributes(self) -> DatasetAttributes:
+        return DatasetAttributes(
+            dataset_id="dwd-icon-eu-forecast",
+            dataset_version="0.1.0",
+            name="DWD ICON-EU Forecast",
+            description="High-resolution weather forecasts for Europe from the ICON-EU model operated by Deutscher Wetterdienst (DWD).",
+            attribution="DWD ICON-EU data processed by dynamical.org from DWD.",
```
aldenks, line 63:
```suggestion
            attribution="DWD ICON-EU data processed by dynamical.org.",
```

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: doh, i think i had my math wrong, it's `# of elements * 4 bytes` for a float32: `1x93×219×153×4bytes`  ~= **12.5MB uncompressed** and then I've found compressing down to 20% of uncompressed size is a reasonable expectation (we can often go a bit lower, sometimes much lower for very compressible variables) so `1×93×219×153×4bytes×0.2` ~= **2.5MB compressed**.

src/reformatters/dwd/icon_eu/forecast/template_config.py
aldenks: lets leave as is for now and come back with a later metadata polish and align to CF Conventions. Same for the other possible naming + unit questions in our metadata.

src/reformatters/dwd/icon_eu/forecast/template_config.py
```diff
@@ -0,0 +1,622 @@
...
+                    comment="Relative humidity at 2m above ground. Other short_names used for this parameter: rh, 2r, r.",
+                    standard_name="relative_humidity",
+                ),
+                internal_attrs=DwdIconEuInternalAttrs(
+                    grib_element="relhum_2m",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            DwdIconEuDataVar(
+                name="soil_water_runoff",
```
aldenks, line 480: I'm not yet sold on including the soil and surface runoff and snow cover/water equivalent variables (they are usually not great from atmospheric models) but we can decide the exact vars later, no change needed now

aldenks (approved):
Beauty! Merging. I have a larger TODO around aligning with CF Conventions and we can do that after we have things running.

im going to make one follow up commit correcting the chunk size comment (edit: done here https://github.com/dynamical-org/reformatters/pull/195)

src/reformatters/dwd/icon_eu/forecast/template_config.py
JackKelly: Cool... TBH, I only included the water runoff variables because I thought that Upstream might use those variables! (I don't know anything about modelling rivers!) So, if you don't need them, then let's not include the water runoff variables (but, like you say, we can tweak the variables list later).

PR #189, opened 2025-07-29
NOAA NDVI CDR: fetch more recent files from NCEI instead of S3

mosegontar:
We noticed more recent files for this dataset are available in NCEI (and THREDDS - see https://www.ncei.noaa.gov/products/climate-data-records/normalized-difference-vegetation-index for links to all available sources) than are S3. We believe that S3 and GCS are on a lag, so we want to check and fetch from NCEI for very recent data, and otherwise fetch from S3.

### DIFF
--- a/src/reformatters/contrib/noaa/ndvi_cdr/analysis/region_job.py
+++ b/src/reformatters/contrib/noaa/ndvi_cdr/analysis/region_job.py
@@ -1,3 +1,4 @@
+import re
 from collections.abc import Callable, Mapping, Sequence
 from pathlib import Path
 from typing import cast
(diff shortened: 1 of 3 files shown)


src/reformatters/contrib/noaa/ndvi_cdr/analysis/region_job.py
```diff
@@ -61,14 +68,20 @@ def out_loc(self) -> Mapping[Dim, CoordinateValueOrRange]:
 class NoaaNdviCdrAnalysisRegionJob(
     RegionJob[NoaaNdviCdrDataVar, NoaaNdviCdrAnalysisSourceFileCoord]
 ):
-    download_parallelism: int = 10
+    # Set lower than would be needed for fetching exclusively from S3
+    # to accomodate the cases where we are downloading from NCEI.
+    download_parallelism: int = 5
```
mosegontar, line 73: i tried to do this dynamically and gave up quickly

aldenks (approved):
Approved thank you!

PR #191, opened 2025-08-04
StoreFactory refactor 

mosegontar:
This PR introduces a `StoreFactory` abstraction that establishes a new pattern for accessing the `final_store` (the target final destination store for the dataset).  The motivation is to provide downstream modules (e.g., `RegionJob`) the means to re-instantiate the store, in order to support stores such as `icechunk` stores, which require instantiating a new store for every commit.

Notes the term `final store` has been replaced with `primary store` in preparation for support for replica stores.

### DIFF
--- a/src/reformatters/common/region_job.py
+++ b/src/reformatters/common/region_job.py
@@ -15,7 +15,6 @@ import numpy as np
 import pandas as pd
 import pydantic
 import xarray as xr
-import zarr
 from pydantic import AfterValidator, Field, computed_field
 
 from reformatters.common import template_utils
@@ -28,6 +27,7 @@ from reformatters.common.reformat_utils import (
     create_data_array_and_template,
 )
 from reformatters.common.shared_memory_utils import make_shared_buffer, write_shards
+from reformatters.common.storage import StoreFactory
 from reformatters.common.types import (
     AppendDim,
     ArrayND,
@@ -36,7 +36,7 @@ from reformatters.common.types import (
     Timestamp,
 )
 from reformatters.common.update_progress_tracker import UpdateProgressTracker
-from reformatters.common.zarr import copy_data_var, get_mode
+from reformatters.common.zarr import copy_data_var
 
 log = get_logger(__name__)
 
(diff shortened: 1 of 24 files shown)


src/reformatters/common/template_utils.py
```diff
@@ -36,6 +51,18 @@ def write_metadata(
         sort_consolidated_metadata(Path(store) / "zarr.json")
 
 
+def _get_mode_from_path_store(store: Path) -> Literal["w", "w-"]:
```
aldenks, line 54: do we need to handle these two cases?

https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/zarr.py#L82-L85

aldenks: I think for gefs stuff that hasn't been ported this will be a FsspecStore and we need to get the `.path`

mosegontar: Those are still calling the old `get_mode` I think and passing in the mode that way. After they are refactored, they'll use a store factory. 

We could definitely move that logic in here as well, but then it might make sense to refactor the legacy implementations to not pass in `mode` as well. In such a case, it seems we could either get rid of `get_mode` entirely, or get rid of this `_get_mode_from_path_store` function and continue to call `get_mode` from within `write_metadata` (after adding the `latest.zarr` bit).

I might be misunderstanding your thought here though or have misunderstood your comment in slack re: copying the old logic for the `path` case and using `get_mode` for the legacy store case, so LMK!

aldenks: ahh no i like that. lets port over as little cruft as possible.

src/reformatters/common/region_job.py
```diff
@@ -489,9 +491,8 @@ def process(self) -> Mapping[str, Sequence[SOURCE_FILE_COORD]]:
                 data_var_groups, self.max_vars_per_download_group
             )
 
-        template_utils.write_metadata(
-            self.template_ds, self.tmp_store, get_mode(self.tmp_store)
-        )
+        # Hardcode w since many region jobs may write metadata to the same tmp store
```
aldenks: can remove outdated comment

aldenks (approved):
Approved!

PR #194, opened 2025-08-05
Refactor zarr data copying to use store.set instead of fsspec operations

mosegontar:
This PR refactors how we handle copying of data shards and metadata from the tmp store to another storage location. Instead of using `fsspec_apply` we can use the `set` method on our destination store. This also sets us up to be able to interface with an `icechunk` store in the same way (where fsspec operations would not work).

### DIFF
--- a/src/reformatters/common/update_progress_tracker.py
+++ b/src/reformatters/common/update_progress_tracker.py
@@ -3,16 +3,31 @@ import queue
 import threading
 from collections.abc import Sequence
 
+import fsspec  # type: ignore
 import zarr.storage
 
 from reformatters.common.config_models import BaseInternalAttrs, DataVar
 from reformatters.common.fsspec import fsspec_apply
 from reformatters.common.logging import get_logger
-from reformatters.common.zarr import _get_fs_and_path
 
 log = get_logger(__name__)
 
 PROCESSED_VARIABLES_KEY = "processed_variables"
+
+
+def _get_fs_and_path(
+    store: zarr.abc.store.Store,
+) -> tuple[fsspec.AbstractFileSystem, str]:
+    """Gross work around to allow us to make other store types quack like FsspecStore."""
+    fs = getattr(store, "fs", None)
+    if not isinstance(fs, fsspec.AbstractFileSystem):
+        raise ValueError(
+            "primary_store must have an fs that is an instance of fsspec.AbstractFileSystem"
+        )
+    path = getattr(store, "path", None)
+    if not isinstance(path, str):
+        raise ValueError("primary_store must have a path attribute that is a string")
+    return fs, path
 
 
 class UpdateProgressTracker:
--- a/src/reformatters/common/zarr.py
+++ b/src/reformatters/common/zarr.py
@@ -10,6 +10,7 @@ from fsspec.implementations.local import LocalFileSystem  # type: ignore
 
 from reformatters.common.config import Config
 from reformatters.common.logging import get_logger
+from reformatters.common.retry import retry
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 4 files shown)


src/reformatters/common/zarr.py
```diff
@@ -151,25 +145,15 @@ def copy_zarr_metadata(
...
-            "primary_store must have an fs that is an instance of fsspec.AbstractFileSystem"
+        relative_path = str(file.relative_to(tmp_store))
+        sync_to_store(primary_store, relative_path, file.read_bytes())
+
+
+def sync_to_store(store: zarr.abc.store.Store, key: str, data: bytes) -> None:
+    zarr.core.sync.sync(
+        store.set(
+            key,
+            zarr.core.buffer.cpu.Buffer.from_bytes(data),
```
mosegontar: Do we need to support some context where we'd use a GPU buffer instead?

For context on this choice, there are two submodules of `zarr.buffer`:
https://zarr.readthedocs.io/en/main/api/zarr/buffer/index.html

* [zarr.buffer.cpu](https://zarr.readthedocs.io/en/main/api/zarr/buffer/cpu/index.html)
* [zarr.buffer.gpu](https://zarr.readthedocs.io/en/main/api/zarr/buffer/gpu/index.html)

There is also a function called. This was what I saw used in some icechunk tests. 
https://zarr.readthedocs.io/en/main/api/zarr/buffer/index.html#zarr.buffer.default_buffer_prototype. 

I got things working previously with that, so I think we could saw out this cpu.Buffer for that one. The documentation is sparse https://zarr.readthedocs.io/en/main/_modules/zarr/core/buffer/core.html#default_buffer_prototype. Ultimately it seems to do some kind of [dynamic look up](https://zarr.readthedocs.io/en/stable/_modules/zarr/registry.html#get_buffer_class) to select the type of buffer to use. There's a chance this is more generalizable/flexible.

src/reformatters/common/update_progress_tracker.py
```diff
@@ -3,18 +3,33 @@
...
 from reformatters.common.fsspec import fsspec_apply
 from reformatters.common.logging import get_logger
-from reformatters.common.zarr import _get_fs_and_path
 
 log = get_logger(__name__)
 
 PROCESSED_VARIABLES_KEY = "processed_variables"
 
 
+def _get_fs_and_path(
```
mosegontar, line 18: I moved this into here because now it is the only module making use of this function and we intend to deprecate/remove when we move onto refactoring progress tracking.

src/reformatters/common/zarr.py
aldenks: i can't see an instance were we would want to use a gpu, but using default_buffer_prototype seems like a good idea. it will use the default config value which is cpu (https://zarr.readthedocs.io/en/stable/user-guide/config.html)

aldenks: If there's any way we can avoid a copy when creating the Buffer that would be ideal. im not sure we can but check out the methods on it

mosegontar: oh good thought, i'll take a look

src/reformatters/common/zarr.py
```diff
@@ -113,15 +111,11 @@ def copy_data_var(
...
-    fs, path = _get_fs_and_path(primary_store)
-    fs.auto_mkdir = True
-
-    # We want to support local and s3fs filesystems. fsspec local filesystem is sync,
-    # but our s3fs from zarr.storage.FsspecStore is async and here we work around it.
-    # The AsyncFileSystem wrapper on LocalFilesystem raises NotImplementedError when _put is called.
-    source = f"{tmp_store / relative_dir}/"
-    dest = f"{path}/{relative_dir}"
-    fsspec_apply(fs, "put", source, dest, recursive=True, auto_mkdir=True)
+    for file in tmp_store.glob(f"{relative_dir}**/*"):
```
aldenks, line 115: this is likely going from parallel to not parallel. lets keep an eye on if this is impacting perf at all, but i think it should be fine

aldenks (approved):
Approved after comments!

src/reformatters/common/zarr.py
mosegontar: For posterity, we tracked this down and it looks like the CPU buffer will use `np.frombuffer` https://zarr.readthedocs.io/en/main/_modules/zarr/core/buffer/cpu.html#Buffer.from_bytes

which won't make a copy https://numpy.org/doc/2.1/reference/generated/numpy.frombuffer.html

aldenks (approved):

PR #197, opened 2025-08-07
UpdateProgressTracker Refactor

mosegontar:
This PR updates the `UpdateProgressTracker` as follows:

- Rather than storing progress tracking files within the on local filesystem/cloud storage zarr path, progress tracking files are written to an adjacent path. 
     -  For example, progress tracked files for `gs://upstream-gridded-zarrs/u-arizona-swann/v0.1.0.zarr` would be stored at `gs://upstream-gridded-zarrs/u-arizona-swann/v0.1.0_update_progress`.
- The progress tracker now takes a path to the store, rather than a store itself -- this allows us to avoid coming up with a way to grab the path off of the store. Because of this we can stop using `setattr` to add `fs` and `path` attributes on local zarr stores, which was a hack.
-  The `UpdateProgressTracker` now instantiates a new synchronous `fs` object that is synchronous and allows us to remove `common.fsspec` and `fsspec_apply`


src/reformatters/common/update_progress_tracker.py
```diff
@@ -104,7 +93,11 @@ def _process_queue(self) -> None:
                 content = json.dumps(
                     {PROCESSED_VARIABLES_KEY: list(self.processed_variables)}
                 )
-                fsspec_apply(self.fs, "pipe", self._get_path(), content.encode("utf-8"))
+
+                retry(
+                    lambda: self.fs.pipe(self._get_path(), content.encode("utf-8")),  # noqa: B023
```
mosegontar: noqa-ed because mypy seemed confused:

```

src/reformatters/common/update_progress_tracker.py:98:60: B023 Function definition does not bind loop variable `content`
    |
 97 |                 retry(
 98 |                     lambda: self.fs.pipe(self._get_path(), content.encode("utf-8")),
    |                                                            ^^^^^^^ B023
 99 |                     max_attempts=1,
100 |                 )
```

aldenks: I think this actually is a subtle bug. the way python lambdas capture variables updated in a loop is very unintuitive imo -- it creates and captures the variable on the _first_ iteration of the loop only(!?!). e.g.

```
>>> a = 1
>>> for _ in range(3):
...   print((lambda: a)())
... 
1
1
1
```

in this case I believe we'd keep writing the same content that was created in the first trip through the while loop.

lets add a test around this and one way to fix is to make an inline function def that both json.dumps the content and writes it and then call retry on that. that will (should?) work because the function will bind `self` which stays constant and then re-access `.processed_variables` each call.

aldenks: lol ignore my example, it doesn't change a. but this is a real thing, getting the details

aldenks: here's an example of what this is checking for: it only matters if you save the closure and use it _later_ outside of the loop. in this case we don't do that so i think you are right it is not an issue. that said, i kinda like using an approach that doesn't require us to noqa this.

```
>>> funcs = [lambda: i for i in range(3)]
>>> for f in funcs:
...   print(f())
... 
2
2
2
```
they all print 2 because the `i` in the lambda captures a _reference_ to the variable (not its contents) and then when you go to execute the functions they all point to i which now holds the last value that was written to it, 2.

src/reformatters/common/update_progress_tracker.py
```diff
@@ -38,21 +22,26 @@ class UpdateProgressTracker:
...
         reformat_job_name: str,
         time_i_slice_start: int,
+        store_path: str,
     ):
         self.reformat_job_name = reformat_job_name
         self.time_i_slice_start = time_i_slice_start
         self.queue: queue.Queue[str] = queue.Queue()
-        self.fs, self.path = _get_fs_and_path(store)
+
+        self.fs, self.path = fsspec.core.url_to_fs(
```
aldenks: rename self.path to self.update_progress_dir? feels clearer

src/reformatters/common/update_progress_tracker.py
```diff
@@ -104,7 +93,11 @@ def _process_queue(self) -> None:
                 content = json.dumps(
                     {PROCESSED_VARIABLES_KEY: list(self.processed_variables)}
                 )
-                fsspec_apply(self.fs, "pipe", self._get_path(), content.encode("utf-8"))
+
+                retry(
+                    lambda: self.fs.pipe(self._get_path(), content.encode("utf-8")),  # noqa: B023
+                    max_attempts=1,
```
aldenks: maybe 3? this is on a background thread so it's not blocking anything besides further updates to this file

aldenks (approved):
Approved after comments! _delighted_ to cut out our sync/async combo fsspec garbage

PR #187, opened 2025-07-25
HRRR Refactor 

acannistra:
⚠️ AI wrote a lot of this code, and I refined much of it. I realize it might be a big lift to review, so please don't rush! I'm going to be away from any ability to work on this for a week or so, but I wanted to create something to discuss. If there are too many wierd things here, I'd be happy to re-write now that I've had a chance to get more familiar with the new abstractions. 

---

  This pull request refactors the HRRR 48-hour forecast reformatter to use the new `DynamicalDataset` framework,
  following the steps outlined in the `dataset_integration_guide.md`. 

  Here's a summary of the changes, aligned with the integration guide:

  1. Dataset Registration
  The new `NoaaHrrrForecast48HourDataset` is registered in src/reformatters/__main__.py, making it available to the
  CLI.

  2. TemplateConfig Implementation
  The dataset's structure, including dimensions, coordinates, and data variables, is now defined in `src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py`. 

  3. `RegionJob` Implementation
  A new `NoaaHrrrForecast48HourRegionJob` class has been implemented in `src/reformatters/noaa/hrrr/forecast_48_hour/region_job.py`. This class encapsulates the logic for processing a
  region of the dataset, including:
   - Generating source file coordinates for HRRR's specific file structure.
   - Downloading + reading GRIB files from the NOAA HRRR S3 bucket with byte-range optimization (via functions in `read_data.py`).
   - Defining the logic for operational updates.

  4. `DynamicalDataset` Implementation
  The NoaaHrrrForecast48HourDataset class in src/reformatters/noaa/hrrr/forecast_48_hour/dynamical_dataset.py
  brings together the TemplateConfig and RegionJob implementations to define the complete dataset.

  5. Code Cleanup
  The old reformatting logic has been removed:
   - src/reformatters/noaa/hrrr/forecast_48_hour/cli.py
   - src/reformatters/noaa/hrrr/forecast_48_hour/reformat.py
   - src/reformatters/noaa/hrrr/forecast_48_hour/reformat_internals.py
   - src/reformatters/noaa/hrrr/forecast_48_hour/template.py

  Helper functions in src/reformatters/noaa/hrrr/read_data.py have been updated to support the new RegionJob
  implementation.

---

## Testing

I have to admit that I didn't have as much time to test this as I'd like to.

 I confirmed that I can perform an inplace local update e.g. 

```bash
uv run main noaa-hrrr-forecast-48-hour backfill-local \
[...]

PR #199, opened 2025-08-08
HRRR Refactor tweaks + updates

acannistra:
The HRRR refactor (#187) had a few requested changes (https://github.com/dynamical-org/reformatters/pull/187#pullrequestreview-3064966873) that I implemented in this PR, and a couple tweaks I noticed along the way. Specifically:

- Fixed `operational_update_jobs` to always require existing data for operational updates (fails if no prior data exists)
- Refactored `download_hrrr_file` to use existing utilities and coordinate URL methods (AI really spun out on that one 😄 )
- Simplified `generate_source_file_coords` logic by trusting data variable groupings and removing complex per-variable filtering, and leveraging existing `init_hour` sets from the template 
- Fixed HRRR data reading band matching to handle actual GRIB file descriptions vs dev placeholders (this was a bug I found) 
- Updated tests to use proper mocking instead of being network-dependent


aldenks:
@acannistra looks like we need the zarr template checked in. run `uv run main noaa-hrrr-forecast-48-hour update-template` and check in the new files (also good to give em a review to see if they contain what you expect)

acannistra:
oops! my bad. added those to the branch. I gave them a once-over and things look good. I remembered actually that I did punt on `spatial_ref` earlier - maybe need to go back and add that for rioxarray?

aldenks:
@acannistra im not going to get to reviewing this before my break, but i will asap the week of Aug 18. sorry for holding you up and thanks for getting on these quick!

src/reformatters/noaa/hrrr/read_data.py
```diff
@@ -345,6 +193,9 @@ def download_hrrr_file(
         ValueError: If variables don't match file type or lead time is invalid
         FileNotFoundError: If download fails
     """
+    # Import here to avoid circular imports
+    from reformatters.noaa.hrrr.forecast_48_hour.region_job import HRRRSourceFileCoord
```
aldenks, line 197: what if this whole method was on the RegionJob? moving it's contents into download_file would allow us to cut out a good bit of code (you'd already have the SourceFileCoord). I'm not too worried about re-use for a different HRRR dataset, we can extract when we get there (subclass? etc)

src/reformatters/noaa/hrrr/read_data.py
```diff
@@ -247,12 +60,47 @@ def read_hrrr_data(
...
+                elif "surface" in grib_index_level:
+                    # Look for surface indicators
+                    if "sfc" in description.lower() or "surface" in description.lower():
+                        matching_bands.append(band_idx)
+                        continue
+                else:
+                    # For other levels, try to match the level description
+                    if grib_index_level in description.lower():
+                        matching_bands.append(band_idx)
+                        continue
```
aldenks, line 103: is there info we could dump from the files into the data vars' internal_attrs that would allow us to do an exact match and remove most of this logic? internal attrs are designed to be any specific data we want that's useful when processing, no need to stick with grib_element and grib_description

src/reformatters/noaa/hrrr/forecast_48_hour/templates/latest.zarr/x/zarr.json
```diff
@@ -0,0 +1,75 @@
+{
+  "shape": [
+    1799
+  ],
+  "data_type": "float64",
+  "chunk_grid": {
+    "name": "regular",
+    "configuration": {
+      "chunk_shape": [
+        180
```
aldenks, line 10: we want to update the chunk size for the `x` and `y` coordinates in the template config so they each have a single chunk. this will help the speed of xr.open_zarr quite a bit. you can verify after updating the template, there should be only a single coordinate array called `x/c/0` and not `x/c/1`, `x/c/2`, etc

aldenks (approved):
Awesome! im going to merge and we're good to start testing this while you do the follow up comments. The last medium sized pieces of work after that are
- review variable metadata for standardization and subselect variables to start
- fine tune chunk sizes?
- do a backfill and validate data

I'm going to make the following small changes now and will drop that PR below for reference -- don't want to make you do another full turn just for little stuffs.
- let failures bubble up to reduce error handling code in region job
- use utils that will raise to reduce error handling code. e.g. item(set(has hour 0), item() vs next + iter for file types in region job
- can reduce code with product + list comp in generate source file coords

aldenks:
here's the follow up PR implementing a number of what would have been my smaller/nitpick comments. it also includes the chunk size update to x and y coords. https://github.com/dynamical-org/reformatters/pull/203

PR #201, opened 2025-08-08
UpdateProgressTracker Refactor with bug fix

mosegontar:
Reverts dynamical-org/reformatters#200 

`auto_mkdir` did not work for s3: 

```
AioSession.__init__() got an unexpected keyword argument 'auto_mkdir'
```

though it appears to work local file systems and GCS. This PR is a revert-of-the-revert, so it still contains the bug.


src/reformatters/common/update_progress_tracker.py
```diff
@@ -38,21 +22,26 @@ class UpdateProgressTracker:
...
+        store_path: str,
     ):
         self.reformat_job_name = reformat_job_name
         self.time_i_slice_start = time_i_slice_start
         self.queue: queue.Queue[str] = queue.Queue()
-        self.fs, self.path = _get_fs_and_path(store)
+
+        self.fs, self.update_progress_dir = fsspec.core.url_to_fs(
+            store_path.replace(".zarr", "_update_progress"),
+            auto_mkdir=True,
```
mosegontar: this is the problematic line

src/reformatters/common/update_progress_tracker.py
```diff
@@ -38,21 +22,28 @@ class UpdateProgressTracker:
...
         self.time_i_slice_start = time_i_slice_start
         self.queue: queue.Queue[str] = queue.Queue()
-        self.fs, self.path = _get_fs_and_path(store)
+
+        self.fs, self.update_progress_dir = fsspec.core.url_to_fs(
+            store_path.replace(".zarr", "_update_progress")
+        )
+
+        if isinstance(self.fs, fsspec.implementations.local.LocalFileSystem):
+            self.fs.makedirs(self.update_progress_dir, exist_ok=True)
```
mosegontar, line 38: this is the fix -- explicitly make directory if it's local, otherwise we should be able to write to the path on object storage.

PR #202, opened 2025-08-20
Load k8s secrets from mounted volume

mosegontar:
This PR refactors how storage options secrets (credentials needed to interface with cloud storage buckets) are loaded. Rather than relying on setting environment variables, we store a json object with the credentials in a mounted volume when a pod deploys. It is assumed that the credentials within the json are stored within a `storage_options.json` key.

### DIFF
--- a/src/reformatters/common/kubernetes.py
+++ b/src/reformatters/common/kubernetes.py
@@ -24,6 +24,8 @@ class Job(pydantic.BaseModel):
     pod_active_deadline: timedelta = timedelta(hours=6)
 
     secret_names: Sequence[str] = pydantic.Field(default_factory=list)
+    # This is to support legacy datasets that have not yet been ported to the DynamicalDataset pattern
+    # (GEFS forecast and GEFS analysis).
     env_var_secret_names: Sequence[str] = pydantic.Field(default_factory=list)
 
     @property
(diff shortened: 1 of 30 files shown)


src/reformatters/common/kubernetes.py
```diff
@@ -22,7 +23,8 @@ class Job(pydantic.BaseModel):
     ttl: timedelta = timedelta(days=1)
     pod_active_deadline: timedelta = timedelta(hours=6)
 
-    secret_names: Sequence[str] = []
+    secret_names: Sequence[str] = pydantic.Field(default_factory=list)
+    env_var_secret_names: Sequence[str] = pydantic.Field(default_factory=list)
```
aldenks, line 29: maybe add a comment about this being legacy for GEFS

src/reformatters/common/storage.py
```diff
@@ -19,9 +28,39 @@ class StorageConfig(FrozenBaseModel):
     """Configuration for the storage of a dataset in production."""
 
     base_path: str
-    k8s_secret_names: Sequence[str] = Field(default_factory=tuple)
+    k8s_secret_name: str = ""
```
aldenks: I'd type it as str | None = None

aldenks (approved):
Yay! the simplicity of use in the update progress tracker seems like good validation of the StoreFactory changes. Approved!

PR #204, opened 2025-08-22
Replica stores

mosegontar:
This PR implements support for writing Dynamical datasets to more than one store. Each dataset has a required `primary_storage_config` set, but optional `replica_storage_configs` can be defined as well. These replica configs define configuration for storage replicas (e.g., storage in a second or third bucket or provider).


src/reformatters/common/dynamical_dataset.py
```diff
@@ -136,16 +138,23 @@ def update(
...
+                copy_zarr_metadata(
+                    updated_template,
+                    tmp_store,
+                    primary_store,
+                    replica_stores=replica_stores,
+                )
 
         logger.info(
-            f"Operational update complete. Wrote to store: {self.primary_store_factory.store()}"
+            f"Operational update complete. Wrote to store: {self.store_factory.primary_store()} with {len(self.store_factory.replica_stores())} replicas"
```
aldenks: slight pref to log them all vs the count

src/reformatters/common/dynamical_dataset.py
```diff
@@ -43,13 +43,15 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
     template_config: TemplateConfig[DATA_VAR]
     region_job_class: type[RegionJob[DATA_VAR, SOURCE_FILE_COORD]]
 
-    storage_config: StorageConfig
+    primary_storage_config: StorageConfig
+    replica_storage_configs: list[StorageConfig] = Field(default_factory=list)
```
aldenks: would this work? that way it's not mutable and the specific container type isn't important
```suggestion
    replica_storage_configs: Sequence[StorageConfig] = Field(default_factory=tuple)
```

src/reformatters/common/storage.py
```diff
@@ -68,7 +68,8 @@ def _load_storage_options_locally(self) -> dict[str, Any]:
 
 
 class StoreFactory(FrozenBaseModel):
-    storage_config: StorageConfig
+    primary_storage_config: StorageConfig
+    replica_storage_configs: list[StorageConfig] = Field(default_factory=list)
```
aldenks: Same type question here re Sequence and default_factory=tuple

src/reformatters/common/zarr.py
```diff
@@ -41,23 +41,28 @@ def copy_data_var(
     append_dim: str,
     tmp_store: Path,
     primary_store: zarr.abc.store.Store,
+    replica_stores: list[zarr.abc.store.Store] | None = None,
```
aldenks: I think we really only need this to be Iterable[zarr...] and that will help us not mutate arguments without meaning to and make refactoring later easier

src/reformatters/common/zarr.py
```diff
@@ -41,23 +41,28 @@ def copy_data_var(
...
     track_progress_callback: Callable[[], None] | None = None,
 ) -> None:
+    if replica_stores is None:
```
aldenks: why not have the default kwarg be `()`?

src/reformatters/common/zarr.py
```diff
@@ -71,10 +76,26 @@ def copy_data_var(
...
+        key = str(file.relative_to(tmp_store))
+        sync_to_store(store, key, file.read_bytes())
+
+
 def copy_zarr_metadata(
-    template_ds: xr.Dataset, tmp_store: Path, primary_store: zarr.abc.store.Store
+    template_ds: xr.Dataset,
+    tmp_store: Path,
+    primary_store: zarr.abc.store.Store,
+    replica_stores: list[zarr.abc.store.Store] | None = None,
```
aldenks: same Q on type and default value here

aldenks (approved):
awesome. approved after comments.

design wise, how simple this is to implement feels like a good sign.

PR #205, opened 2025-08-22
HRRR: Spatial coordinates and dimension order + storage configuration

aldenks:
* Add new storage config for HRRR AWS open data bucket

* Small adjustments to commented HRRR resources

* y, x dimension order so plots are north-up by default and match other datasets

* float32 lat and lon coords

* Assign real x, y coords in meters

* Fill ingested forecast length with NaT by default


aldenks:
towards #81

PR #206, opened 2025-08-26
NOAA HRRR read_data fix: don't transpose result

mosegontar:
We encountered the following error when running a backfill:

```
could not broadcast input array from shape (1799,1059) into shape (1059,1799)
```

Removing the transpose call fixed.


aldenks (approved):
thanks for fixing my unfinished update

PR #207, opened 2025-08-27
Icechunk support

mosegontar:
This PR adds support for Icechunk to Dynamical. 

Currently, we are only implementing Icechunk support for replica stores. This is because for primary stores, we need to also write progress tracking files to cloud storage. The format of the kwargs that are used to initialize Icechunk storage with credentials is slightly different than the format of the kwargs passed to `fsspec` to grant cloud storage access (eg we would pass `key` to `fsspec` and `access_key_id`  to Icechunk). We should be able to write a utility function to translate between these forms, which should unblock using Icechunk as a primary store, but we have deferred this work for now.


src/reformatters/common/storage.py
```diff
@@ -146,17 +160,107 @@ def get_local_tmp_store() -> Path:
     return Path(f"data/tmp/{uuid4()}-tmp.zarr").absolute()
 
 
-def _get_store_path(dataset_id: str, version: str, base_path: str) -> str:
-    if not Config.is_prod:
+def _get_store_path(
+    dataset_id: str, version: str, storage_config: StorageConfig
+) -> str:
+    if Config.is_prod:
+        base_path = storage_config.base_path
```
mosegontar, line 167: inverted the logic here and in `_get_store` so the conditional is positive

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "xarray>=2025.1.2",
     "zarr>=3.0.4",
     "kubernetes>=33.1.0",
+    "icechunk==1.1.4",
```
aldenks: Let's allow to update with >=

src/reformatters/common/region_job.py
```diff
@@ -456,7 +457,13 @@ def get_jobs(
 
         return get_worker_jobs(all_jobs, worker_index, workers_total)
 
-    def process(self) -> Mapping[str, Sequence[SOURCE_FILE_COORD]]:
+    def process(
+        self,
+    ) -> tuple[
+        Mapping[str, Sequence[SOURCE_FILE_COORD]],
+        zarr.abc.store.Store,
```
aldenks, line 464: Is there a downside to committing within process()? I'm sure you thought about this

mosegontar: I think that works in the backfill case but not the operational update case, where we separately adjust the template and then copy metadata: 

```
                process_results, primary_store, replica_stores = job.process()
                updated_template = job.update_template_with_results(process_results)
                # overwrite the tmp store metadata with updated template
                template_utils.write_metadata(updated_template, tmp_store)
                copy_zarr_metadata(
                    updated_template,
                    tmp_store,
                    primary_store,
                    replica_stores=replica_stores,
                )

                storage.commit_if_icechunk(
 ```

So for simplicity and consistency committing outside of `process` in all cases seemed like the best option

Comment (aldenks) [reply src/reformatters/common/region_job.py:464]: Ah, sounds good

File: src/reformatters/common/dynamical_dataset.py [hunk 4]
```diff
@@ -153,6 +150,12 @@ def update(
                     replica_stores=replica_stores,
                 )
 
+                storage.commit_if_icechunk(
+                    f"Completed operational update for {job}",
```
aldenks: One more thought: let’s not put the job name in the public commit message. We could include an iso formatted process_results[append_dim].max() if that’s easy

mosegontar: Update: we discussed and are just including the timestamp of the commit in the message for now: https://github.com/dynamical-org/reformatters/pull/207/commits/f981ad35045fde3f7839835c10fb1ebd1df5652e

src/reformatters/common/dynamical_dataset.py
```diff
@@ -153,6 +150,12 @@ def update(
                     replica_stores=replica_stores,
                 )
 
+                storage.commit_if_icechunk(
+                    f"Automated update at {pd.Timestamp.now().isoformat()}",
```
aldenks: .now("UTC") so the message gets a tz offset

PR #210, opened 2025-08-29
HRRR download_file and spatial coordinate fixes, cleanup and optimize

aldenks:


tests/noaa/gfs/forecast/region_job_test.py
```diff
@@ -157,12 +157,12 @@ def test_operational_update_jobs(
 
     # Set the append_dim_end for the update
     monkeypatch.setattr(
-        pd.Timestamp, "now", classmethod(lambda cls: pd.Timestamp("2025-01-01T12:34"))
+        pd.Timestamp, "now", classmethod(lambda cls: pd.Timestamp("2022-01-01T12:34"))
```
aldenks, line 160: totally unrelated but this test was slow and this makes it faster

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -58,74 +57,51 @@ def dataset_attributes(self) -> DatasetAttributes:
...
+
         return {
             "init_time": self.append_dim_coordinates(
                 self.append_dim_start + self.append_dim_frequency
             ),
             "lead_time": pd.timedelta_range("0h", "48h", freq=pd.Timedelta("1h")),
-            "y": np.arange(1059),  # y-dimension for projected coordinates
-            "x": np.arange(1799),  # x-dimension for projected coordinates
+            "y": y_coords,
+            "x": x_coords,
```
aldenks, line 75: These were being overwritten in derive coordinates with their correct values, moved that computation to this method

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -340,11 +319,203 @@ def data_vars(self) -> Sequence[HRRRDataVar]:
...
-                    keep_mantissa_bits=10,
+                    grib_description="REFC:entire atmosphere",
+                    index_position=1,
+                    keep_mantissa_bits=default_keep_mantissa_bits,
                     grib_index_level="entire atmosphere",
                     hrrr_file_type="sfc",
                 ),
-            )
+            ),
+            # HRRRDataVar(
```
aldenks, line 329: Will add these back, stared adding them but then PR got too big

src/reformatters/noaa/hrrr/forecast_48_hour/region_job.py
```diff
@@ -136,31 +144,29 @@ def generate_source_file_coords(
...
-            var
-            for var in self.data_vars
-            if var.internal_attrs.hrrr_file_type == coord.file_type
-        ]
+        """Download a subset of variables from a HRRR file and return the local path."""
+        idx_url = coord.get_idx_url()
+        idx_local_path = http_download_to_disk(idx_url, self.dataset_id)
+
+        byte_range_starts, byte_range_ends = parse_hrrr_index_byte_ranges(
+            idx_local_path, coord.data_vars
```
aldenks, line 159: This is the most subtle change: we use coord.data_vars, vs all data vars in our dataset that match coord.file_type. We previously did not have data_vars on a coordinate, and therefore in download were downloading all variables for the hrrr file_type (e.g. sfc) in the dataset, regardless of if they were in the download group. this meant `max_vars_per_download_group` was ignored and we would have downloaded the same variables multiple times (the number of groups)

PR #218, opened 2025-09-04
Fix off+test off by one error in noaa grib parser and integrate into first dataset to use it (HRRR)

aldenks:


src/reformatters/noaa/noaa_grib_index.py
```diff
@@ -76,7 +78,6 @@ def grib_message_byte_ranges_from_index(
         # We fall back to adding a large value (+10 GiB) to get an end point since obstore
         # doesn't support omitting the end byte but will return bytes up to the end of the file.
         end = int(end_match) if end_match else start + 10 * (2**30)
-        end -= 1  # HTTP Range header endpoint is inclusive
```
aldenks, line 79: Bug fix: the return values here are meant to work with obstore's get_ranges method which expects end offsets to be _exclusive_.

tests/noaa/noaa_grib_index_test.py
```diff
@@ -45,6 +45,15 @@ def test_grib_index_geavg_s_f000() -> None:
     assert all(isinstance(e, int) and e > 0 for e in ends)
     assert all(start < stop for start, stop in zip(starts, ends, strict=True))
 
+    # First match is "PRES:surface:anl:ens mean"
+    assert starts[0] == 1969991
+    assert ends[0] == 2768859
```
aldenks, line 50: verified this value in the real index file fixture, confirms bug fix

src/reformatters/noaa/noaa_grib_index.py
```diff
@@ -63,7 +63,9 @@ def grib_message_byte_ranges_from_index(
         # 2nd capture which wraps to capture the byte start of the next line to get our end byte.
         # 2nd capture is optional because we need to support matching the final variable in the index.
         matches = re.findall(
-            rf"\d+:(\d+):{var_match_str}:(?:\n\d+:(\d+))?",
+            rf"\d+:(\d+):"  # row number : start byte offset (captured)
+            rf"{var_match_str}:"  # uniquely identify the variable we want
+            rf"(?:.*\n\d+:(\d+))?",  # end of line and wrap to capture next line's byte offset to get end byte (optional capture to handle last line of index)
```
aldenks, line 68: `.*` was also added to this uncaptured group to handle ensemble index files where there's something like `:ens mean` after the lead time

PR #219, opened 2025-09-04
HRRR read data updates

aldenks:


src/reformatters/common/region_job.py
```diff
@@ -651,7 +653,9 @@ def _read_and_write_one(coord: SOURCE_FILE_COORD) -> ArrayND[np.generic]:
                     updated_coords[index] = replace(
                         coord, status=SourceFileStatus.Succeeded
                     )
-                except Exception:
+                except Exception as e:
+                    if isinstance(e, AssertionError):
+                        raise
```
aldenks, line 658: I added these two re-raises. I think if an assert fails we don't want to log and continue in any case

src/reformatters/noaa/gfs/read_data.py
aldenks, line 1: unrelated but no longer used!

src/reformatters/noaa/hrrr/read_data.py
aldenks, line 1: All this is now just part of the region job

aldenks:
towards #81

src/reformatters/common/region_job.py
mosegontar: I think this makes sense. My only question would be if there are other exceptions that we would like to halt processing for, but given that there's definitely a set of exceptions where we _don't_ want to halt (404s, bad grib reads) this seems like a safe choice

src/reformatters/noaa/hrrr/forecast_48_hour/region_job.py
```diff
@@ -186,4 +203,4 @@ def _update_append_dim_end(cls) -> pd.Timestamp:
     def _update_append_dim_start(cls, existing_ds: xr.Dataset) -> pd.Timestamp:
         """Get the start time for operational updates based on existing data."""
         ds_max_time = existing_ds["init_time"].max().item()
-        return pd.Timestamp(ds_max_time) - pd.Timedelta(hours=6)
+        return pd.Timestamp(ds_max_time)
```
mosegontar, line 206: What was the reason to drop the 6 hour look back?

tests/noaa/hrrr/forecast_48_hour/dynamical_dataset_test.py
```diff
@@ -12,6 +16,129 @@ def dataset() -> NoaaHrrrForecast48HourDataset:
...
+    point_ds = backfill_ds.sel(init_time=init_time_start, lead_time="2h").isel(
+        x=1, y=-2
+    )
+
+    assert point_ds["temperature_2m"] == 22.875
+    assert point_ds["precipitation_surface"] == 0.16992188
+    assert point_ds["downward_short_wave_radiation_flux_surface"] == 8.1875
+
+    # Operational update
+    append_dim_end = pd.Timestamp("2018-07-14T00:00")  #
```
mosegontar:
```suggestion
    append_dim_end = pd.Timestamp("2018-07-14T00:00") 
```

tests/noaa/hrrr/forecast_48_hour/dynamical_dataset_test.py
```diff
@@ -12,6 +16,129 @@ def dataset() -> NoaaHrrrForecast48HourDataset:
     return NoaaHrrrForecast48HourDataset(primary_storage_config=NOOP_STORAGE_CONFIG)
 
 
+@pytest.mark.slow
+def test_backfill_local_and_operational_update(
+    dataset: NoaaHrrrForecast48HourDataset, monkeypatch: pytest.MonkeyPatch
+) -> None:
```
mosegontar, line 31: HRRR has an Icechunk replica (correct me if I'm wrong if we don't plan to keep that); would it be worth including an icechunk replica in the integration test or is this too slow? (Maybe not worth it -- maybe we use keep the GFS one as our icechunk replica integration test)

tests/noaa/hrrr/forecast_48_hour/dynamical_dataset_test.py
```diff
@@ -43,3 +170,7 @@ def test_validators(dataset: NoaaHrrrForecast48HourDataset) -> None:
     validators = tuple(dataset.validators())
     assert len(validators) == 3
     assert all(isinstance(v, validation.DataValidator) for v in validators)
+
+
+def assert_no_nulls(ds: xr.Dataset) -> None:
```
mosegontar: seems like maybe a useful utility generally

tests/noaa/hrrr/forecast_48_hour/dynamical_dataset_test.py
aldenks: Good call added. Although i think really that's something we should be checking in the common/dynamical_dataset_test.py

tests/noaa/hrrr/forecast_48_hour/dynamical_dataset_test.py
aldenks: :+1:  started tests/xarray_testing.py and put it in there

src/reformatters/noaa/hrrr/forecast_48_hour/region_job.py
aldenks: This will have us re-process the previous init time and then process the current.

If we process the current max init time in the dataset and anything newer, that will already be reprocessing an existing init time in addition to the new one. if any data exists for an init time, we know that means the init time before it is no longer being updated, meaning we don't need to reprocess 3 inits which is what subtracting 6 hours from the prev max would do.

src/reformatters/common/region_job.py
aldenks: I can't think of any specifically, and while it feels a little "uncautious" to have a specific list we hard stop processing on vs the other way around (a specific list we allow through), I don't think we have a good and complete list of all the possible ways a download or read could fail and we have higher level mechanisms to handle individual vs whole update failures (if some files fail they will leave nans and the available data will be present, if all files in an update fail the dataset will not be extended)

PR #226, opened 2025-09-07
ICON-EU step 2: Implement `region_job.py`

JackKelly:


aldenks:
@JackKelly sorry do make this change while you have an open PR but there is an API change to region job being merged, rebase to get the changes. https://github.com/dynamical-org/reformatters/pull/237

giving this a skim i think the only change you'll need in this PR to conform is to change `operational_update_jobs` to take a `primary_store: zarr.abc.store.Store` rather than a StoreFactory

src/reformatters/dwd/icon_eu/forecast/region_job.py
```diff
@@ -25,116 +32,95 @@
...
-    #
-    # max_vars_per_download_group: ClassVar[int | None] = None
-
-    # Implement this method only if different variables must be retrieved from different urls
-    #
-    # # @classmethod
-    # def source_groups(
-    #     cls,
-    #     data_vars: Sequence[DwdIconEuDataVar],
-    # ) -> Sequence[Sequence[DwdIconEuDataVar]]:
```
aldenks, line 68: because it looks like there's one file per variable, you can implement this to return a sequence of one element sequences (e.g. `[[var1], [var2], [var3], ...]`) because there's no ability/benefit from downloading multiple variables from the same file at the same time.

then youll get those sequences of individual variables passed into `generate_source_file_coords` and can do `grib_element = item(data_var_group).internal_attrs.grib_element`  (`from common.iterating import item` will check there's only one element)

src/reformatters/dwd/icon_eu/forecast/region_job.py
```diff
@@ -267,23 +251,22 @@ def operational_update_jobs(
...
-        #     reformat_job_name=reformat_job_name,
-        #     filter_start=append_dim_start,
-        # )
-        # return jobs, template_ds
-
-        raise NotImplementedError(
-            "Subclasses implement operational_update_jobs() with dataset-specific logic"
+        # TODO(Jack): The implementation below is the unmodified code from
+        #             the dynamical.org example code. I _think_ this is what
+        #             ICON-EU needs, but I'm not 100% sure TBH!
```
aldenks: looks reasonable to me!

aldenks (commented):
Nice progress!

JackKelly:
@aldenks wrote:

> i think the only change you'll need in this PR to conform is to change `operational_update_jobs` to take a `primary_store: zarr.abc.store.Store` rather than a StoreFactory

Done! That was easy :slightly_smiling_face:. I think this PR is up-to-date with the `main` now.

I'll work on getting this PR ready for merging (hopefully by the end of the week).

JackKelly:
Note to self: The tests pass & the code is up-to-date with the `main`. But I forgot to tell the code that it needs to un-bzip2 the grib files! So the command:

```shell
uv run main dwd-icon-eu-forecast backfill-local 2025-09-29 --filter-variable-names temperature_2m
```

currently fails with the following error:

`rasterio.errors.RasterioIOError: 'data/download/dwd-icon-eu-forecast/weather/nwp/icon-eu/grib/00/t_2m/icon-eu_europe_regular-lat-lon_single-level_2025092900_117_T_2M.grib2.bz2' not recognized as being in a supported file format.`

Hopefully I'll have time later this week to get this PR finished :slightly_smiling_face:

JackKelly:
OK, I think this PR is finally ready!  `uv run main dwd-icon-eu-forecast backfill-local 2025-10-03` runs locally :slightly_smiling_face:

aldenks (approved):
Wonderful!

Merging this now and here's a list of follow ups for us to remember
- [ ] add deaccumulation of precipitation by implementing apply_data_transformations on the region job ([example](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/noaa/gfs/forecast/region_job.py#L147-L166)
- [ ] makes more sense logically to decompress in the read_data step, but doesn't technically matter
- [ ] might as well store the data_var rather than the variable_name_in_filename on the source file coord, may be useful later to have the full info there
- [ ] fill out region job test coverage

PR #231, opened 2025-09-09
Refactor GEFS to use the new DynamicalDataset structure

samn:
Thank you Claude! Here's the plan for posterity 
[PLAN.md](https://github.com/user-attachments/files/22256147/PLAN.md)


src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,261 @@
...
+            coords.append(
+                GefsAnalysisEnsembleSourceFileCoord(
+                    init_time=init_time,
+                    ensemble_member=0,  # Control member for analysis
+                    lead_time=lead_time,
+                )
+            )
+        return coords
+
+    def _get_gefs_file_type(
```
aldenks: See common.iterating.item

src/reformatters/noaa/gefs/analysis/template_config.py
```diff
@@ -19,100 +21,89 @@
...
+            spatial_resolution="0.25 degrees (~20km)",
+            time_domain=f"{self.append_dim_start} UTC to Present",
+            time_resolution=f"{self.append_dim_frequency.total_seconds() / (60 * 60)} hours",
+        )
+
+    def dimension_coordinates(self) -> dict[str, Any]:
+        """Returns dimension coordinates for the dataset."""
+        return {
+            "time": self.append_dim_coordinates(
+                self.append_dim_start + self.append_dim_frequency
```
samn, line 62: Note: this used to extend time by 50 years, now it's 40 years (see https://github.com/dynamical-org/reformatters/blob/3a5480511e37dd93de7e8bd052e172a2ea138bb8/src/reformatters/common/template_config.py#L171)

This caused small differences in the template but not sure if it matters. Should I chagne this back to a 50 year window?

src/reformatters/noaa/gefs/forecast_35_day/templates/latest.zarr/expected_forecast_length/zarr.json
```diff
@@ -7,7 +7,7 @@
     "name": "regular",
     "configuration": {
       "chunk_shape": [
-        5475
+        7300
```
aldenks: We'll want to make sure this stays the same for a seamless update. you can do this by overriding https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/template_config.py#L164

src/reformatters/noaa/gefs/forecast_35_day/templates/latest.zarr/ingested_forecast_length/zarr.json
```diff
@@ -70,7 +70,7 @@
   },
   "dimension_names": [
     "init_time",
-    "ensemble_member"
+    "lead_time"
```
aldenks: this should be back to ensemble_member (no change), the suggestion in template_config.py will do that

src/reformatters/noaa/gefs/gefs_config_models.py
```diff
@@ -20,3 +25,220 @@ class GEFSInternalAttrs(NoaaInternalAttrs):
...
+
+class GefsEnsembleSourceFileCoord(GefsSourceFileCoord):
+    """Source file coordinate for GEFS forecast ensemble data."""
+
+    ensemble_member: int
+
+    def out_loc(self) -> Mapping[Dim, CoordinateValueOrRange]:
+        """Use existing coordinate structure directly for forecast dataset."""
+        return {
+            # "time": self.init_time + self.lead_time,
```
aldenks: we don't need/want this line, can remove from both of these source file coords

tests/noaa/gefs/analysis/dynamical_dataset_test.py
```diff
@@ -0,0 +1,92 @@
...
+from tests.common.dynamical_dataset_test import NOOP_STORAGE_CONFIG
+
+
+@pytest.fixture
+def dataset() -> GefsAnalysisDataset:
+    return GefsAnalysisDataset(
+        primary_storage_config=NOOP_STORAGE_CONFIG,
+        region_job_class=GefsAnalysisRegionJob,
+    )
+
```
aldenks, line 16: you can follow the tests/example/dynamical_dataset_test.py for a backfill and update test template

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+
+class GefsAnalysisRegionJob(RegionJob[GEFSDataVar, GefsAnalysisSourceFileCoord]):
+    """RegionJob for GEFS Analysis dataset processing."""
+
+    # 1 makes logic simpler when accessing GEFSv12 reforecast which has a file per variable
+    max_vars_per_backfill_job = 1
+
+    def get_processing_region(self) -> slice:
```
samn, line 72: Does this look right?

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,270 @@
...
+        Group variables by GEFS file type and ensemble statistic.
+        """
+        grouper: dict[
+            tuple[GEFSFileType, EnsembleStatistic | None, bool], list[GEFSDataVar]
+        ] = defaultdict(list)
+        for data_var in data_vars:
+            gefs_file_type = data_var.internal_attrs.gefs_file_type
+            ensemble_statistic = data_var.attrs.ensemble_statistic
+            var_has_hour_0_values = has_hour_0_values(data_var)
+            grouper[(gefs_file_type, ensemble_statistic, var_has_hour_0_values)].append(
```
samn: is this right?

src/reformatters/noaa/gefs/analysis/template_config.py
```diff
@@ -19,100 +21,89 @@
...
+    append_dim_start: Timestamp = pd.Timestamp("2000-01-01T00:00")
+    append_dim_frequency: Timedelta = pd.Timedelta("3h")
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def dataset_attributes(self) -> DatasetAttributes:
+        """Dataset metadata attributes."""
+        return DatasetAttributes(
+            dataset_id="noaa-gefs-analysis",
+            dataset_version="0.1.2",
```
samn, line 39: Should we bump the versions?

src/reformatters/noaa/gefs/common_gefs_template_config.py
```diff
@@ -69,15 +69,25 @@ def get_shared_coordinate_configs() -> Sequence[Coordinate]:
             encoding=Encoding(
                 dtype="int64",
                 fill_value=0,
-                chunks=1,  # Scalar coordinate
-                shards=1,
+                chunks=(),  # Scalar coordinate
```
samn, line 72: Needed to change these to avoid errors about the chunk & shard shapes not being compatible

src/reformatters/noaa/gefs/forecast_35_day/template_config.py
```diff
@@ -20,210 +23,242 @@
...
+                    compressors=[BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE],
+                    units="seconds",
+                    chunks=(
+                        append_dim_coordinate_chunk_size,
+                        len(dim_coords["lead_time"]),
+                    ),
+                    shards=(
+                        append_dim_coordinate_chunk_size,
+                        len(dim_coords["lead_time"]),
+                    ),
```
aldenks: this should be `len(dim_coords["ensemble_member"])` in the 2nd dim

src/reformatters/noaa/gefs/forecast_35_day/template_config.py
```diff
@@ -20,210 +23,242 @@
...
+            Coordinate(
+                name="init_time",
+                encoding=Encoding(
+                    dtype="int64",
+                    fill_value=0,
+                    compressors=[BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE],
+                    calendar="proleptic_gregorian",
+                    units="seconds since 1970-01-01 00:00:00",
+                    chunks=append_dim_coordinate_chunk_size,
+                    shards=append_dim_coordinate_chunk_size,
```
aldenks: we want shards=None for all the coords

src/reformatters/noaa/gefs/analysis/template_config.py
aldenks: updating the version would require us to reprocess the whole thing which is expensive for GEFS _forecast 35 day_. We should be able to match behavior and not need to update it

aldenks (commented):
awesome! this isn't an incomplete review, but i saw your questions in slack and thought id respond to those

src/reformatters/noaa/gefs/analysis/template_config.py
aldenks: yes, in order to not bump the version we do want to have no changes in the template config, including in coord array length

src/reformatters/noaa/gefs/analysis/region_job.py
aldenks: I don't think data_var.attrs.ensemble_statistic is an attribute. We do want to group by gefs_file_type and var_has_hour_0_values.

You can remove ensemble statistic from anywhere in the GEFS code for the datasets we have, it's a relic from when i was thinking we would put ensemble statistic variables into the same dataset as the ensemble itself (instead it wll be a new dataset, noaa-gefs-forecast-35-day-statistics with data variables that share the same name for easier interop between GEFS datasets.

src/reformatters/noaa/gefs/analysis/region_job.py
aldenks: yes but you can skip the min with the len of the template (because a slice off the end just slices through the end) and copy the logic here https://github.com/dynamical-org/reformatters/pull/231/files#diff-eff2a5db70ad18de22be8a26e1197630a0e544d84da14a27f1fd14057cf620d9L153

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,270 @@
...
+        self, processing_region_ds: xr.Dataset, data_var_group: Sequence[GEFSDataVar]
+    ) -> Sequence[GefsAnalysisSourceFileCoord]:
+        """Generate source file coordinates for analysis data from forecast files."""
+
+        times = filter_available_times(
+            pd.to_datetime(processing_region_ds["time"].values)
+        )
+
+        # Check if any variable in the group doesn't have hour 0 values
+        var_has_hour_0_values = any(has_hour_0_values(var) for var in data_var_group)
```
aldenks: use common.iterating.item vs any to ensure that all vars in the group either have or don't have hour 0 values

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,270 @@
...
+            # Sort by index position for consistent ordering
+            idx_data_vars = sorted(
+                idx_data_vars, key=lambda dv: dv.internal_attrs.index_position
+            )
+            groups.append(idx_data_vars)
+
+        # Sort groups for consistent ordering
+        return sorted(groups, key=lambda g: str(g[0].internal_attrs.gefs_file_type))
+
+    def generate_source_file_coords(
```
aldenks, line 89: I'd be more confident in the logic in here if we copied what was here https://github.com/dynamical-org/reformatters/pull/231/files#diff-eff2a5db70ad18de22be8a26e1197630a0e544d84da14a27f1fd14057cf620d9L233-L273

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def download_file(self, coord: GefsAnalysisSourceFileCoord) -> Path:
+        """Download the source file for the given coordinate."""
+        # Download grib index file
+        idx_url = f"{coord.get_url()}.idx"
+        idx_local_path = http_download_to_disk(idx_url, self.dataset_id)
+        index_contents = idx_local_path.read_text()
+
+        # Download the grib messages for the data vars in the coord using byte ranges
+        starts, ends = parse_grib_index(index_contents, coord)
```
aldenks: We have a new common to NOAA in general grib parser, try using that https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/noaa/noaa_grib_index.py#L12

src/reformatters/noaa/gefs/forecast_35_day/region_job.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def download_file(self, coord: GefsForecast35DaySourceFileCoord) -> Path:
+        """Download the source file for the given coordinate."""
+        # Download grib index file
+        idx_url = f"{coord.get_url()}.idx"
+        idx_local_path = http_download_to_disk(idx_url, self.dataset_id)
+        index_contents = idx_local_path.read_text()
+
+        # Download the grib messages for the data vars in the coord using byte ranges
+        starts, ends = parse_grib_index(index_contents, coord)
```
aldenks: try using this https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/noaa/noaa_grib_index.py#L12  I think we can remove the gefs specific grib parsing

src/reformatters/noaa/gefs/forecast_35_day/region_job.py
```diff
@@ -0,0 +1,188 @@
...
+        Note: forecast version doesn't include has_hour_0_values in grouping.
+        """
+        grouper: dict[GEFSFileType, list[GEFSDataVar]] = defaultdict(list)
+        for data_var in data_vars:
+            gefs_file_type = data_var.internal_attrs.gefs_file_type
+            grouper[gefs_file_type].append(data_var)
+
+        groups = []
+        for idx_data_vars in grouper.values():
+            # Sort by index position for consistent ordering
```
aldenks:
```suggestion
            # Sort by index position for better coalescing of byte range requests to the grib
```

src/reformatters/noaa/gefs/forecast_35_day/region_job.py
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Sequence[Sequence[GEFSDataVar]]:
+        """
+        Group variables by GEFS file type and ensemble statistic.
+
+        Note: forecast version doesn't include has_hour_0_values in grouping.
+        """
+        grouper: dict[GEFSFileType, list[GEFSDataVar]] = defaultdict(list)
+        for data_var in data_vars:
+            gefs_file_type = data_var.internal_attrs.gefs_file_type
+            grouper[gefs_file_type].append(data_var)
```
aldenks: we also want to add in has_hour_0_values to the grouping. that logic was subtly in a filtering step before. Then in generate_source_file_coords each call will either have or not have hour 0 values and we can create the appropriate coords (and our `item` call will always succeed)

src/reformatters/noaa/gefs/forecast_35_day/region_job.py
```diff
@@ -0,0 +1,188 @@
...
+    ) -> None:
+        """
+        Apply transformations to data array in place.
+
+        Based on existing transformation logic, but forecast data doesn't need
+        the same missing value interpolation as analysis data.
+        """
+        if data_var.internal_attrs.deaccumulate_to_rate:
+            reset_freq = data_var.internal_attrs.window_reset_frequency
+            if reset_freq is not None:
```
aldenks: assert reset_freq is not None if deaccumulate_to_rates is true

src/reformatters/noaa/gefs/forecast_35_day/region_job.py
```diff
@@ -0,0 +1,188 @@
...
+                try:
+                    deaccumulate_to_rates_inplace(
+                        data_array,
+                        dim="lead_time",
+                        reset_frequency=reset_freq,
+                    )
+                except ValueError:
+                    # Log and continue - errors are expected for some data
+                    pass
```
aldenks: log.exception so we get notified

src/reformatters/noaa/gefs/gefs_config_models.py
```diff
@@ -20,3 +25,218 @@ class GEFSInternalAttrs(NoaaInternalAttrs):
...
+
+        Note: Statistic variables in forecast dataset don't have a 'statistic' dimension.
+        They are separate data variables that don't include the ensemble_member dimension.
+        The statistic coordinate is used for file selection during download/read,
+        but the output location only uses init_time and lead_time.
+        """
+        return {
+            "init_time": self.init_time,
+            "lead_time": self.lead_time,
+        }
```
aldenks: We aren't technically using this yet but we'll have a statistics dimension to write into
```suggestion
        return {
            "init_time": self.init_time,
            "lead_time": self.lead_time,
            "statistic": self.statistic,
        }
```

aldenks (commented):
nice nice nice

src/reformatters/noaa/gefs/gefs_config_models.py
samn: But `statistic` isn't on ` reformatters.common.types.Dim`, do we want that everywhere?

src/reformatters/noaa/gefs/analysis/region_job.py
samn: Updated this to use that logic but oddly hit some type errors, modified the original code a little bit to account for that (some explicit calls to the `pd.DatetimeIndex` constructor, and using `pd.concat` since `append` is deprecated)

src/reformatters/noaa/gefs/forecast_35_day/template_config.py
samn: Sorry I'm a bit lost here, do you mean 

```python
                    shards=(
                        append_dim_coordinate_chunk_size,
                        len(dim_coords["ensemble_member"]),
                    ),
```

I'm getting a validation error because `Each shard size (31) must be a multiple of its corresponding chunk size (181) `

tests/noaa/gefs/forecast_35_day/dynamical_dataset_test.py
```diff
@@ -0,0 +1,187 @@
...
+            init_time_end,
+            freq=dataset.template_config.append_dim_frequency,
+            inclusive="left",
+        ),
+    )
+    subset_ds = updated_ds.sel(latitude=0, longitude=0, method="nearest")
+    np.testing.assert_array_equal(
+        subset_ds["temperature_2m"].values,
+        [
+            np.nan,
```
samn: The first element being `nan` points to a bug, right?

src/reformatters/noaa/gefs/gefs_config_models.py
```diff
@@ -20,3 +25,211 @@ class GEFSInternalAttrs(NoaaInternalAttrs):
...
+def is_v12_index(times: pd.DatetimeIndex) -> np.ndarray[Any, np.dtype[np.bool]]:
+    return (times < GEFS_REFORECAST_END) | (GEFS_CURRENT_ARCHIVE_START <= times)
+
+
+def get_grib_element(var_info: GEFSDataVar, init_time: pd.Timestamp) -> str:
+    grib_element = var_info.internal_attrs.grib_element
+    if init_time < GEFS_REFORECAST_END:
+        return GEFS_REFORECAST_GRIB_ELEMENT_RENAME.get(grib_element, grib_element)
+    else:
+        return grib_element
```
aldenks, line 87: I think this same code is also in gefs/read_data.py

src/reformatters/noaa/gefs/analysis/template_config.py
```diff
@@ -19,100 +21,98 @@
...
+            Coordinate(
+                name="time",
+                encoding=Encoding(
+                    dtype="int64",
+                    fill_value=0,
+                    compressors=[BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE],
+                    calendar="proleptic_gregorian",
+                    units="seconds since 1970-01-01 00:00:00",
+                    chunks=append_dim_coordinate_chunk_size,
+                    shards=append_dim_coordinate_chunk_size,
```
aldenks: shards=None on all coordinates

src/reformatters/noaa/gefs/forecast_35_day/templates/latest.zarr/ingested_forecast_length/zarr.json
```diff
@@ -9,7 +9,7 @@
     "configuration": {
       "chunk_shape": [
         5475,
-        181
+        31
```
aldenks, line 12: this was previously "wrong" in that it was longer than necessary but had no practical impact because the shape was and still is (N, 31)

src/reformatters/noaa/gefs/forecast_35_day/templates/latest.zarr/ingested_forecast_length/zarr.json
```diff
@@ -22,42 +22,19 @@
   "fill_value": 0,
   "codecs": [
     {
-      "name": "sharding_indexed",
+      "name": "bytes",
```
aldenks, line 25: known and expected change, we don't need/want small coordinate arrays to be sharded. we can see small reductions in coordinate array sizes as well in this pr as expected

src/reformatters/noaa/gefs/analysis/region_job.py
```diff
@@ -0,0 +1,244 @@
...
+            if reset_freq is not None:
+                try:
+                    deaccumulate_to_rates_inplace(
+                        data_array,
+                        dim="time",
+                        reset_frequency=reset_freq,
+                        skip_step=expected_missing,
+                    )
+                except ValueError:
+                    # Log and continue - errors are expected for some data
```
aldenks: log.exception here so we are notified

aldenks (commented):
Last thing in addition to my comments is adding in the progress tracker which you're already doing

samn:
Temporarily adjusting the dataset versions to run a backfill in parallel to production, will revert https://github.com/dynamical-org/reformatters/pull/231/commits/8425e49613916ebd4a9732432dc05ab51c24d164 once everything's validated

aldenks (approved):
YES

thank you

PR #234, opened 2025-09-11
Default validator checking that replica stores are in sync with primary store

mosegontar:


src/reformatters/common/validation.py
```diff
@@ -156,3 +157,46 @@ def check_analysis_recent_nans(
...
+    if problem_coords:
+        message = f"Data in replica and primary stores are different for coords: {problem_coords}"
+        return ValidationResult(passed=False, message=message)
+
+    num_variables_to_check = min(5, len(primary_ds.data_vars))
+    variables_to_check = np.random.choice(
+        list(primary_ds.data_vars.keys()), num_variables_to_check
+    )
+
+    last_chunk = iterating.dimension_slices(primary_ds, append_dim, "chunks")[-1]
```
mosegontar, line 182: Open to thoughts on default configuration here. How many variables do we want to check (currently set to no more than 5)? How many chunks should be checked? The downside of doing a single chunk and a subset of variables is that we just don't check the other variables at all. Maybe that's fine. FWIW, if we had a look back period (say most recent N chunks) then there's a better chance we'll cover both more time and more variables. But of course that'd be slower.

mosegontar: We could also alternatively scope to a single point lat/lon and` check more variables, but this requires more dataset specific knowledge (eg HRRR doesn't have lat/lon dims, though it does have them as derived coordinates)

src/reformatters/common/validation.py
```diff
@@ -156,3 +157,46 @@ def check_analysis_recent_nans(
         passed=True,
         message=f"All variables have acceptable NaN percentages (<{max_nan_percentage}%) in sampled location of latest data",
     )
+
+
+def compare_replica_and_primary(
```
mosegontar, line 162: For GFS, this validator took about 60ish seconds to run.

src/reformatters/common/validation.py
```diff
@@ -156,3 +157,46 @@ def check_analysis_recent_nans(
...
+    except AssertionError as e:
+        log.exception(e)
+        return ValidationResult(
+            passed=False,
+            message=f"Data in replica and primary stores are different for chunks: {last_chunk}",
+        )
+
+    return ValidationResult(
+        passed=True,
+        message="Data in replica and primary stores are the same",
```
aldenks:
```suggestion
        message="Data in tested subset of replica and primary stores is the same",
```

src/reformatters/common/validation.py
```diff
@@ -156,3 +157,46 @@ def check_analysis_recent_nans(
...
+        list(primary_ds.data_vars.keys()), num_variables_to_check
+    )
+
+    last_chunk = iterating.dimension_slices(primary_ds, append_dim, "chunks")[-1]
+    replica_ds_last_chunk = replica_ds.isel({append_dim: last_chunk})[
+        variables_to_check
+    ]
+    primary_ds_last_chunk = primary_ds.isel({append_dim: last_chunk})[
+        variables_to_check
+    ]
```
aldenks: selecting variables then isel might look aesthetically a little more expected

src/reformatters/common/validation.py
aldenks: i think how you have it is a good compromise. We may need to extend the time limit for the GEFS forecast validation pod but that'll become obvious once we have a replica

aldenks (approved):
Great!

PR #237, opened 2025-09-12
Pass primary and replica stores to RegionJob.process and remove progress tracker

mosegontar:
This PR removes `store_factory` as a field on RegionJob. Instead callers (DynamicalDataset instances) will pass instantiated `primary` and `replica` stores to `RegionJob.process` directly. This simplifies `RegionJob` and its responsibilities and it also keeps all of our Icechunk specific handling (store reinstantiation, metadata updates, and commits) within DynamicalDataset.

As part of this work, the progress tracker has been removed. It is incompatible with how we need to manage Icechunk stores. For most of our current datasets, processing is fast enough that we can afford to redo a little work on retries. For  GEFS which does need a progress tracker, it is currently using a legacy version of the tracker, which has not been removed. Ultimately we plan to refactor operational updates to allow for parallelization (mitigating the need for a progress tracker), which will unblock GEFS from being ported over to the RegionJob pattern.


src/reformatters/common/dynamical_dataset.py
```diff
@@ -139,8 +140,27 @@ def update(
             template_utils.write_metadata(template_ds, tmp_store)
 
             for job in jobs:
-                process_results, primary_store, replica_stores = job.process(
-                    expand_icechunk_stores=True
+                # We need to reinstantiate the primary store and replica stores
+                # for each job. This is because if they are an IcechunkStore,
+                # the associated Icechunk session will become read_only after
+                # the first job commits.
```
```suggestion
                # New store associated with an uncommitted icechunk session for each job
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -139,8 +140,27 @@ def update(
...
+                primary_store = self.store_factory.primary_store()
+                replica_stores = self.store_factory.replica_stores()
+
+                icechunk_stores = [
+                    s for s in replica_stores if isinstance(s, IcechunkStore)
+                ]
+                if isinstance(primary_store, IcechunkStore):
+                    icechunk_stores.append(primary_store)
+                for icechunk_store in icechunk_stores:
```
aldenks: factor out with the similar patterns in commit_if_icechunk in some way?

mosegontar: Yeah sounds good. I debated a bit about it, since the need to do it is tied very specifically to this update flow, but I think it will make this higher level code cleaner :+1:

src/reformatters/common/dynamical_dataset.py
mosegontar: Technically you can't tell whether you actually have an icechunk store there , but yeah can tweak this to make it more concise

mosegontar: Maybe 

> New stores to ensure we have an uncommitted icechunk session for each job, if any of the stores are Icechunk stores.

aldenks (approved):
Ship it!

PR #239, opened 2025-09-15
Call write_metadata before copy_zarr_metadata

mosegontar:


mosegontar:
Fix went out with https://github.com/dynamical-org/reformatters/pull/240

PR #241, opened 2025-09-16
Add default validator: check for expected shards

mosegontar:
This PR adds a default validator to be run on all stores checking that all expected shards (based on dataset dimension sizes and encoding) actually exist in the store. The intention here is to catch a case where some region job were to fail to write a shard.

### DIFF
--- a/src/reformatters/common/validation.py
+++ b/src/reformatters/common/validation.py
@@ -1,5 +1,7 @@
+import itertools
 from collections.abc import Sequence
 from datetime import timedelta
+from functools import partial
 from typing import Protocol, runtime_checkable
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


src/reformatters/common/validation.py
```diff
@@ -200,3 +203,53 @@ def compare_replica_and_primary(
...
+
+def check_for_expected_shards(
+    store: zarr.abc.store.Store, ds: xr.Dataset
+) -> ValidationResult:
+    """Check that the expected shards are present in the store."""
+    log.info(f"Checking for expected shards in {store}")
+
+    shard_counts_per_dim = [
+        len(iterating.dimension_slices(ds, str(dim), "shards"))
+        for dim in ds.sizes.keys()
```
aldenks: I think we got lucky on ordering here, dims on a Dataset are unordered. They are ordered on a DataArray though, so i'd move this down into the loop over vars and then you can do data_array.dims (a tuple of string dim names)

aldenks (approved):
Approved after comment!

PR #243, opened 2025-09-18
Adopt a number of new ruff check rules

aldenks:


aldenks:
Waiting until #231 merges to rebase and apply fixes to avoid making conflicts in files that will be deleted

PR #246, opened 2025-09-18
Hrrr full initial variables

aldenks:


aldenks:
replaced by https://github.com/dynamical-org/reformatters/pull/248

PR #248, opened 2025-09-18
HRRR: Round out initial set of data variables

aldenks:
Values pulled from gdalinfo/grib indexes and cross checked w/ GFS template config. Key differences
- Pressure reduced to mean sea level is grib element MSLMA in HRRR not PRMSL. This difference won't be visible to our users because the variable name, short_name, long_name and units are the same.
- Some variables in HRRR are instantaneous while they are trailing last 3/6h avgs in GFS
- HRRR doesn't have 100m winds but it does have 80m so including those


aldenks:
towards #81

PR #250, opened 2025-09-22
Check all AppendDims when ignoring 404s from recent files

samn:
Fixes https://dynamical.sentry.io/issues/6894613970/

The append dim on a `GefsAnalysisSourceFileCoord` is `time`, it's not present as a field on the class. This change correctly checks that a source file coord is recent, silencing errors about missing files that haven't been created yet for noaa-gefs-analysis.


aldenks (approved):
<3

PR #253, opened 2025-09-22
Trim off the last hour for GEFS Analysis in update_template_with_results

samn:
since not all variables has a 6 hour value (only precip), which causes nans when the template extends past that.


aldenks (approved):
Great

PR #260, opened 2025-09-23
Send logs to Sentry

samn:
https://docs.sentry.io/platforms/python/logs/

### DIFF
--- a/src/reformatters/__main__.py
+++ b/src/reformatters/__main__.py
@@ -1,4 +1,5 @@
 import multiprocessing
+import os
 from collections.abc import Sequence
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


src/reformatters/__main__.py
```diff
@@ -104,10 +105,13 @@ class UpstreamGriddedZarrsDatasetStorageConfig(StorageConfig):
         project_root="src/",
         in_app_include=["reformatters"],
         default_integrations=True,
+        enable_logs=True,
         integrations=[
             TyperIntegration(),
         ],
     )
+    sentry_sdk.set_tag("env", Config.env.value)
+    sentry_sdk.set_tag("job_name", os.getenv("JOB_NAME"))
```
aldenks, line 114: would be great to have the _cronjob_ as a tag too, but that seems not quite trivial to thread through. bonus for later if sentry logs is good

aldenks (approved):

src/reformatters/__main__.py
samn: done!

PR #262, opened 2025-09-23
Use pytest-xdist to run tests in parallel on all cores

samn:
This saves about 90 seconds when running all tests, though it's slower when only running a few tests (due to the fixed overhead of spawning worker processes). To run tests serially use `uv run pytest -n 0`.


aldenks (approved):

aldenks:
Thank ya!

PR #263, opened 2025-09-24
Nodepool: use generation >= 8 aws nodes

aldenks:
They are faster and a better value for our workloads than gen 7


aldenks:
This has already been applied with `kubectl apply -f deploy/aws/nodepool.yaml`

PR #264, opened 2025-09-24
Validation plots: use obstore, handle edge cases, little cleanup

aldenks:


aldenks:
going to gopher it, lower risk plotting code

PR #267, opened 2025-09-25
Temporarily set encoding fill value for GEFS and GFS to 0

mosegontar:


src/reformatters/noaa/gfs/forecast/template_config.py
```diff
@@ -264,7 +264,11 @@ def data_vars(self) -> Sequence[NoaaDataVar]:
 
         encoding_float32_default = Encoding(
             dtype="float32",
-            fill_value=np.nan,
+            # Note: fill_value was previously 0, which caused a subtle bug:
+            # for some categorical variables, shards filled with 0 (a valid value) were skipped (zarr writes default to write_empty_chunks=False),
+            # so unwritten shards were interpreted as all 0. After upgrading to use np.nan as fill_value, unwritten shards are now interpreted as all nan,
+            # which is incorrect for those variables. For now, we're setting fill_value to 0 to ensure that those values are read correctly.
+            fill_value=0,
```
aldenks: "While in general we use nan as a fill_value, these datasets were backfilled with fill_value = 0 and write_missing_chunks defaulting to false so we retain the 0 fill value"

src/reformatters/noaa/gefs/common_gefs_template_config.py
```diff
@@ -100,7 +100,11 @@ def get_shared_data_var_configs(
 
     encoding_float32 = Encoding(
         dtype="float32",
-        fill_value=np.nan,
+        # Note: fill_value was previously 0, which caused a subtle bug:
+        # for some categorical variables, shards filled with 0 (a valid value) were skipped (zarr writes default to write_empty_chunks=False),
+        # so unwritten shards were interpreted as all 0. After upgrading to use np.nan as fill_value, unwritten shards are now interpreted as all nan,
+        # which is incorrect for those variables. For now, we're setting fill_value to 0 to ensure that those values are read correctly.
+        fill_value=0,
```
aldenks, line 107: We should be seeing encoding changes in the zarr.json files for these variables reflecting this change right? im not seeing those, i would expect to see an updated fill_value in each encoding

aldenks (approved):

PR #269, opened 2025-09-25
Integration test to ensure that Dynamical dataset template configs always serialize correctly

mosegontar:
This PR adds a a test that serializes all datasets listed in DYNAMICAL_DATASETS and ensures that the resulting template matches the official, committed `latest.zarr` template for that dataset. 

As part of this, individual dataset tests doing the same work are removed.


aldenks (approved):
Beauty

PR #270, opened 2025-09-25
Include cron/job name in Sentry log attributes

samn:
So it's easier to search for all logs for a [Cron]Job.

https://docs.sentry.io/platforms/python/logs/#before_send_log


samn:
We pair reviewed this and Alex approved over my virtual shoulder

PR #271, opened 2025-09-26
get_template() -> write_metadata() round trip

aldenks:


aldenks:
will re-open when ready to work on this

PR #278, opened 2025-09-30
Implement overwrite_existing flag and add env guards

mosegontar:
This PR

* Adds an `overwrite_existing` flag that can be passed to `backfill_kubernetes`
* Adds env guards for `backfill_local` and `backfill_kubernetes`.
   * The guard for backfill kubernetes is wrapped in a new private method that can be directly monkeypatched without dangerous monkeypatching the actual env (which controls what stores we load etc).

### DIFF
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -213,12 +213,21 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         filter_contains: list[datetime] | None = None,
         filter_variable_names: list[str] | None = None,
         docker_image: str | None = None,
+        overwrite_existing: bool = False,
     ) -> None:
         """Run dataset reformatting using Kubernetes index jobs."""
+        assert self._can_run_in_kubernetes(), (
+            "backfill_kubernetes is only supported in prod environment"
+        )
+
         image_tag = docker_image or docker.build_and_push_image()
 
         template_ds = self._get_template(append_dim_end)
-        template_utils.write_metadata(template_ds, self.store_factory)
+
+        if overwrite_existing:
+            log.info("Writing to existing store")
+        else:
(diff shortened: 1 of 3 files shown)


src/reformatters/common/dynamical_dataset.py
```diff
@@ -444,6 +457,9 @@ def _tmp_store(self) -> Path:
     def _get_template(self, append_dim_end: DatetimeLike) -> xr.Dataset:
         return self.template_config.get_template(pd.Timestamp(append_dim_end))
 
+    def _can_run_in_kubernetes(self) -> bool:
+        return Config.is_prod
```
aldenks, line 465:
```suggestion
        # This method exists to support testing without changing the Config.env
        return Config.is_prod
```

tests/common/dynamical_dataset_test.py
```diff
@@ -448,3 +453,50 @@ def fail_resources(image_tag: str) -> None:
     # this should not raise
     with dataset._monitor(ReformatCronJob, "job"):
         pass
+
```
aldenks, line 457: Will you add a test that it raises assertion error if you violate either of the backfill local or back fill k8s asserts?

src/reformatters/common/dynamical_dataset.py
```diff
@@ -213,12 +213,21 @@ def backfill_kubernetes(
...
+            "backfill_kubernetes is only supported in prod environment"
+        )
+
         image_tag = docker_image or docker.build_and_push_image()
 
         template_ds = self._get_template(append_dim_end)
-        template_utils.write_metadata(template_ds, self.store_factory)
+
+        if overwrite_existing:
+            log.info("Writing to existing stores, skipping metadata write.")
```
aldenks: before this line could we xr.open_zarr each of the primary and replica stores (and then discard the result) to confirm they really exist?

aldenks (approved):
Approved after changes!

aldenks (approved):

PR #280, opened 2025-10-02
Control store writability, defaulting to read only

mosegontar:


aldenks:
I wondering if this may be too much magic. What about an implementation where the StoreFactory.primary_store and replica_stores methods take a writeable=False kwarg? I think this is all of the cases, am i missing anything?
- directly calling those within dynamical dataset methods: always obvious what to pass for writable
- template_utils.write_metadata takes the StoreFactory and calls primary_store and replica_stores. i can't see any case where it shouldn't pass writeable=True when getting the stores
- the progress tracker - this takes the store factory but doesn't call primary nor replica_stores

mosegontar:
> I wondering if this may be too much magic. What about an implementation where the StoreFactory.primary_store and replica_stores methods take a writeable=False kwarg? I think this is all of the cases, am i missing anything?
> 
> * directly calling those within dynamical dataset methods: always obvious what to pass for writable
> * template_utils.write_metadata takes the StoreFactory and calls primary_store and replica_stores. i can't see any case where it shouldn't pass writeable=True when getting the stores
> * the progress tracker - this takes the store factory but doesn't call primary nor replica_stores

yeah i think that definitely works! a similar implementation crossed my mind, i think it has the minor downside of ensuring the flag is configured correctly in the downstream places, but it's not many places and it's more explicit. I can make that change!

aldenks:
awesome, thank you! if it's default to writable=False i think we're unlikely to in inadvertently set it incorrectly (default is safe, if we need to write and we don't set it we'll know :) )

src/reformatters/common/storage.py
```diff
@@ -166,31 +166,35 @@ def _get_store_path(
...
 
 def _get_zarr3_store(
-    store_path: str, storage_config: StorageConfig
+    store_path: str, storage_config: StorageConfig, writable: bool
 ) -> zarr.abc.store.Store:
     if Config.is_prod:
         return zarr.storage.FsspecStore.from_url(
             store_path, storage_options=storage_config.load_storage_options()
-        )
+        ).with_read_only(not writable)
```
aldenks, line 189: oh nice! TIL

src/reformatters/common/storage.py
```diff
@@ -216,6 +220,13 @@ def _get_icechunk_store(
...
 
+    if not writable:
+        log.info(f"Opening icechunk store {store_path} in readonly mode")
+        repo = icechunk.Repository.open(storage)
+        session = repo.readonly_session("main")
+        return session.store
+
+    log.info(f"Opening icechunk store {store_path} in writable mode")
     repo = icechunk.Repository.open_or_create(storage)
     session = repo.writable_session("main")
```
aldenks: will you organize into an if, else vs an early return? a little clearer imo and also feels less likely to introduce future bugs

src/reformatters/common/storage.py
```diff
@@ -86,24 +86,24 @@ def k8s_secret_names(self) -> list[str]:
...
+    def replica_stores(self, writable: bool = False) -> list[zarr.abc.store.Store]:
         # Disable replica stores in dev environment
         if Config.is_dev:
             return []
 
         stores = []
         for config in self.replica_storage_configs:
             store_path = _get_store_path(self.dataset_id, self.version, config)
-            store = _get_store(store_path, config)
+            store = _get_store(store_path, config, writable=writable)
```
aldenks: can pass writable to _get_store positionally, don't need the dupe can pass writable to _get_store positionally, don't need the dupe kwarg name=value

src/reformatters/common/dynamical_dataset.py
```diff
@@ -27,7 +27,11 @@
 from reformatters.common.logging import get_logger
 from reformatters.common.pydantic import FrozenBaseModel
 from reformatters.common.region_job import RegionJob, SourceFileCoord
-from reformatters.common.storage import StorageConfig, StoreFactory, get_local_tmp_store
+from reformatters.common.storage import (
+    StorageConfig,
+    StoreFactory,
+    get_local_tmp_store,
```
aldenks: would go back to one line without trailing comma

(i'd like to make it do this automatically..)

aldenks (approved):
Love it! small comments then approved after those

PR #285, opened 2025-10-07
Common secrets

aldenks:
Generalize the secret mounting, decoding, k8s-api-fetching-when-running-locally pattern we use for storage_options.json to be general k8s utils for us so we can use the same pattern to load secrets for other things (e.g. credentials to access some data).

Note: this change requires we edit our existing secrets before merging to have their key be `contents` not `storage_options.json`.  (More general name, and doesn't have dots in it so easier to use with jsonpath=...)

### DIFF
--- a/src/reformatters/common/kubernetes.py
+++ b/src/reformatters/common/kubernetes.py
@@ -255,10 +255,11 @@ class ValidationCronJob(CronJob):
 
 
 def load_secret(secret_name: str) -> dict[str, str | int | float | bool | None]:
-    """Load a secret from k8s, either from mounted file or directly from k8s API.
+    """
+    Load a secret from kubernetes, either from mounted file or directly from k8s API.
 
     Returns empty dict in non-prod environments.
-    In prod, loads from mounted secret file, or falls back to k8s API if running locally.
+    When env is prod, loads from mounted secret file, or falls back to k8s API if running locally.
     """
     if not Config.is_prod:
         return {}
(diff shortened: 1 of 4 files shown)


src/reformatters/common/kubernetes.py
```diff
@@ -1,12 +1,21 @@
...
+from pathlib import Path
 from typing import Annotated, Any
 
 import pydantic
 from kubernetes import client, config  # type: ignore[import-untyped]
 
+from reformatters.common.config import Config
+
+_SECRET_MOUNT_PATH = "/secrets"  # noqa: S105
+_SECRET_CONTENTS_KEY = "contents"  # noqa: S105
```
aldenks, line 17: Note: this change requires we edit our existing secrets before merging

src/reformatters/common/storage.py
```diff
@@ -48,34 +43,10 @@ class StorageConfig(FrozenBaseModel):
 
     def load_storage_options(self) -> dict[str, Any]:
         """Load the storage options from the Kubernetes secret."""
-        if not Config.is_prod or self.k8s_secret_name == _NO_SECRET_NAME:
+        if self.k8s_secret_name == _NO_SECRET_NAME:
             return {}
```
aldenks, line 47: Opted to leave this in here because it really is there because our pydantic StorageConfig.k8s_secret_name is non-nullable and we want to be able to test easily.  For the new, general function we're able to test it fine with monkeypatching

mosegontar (approved):

PR #287, opened 2025-10-08
Contrib NDVI: Bump NaN threshold for validator to 95%

mosegontar:
From a previous spot check of recent data, we saw some days had a nan percentage as low as ~93.5. Recently we've had a number of days where the percentage was a bit higher than 94%. Since this seems to be a somewhat regular occurrence, and there is also little we can do to mitigate it, we are raising the threshold for validation failure. 95% is selected since it would mean that close to 25% of the normally non-nan data happens to be NaN.

```
In [47]: ds.isel(time=-1).ndvi_usable.shape
Out[47]: (3600, 7200)

In [48]: 3600*7200
Out[48]: 25920000

In [49]: total_values = 3600*7200

In [51]: (total_values - (total_values * 0.95)) / (total_values - (total_values * 0.935))
Out[51]: 0.7692307692307693
```


aldenks (approved):
why not make it 96% so this doesn't false alarm on us as much? 

approved after test fixes

mosegontar:
> why not make it 96% so this doesn't false alarm on us as much?
> 
> approved after test fixes

Sure :+1: that ends up being quite a bit of missing "expected" data (~40%), but there's not much we can do so seems reasonable.

PR #283, opened 2025-10-03
ECMWF IFS ENS 15 day forecast: introduce template config

laurengulland:
**This is the first of several PRs introducing an ECMWF IFS ENS 15 day forecast dataset.** 

## Notes about this dataset
Dataset information:
* Begins Feb 1 2024, and will be updating operationally
* 15 day forecast, with a 3-hourly resolution from 0-6 days, and 6-hourly resolution from 6-15 days.
* 0.25 degree spatial resolution
* 51 ensemble members (1 control + 50 perturbed)

Source data location: `s3://ecmwf-forecasts/20250101/00z/ifs/0p25/enfo/`
* This source contains forecasts from four daily runs (0Z, 06Z, 12Z, and 18Z), though this PR creates the template config for 0z forecasts only. (Note also: while the 0z and 12z forecasts go a full 15 days into the future, the 06z and 18z forecast runs truncate at 6 days.)
* This archive begins in January 2023, but forecasts up until Feb 1 2024 are only in 0.4 degree resolution. Forecasts are available with 0.25 degree resolution starting Feb 1 2024 through present.

## Changes included in this PR:
* creates `ecmwf/.../template_config.py` and associated `template_config_tests.py`. The template config is almost entirely complete, save for the internal data var attributes, which are to be completed in tandem with developing the dataset processing code in a follow up PR.
* adds the autogenerated template, generated via `uv run main ecmwf-ifs-ens-forecast-15-day-0-25-degree update-template`
* creates empty `dynamical_dataset`, `regionjob`, & associated test files for next PRs to build on

## Work left for follow up PRs:
* fill out `region_job.py` and `region_job_test.py`
* fill out `dynamical_dataset.py` and `dynamical_dataset_test.py`

PR #289, opened 2025-10-15
ECMWF IFS ENS 15 day forecast: introduce region job

laurengulland:
**Follow up to #283; this is the second of several PRs introducing an ECMWF IFS ENS 15 day forecast dataset.**

PR completed with support from @mosegontar!

## Changes included in this PR:
* **Creates `ecmwf/.../region_job.py` and associated `region_job_tests.py`.** 
* Fleshes out `EcmwfInternalAttrs` on the template config, adding metadata that are useful for processing the index & grib files. Also moves it and `EcmwfDataVar` to a generalized `ecmwf_config_models.py` file for use with other ECMWF datasets.
* Adds some index parsing & byte range window helper functions for ECMWF to a common `ecmwf/ecmwf_grib_index.py` file. These may not be fully generalized to all ECMWF datasets and may be expanded/updated in the future. (A test in `region_job_tests.py` will ensure any future changes to generalize don't break things for this dataset inadvertently.)
* Updates the **start date for this dataset to two days later** -- the first two days of the 0.25 degree resolution period of the dataset (starting at 2024-02-01) only have longitudes -180 through 179.5 (not 179.75 as expected), so we begin this dataset on 2024-02-03 now, when we have full files.
* Updates the template zarr with the above changes.

## Work left for follow up PRs:
* fill out dynamical_dataset.py and dynamical_dataset_test.py


src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
```diff
@@ -273,22 +244,19 @@ def operational_update_jobs(
...
-        #     append_dim=append_dim,
-        #     all_data_vars=all_data_vars,
-        #     reformat_job_name=reformat_job_name,
-        #     filter_start=append_dim_start,
-        # )
-        # return jobs, template_ds
-
-        raise NotImplementedError(
-            "Subclasses implement operational_update_jobs() with dataset-specific logic"
+        existing_ds = xr.open_zarr(primary_store, decode_timedelta=True, chunks=None)
```
laurengulland: do I actually need the decode_timedelta=True & chunks=None here? stole this from HRRR but maybe unnecessary.

src/reformatters/ecmwf/ecmwf_grib_index.py
laurengulland, line 1: very much welcome thoughts on this file placement, naming, or whether it's worth creating (I'm not sure how generalizable this index file structure is or is not to other ECMWF datasets!)

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job_test.py
```diff
@@ -1,36 +1,165 @@
...
+{"domain": "g", "date": "20240201", "time": "0000", "expver": "0001", "class": "od", "type": "cf", "stream": "enfo", "step": "0", "levtype": "pl", "levelist": "850", "param": "2t", "_offset": 665525, "_length": 100000}
+{"domain": "g", "date": "20240201", "time": "0000", "expver": "0001", "class": "od", "type": "cf", "stream": "enfo", "step": "0", "levtype": "sfc", "param": "10u", "_offset": 765525, "_length": 888917}
+{"domain": "g", "date": "20240201", "time": "0000", "expver": "0001", "class": "od", "type": "pf", "stream": "enfo", "step": "0", "levtype": "sfc", "number": "1", "param": "2t", "_offset": 1554442, "_length": 664922}
+{"domain": "g", "date": "20240201", "time": "0000", "expver": "0001", "class": "od", "type": "pf", "stream": "enfo", "step": "0", "levtype": "pl", "levelist": "850", "number": "1", "param": "2t", "_offset": 2219364, "_length": 664716}
+{"domain": "g", "date": "20240201", "time": "0000", "expver": "0001", "class": "od", "type": "pf", "stream": "enfo", "step": "0", "levtype": "sfc", "number": "1", "param": "10u", "_offset": 2884080, "_length": 889546}
```
laurengulland: @mosegontar the only slight tweak I made here was to make `"step"` be `"0"` for these -- it's not currently used in our code anywhere, but I think this makes it more true to life! (there are occasional variables with step=3, but I think they're weird ones?)

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -390,10 +399,19 @@ def data_vars(self) -> Sequence[EcmwfIfsEnsDataVar]:
...
+                    grib_comment="(prodType 0, cat 1, subcat 193) [-]",
+                    grib_description='0[-] SFC="Ground or water surface"',
+                    grib_element="unknown",
+                    grib_index_param="tp",
                     deaccumulate_to_rate=True,
-                    window_reset_frequency=default_window_reset_frequency,
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                    window_reset_frequency=pd.Timedelta(
+                        "360h"
+                    ),  # accumulates over the full dataset, never resetting
```
laurengulland: TODO(lauren) in follow up PR: implement a `window_reset_frequency=None` functionality to mean deaccumulate over the whole dataset, and then come back and set this to that (didn't want to block this PR on that though!)

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -36,11 +36,10 @@ class EcmwfIfsEnsInternalAttrs(BaseInternalAttrs):
...
     )
-    # TODO(lauren): skipping this for now, will come back and add when developing processing for real
-    # grib_band_index: int
-    # grib_element_name: str
-    # grib_description: str
-    # grib_index_level: str
+    grib_index_param: str
+    grib_description: str
+    grib_element: str
+    grib_comment: str
```
laurengulland: one q for reviewer: Alex and I are including extra metadata in here beyond what we strictly use, in case it's helpful for any future applications or development. What we actually use in this PR is solely `grib_index_param` and `grib_description`, but not the other two. If it's preferable to leave the unused attrs off though, let us know.

src/reformatters/ecmwf/parse_data.py
```diff
@@ -0,0 +1,62 @@
...
+    byte_range_starts = []
+    byte_range_ends = []
+    index_file_df = parse_index_file(index_local_path)
+    for data_var in data_vars:
+        start, length = index_file_df.loc[
+            (ensemble_member, data_var.internal_attrs.grib_index_param),
+            ["_offset", "_length"],
+        ].values[0]
+        byte_range_starts.append(start)
+        byte_range_ends.append(start + length + 1)  # +1 to make exclusive end
```
laurengulland: @mosegontar one final thing to gut check with you -- I think this needs the +1? can you make sure I'm not goofing this though? (this is also related to the one failing test, I didn't want to update the digest suffix in the test without confirmation.)

mosegontar: Hm I"m not sure we want `+1` here. Since the second value is `length` a start value of 1 and length of 2, should be start=1 end=3, which we get from just start+length itself. I can try to verify though.

src/reformatters/ecmwf/ecmwf_grib_index.py
mosegontar: The noaa equivalent file is at `src/reformatters/noaa/noaa_grib_index.py`. I think we should name it similarly for consistency, I'll make that change this morning.

mosegontar: (i went ahead and made this change)

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job_test.py
mosegontar: Actually, @laurengulland i think this is intended to match the lead time, 
image in a 3h file

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -92,11 +91,10 @@ def dimension_coordinates(self) -> dict[str, Any]:
...
                 )
             ),
-            "ensemble_member": np.arange(
-                0, 51
-            ),  # single control member (0) + 50 perturbed members (1-50)
+            # single control member (0) + 50 perturbed members (1-50)
+            "ensemble_member": np.arange(0, 51),
             "latitude": np.flip(np.arange(-90, 90.25, 0.25)),
-            "longitude": np.arange(-180, 180, 0.25),
+            "longitude": np.arange(-180, 179.75, 0.25),
```
mosegontar: No longer convinced this is correct for all the gribs. Looking more closely. I'm getting shape errors when backfilling more than 2 days.

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
aldenks: you no longer need decode_timedelta! but it won't hurt. chunks=None we do want, it turns off dask and is generally faster when accessing small amounts of data (or larger amounts too if you manage your own threading)

src/reformatters/ecmwf/ecmwf_grib_index.py
aldenks: I like the name and location

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/templates/latest.zarr/longitude/zarr.json
```diff
@@ -40,7 +40,7 @@
     "units": "degrees_east",
     "statistics_approximate": {
       "min": -180.0,
-      "max": 179.75
+      "max": 179.5
```
aldenks: will you share the result of opening opening a source file with cfgrib and printing out the longitude coordinates to verify this?

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/templates/latest.zarr/longitude/zarr.json
```diff
@@ -1,13 +1,13 @@
 {
   "shape": [
-    1440
+    1439
```
aldenks: this is ... sus. I know you two are looking into it

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: Fine to leave them! leave a comment in this definition that they are currently informational

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: you also could use `pd.Timedelta.max` and be done!

src/reformatters/ecmwf/ecmwf_grib_index.py
```diff
@@ -0,0 +1,63 @@
...
+
+from reformatters.common.config_models import DataVar
+from reformatters.ecmwf.ifs_ens.forecast_15_day_0_25_degree.template_config import (
+    EcmwfIfsEnsInternalAttrs,
+)
+
+
+def get_message_byte_ranges_from_index(
+    index_local_path: PathLike[str],
+    data_vars: Sequence[DataVar[EcmwfIfsEnsInternalAttrs]],
```
aldenks: Might be worth renaming this internal attrs class to EcmwfInternalAttrs and moving its definition into a general ecmwf file. I think all the attrs on it are common for ECMWF dataset i know of

src/reformatters/ecmwf/ecmwf_grib_index.py
```diff
@@ -0,0 +1,63 @@
...
+        Has a MultiIndex of (number, param) representing ensemble member & data variable short name.
+        Useful columns include: type (control/perturbed), _offset (start of byte window within grib file), _length (length of byte window)
+        Additional columns include: levtype (sfc/pl/sol/...), levelist, domain, date, time, step, expver, class, stream
+    """
+    df = pd.read_json(index_local_path, lines=True)
+
+    # Control members by default don't have "number" field. We fill with 0
+    df["number"] = df["number"].fillna(0).astype(int)
+    # Ensure that every row we filled with number=0 was indeed type "cf" (control forecast)
+    assert all(df[df["number"] == 0]["type"] == "cf"), (
```
aldenks: this isn't actually a perf hit here (since it'll be iterating over a single item), but in general prefer a numpy/pandas .all() or np.all vs the built in all because the loop will be in compiled vs interpreted code

src/reformatters/ecmwf/ecmwf_grib_index.py
```diff
@@ -0,0 +1,63 @@
...
+        "Parsed row as control member that didn't have type='cf'"
+    )
+
+    return df.set_index(["number", "param"]).sort_index()
```
aldenks, line 64: do we need the sort_index()? wondering if that's worth it

src/reformatters/ecmwf/ecmwf_grib_index.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    byte_range_starts: list[int] = []
+    byte_range_ends: list[int] = []
+    index_file_df = _parse_index_file(index_local_path)
+    for data_var in data_vars:
+        rows = index_file_df.loc[
+            (ensemble_member, data_var.internal_attrs.grib_index_param),
+            ["_offset", "_length"],
+        ]
+        start, length = rows.values[0]
```
aldenks: lets assert there's only one matching row
```suggestion
        start, length = item(rows.values)
```

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job_test.py
```diff
@@ -1,36 +1,165 @@
...
+        url
+        == "https://ecmwf-forecasts.s3.eu-central-1.amazonaws.com/20240201/00z/0p25/enfo/20240201000000-3h-enfo-ef.grib2"
+    )
+    assert dataset_id == "ecmwf-ifs-ens-forecast-15-day-0-25-degree"
+    assert (
+        local_path_suffix == "-4f434771"
+    )  # result of calling digest on the byte ranges
+    assert byte_ranges == ([0], [665525])
+
+
```
aldenks, line 124: some sort of read_data test here would be ideal

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config_test.py
```diff
@@ -76,12 +76,12 @@ def test_dimension_coordinates_shapes_and_values() -> None:
...
 
-    # longitude: from -180 to +179.75, 0.25° steps
+    # longitude: from -180 to +179.50, 0.25° steps
     lon = dim_coords["longitude"]
     assert isinstance(lon, np.ndarray)
     assert lon[0] == -180.0
-    assert lon[-1] == 179.75
-    assert len(lon) == 1440
+    assert lon[-1] == 179.50
+    assert len(lon) == 1439
```
aldenks: leaving a comment to make sure we change these

aldenks:
Summary: it looks like the first couple days of 0.25 degree data had an inconsistent length in the logitude direction. Assuming that is confirmed we'll start this dataset a couple days later once it becomes consistently 1440 pixels

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/templates/latest.zarr/longitude/zarr.json
mosegontar:
```
In [15]: cfgrib.open_dataset("20240201000000-0h-enfo-ef.grib2", filter_by_keys={"dataType": 
Out[15]: array([-180.  , -179.75, -179.5 , ...,  179.  ,  179.25,  179.5 ])

In [16]: cfgrib.open_dataset("20240203000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "pf"}).longitude.values
Out[16]: array([-180.  , -179.75, -179.5 , ...,  179.25,  179.5 ,  179.75])

In [17]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "pf"}).longitude.values
Out[17]: array([-180.  , -179.75, -179.5 , ...,  179.25,  179.5 ,  179.75])
```

(Switching things to begin on the 3rd, with expectation that max is 179.75 onward.

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job_test.py
laurengulland: d'oh I'm so sorry I had been looking at the `lead_time=pd.Timedelta("0h")` in the above test rather than the one in this test!! thanks :facepalm:

src/reformatters/ecmwf/parse_data.py
laurengulland: (we chatted about this live, totally doesn't need the +1 -- that was indeed a goof!)

src/reformatters/ecmwf/ecmwf_grib_index.py
laurengulland: Alex and I both couldn't remember what this was for, removed it, and then got `PerformanceWarning: indexing past lexsort depth may impact performance` without it! We added it back in and all of those are resolved. (If you'd recommend a different solution though let us know!)

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job_test.py
laurengulland: added!

src/reformatters/ecmwf/ecmwf_grib_index.py
laurengulland: done! thanks!

laurengulland:
@aldenks: ready for a re-review when you get a chance! @mosegontar and I think this is good to go. 

(in the meantime, gonna get started on the dynamical dataset + integration tests!)

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
```diff
@@ -1,223 +1,203 @@
...
-    #     the region to process. In most cases this is exactly self.region, but if additional data outside
-    #     the region is required, for example for correct interpolation or deaccumulation, this method can
-    #     return a modified slice (e.g. `slice(self.region.start - 1, self.region.stop + 1)`).
-    #     """
-    #     return self.region
+    # Limits the number of variables downloaded together.
+    # All variables are scattered throughout the grib file without any organization,
+    # so it's more efficient to do separate windowed downloads & reads for each
+    # variable that we can parallelize.
+    max_vars_per_download_group: ClassVar[int] = 1
```
aldenks, line 97: if follow ups we may want to also set `max_vars_per_backfill_job` so that a job takes about 10 mins and and perhaps try setting `download_parallelism` higher than the default (cpus * 2) because we are downloading individual var+ensemble members which are small requests

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
```diff
@@ -260,35 +231,32 @@ def operational_update_jobs(
...
-        #     reformat_job_name=reformat_job_name,
-        #     filter_start=append_dim_start,
-        # )
-        # return jobs, template_ds
-
-        raise NotImplementedError(
-            "Subclasses implement operational_update_jobs() with dataset-specific logic"
+        existing_ds = xr.open_zarr(primary_store, chunks=None)
+        # start after the last init time in the dataset (so we don't reprocess the last forecast)
+        append_dim_start = existing_ds[append_dim].max() + pd.Timedelta("1s")
```
aldenks: Unless we're sure the last forecast was complete we do actually want to reprocess the last ingested forecast

src/reformatters/ecmwf/ecmwf_grib_index.py
aldenks: ah if its not sorted then pandas index lookups on .loc can't to do nice O(log(n)) indexing and have to scan. the way we call this it actually only calls .loc once so i don't know if the time to sort vs the loc lookup improvement is worth it but either way this is a trivial amount of work so lets keep it and keep the warning quiet. the way to do it without the warning and without sorting would be to make a boolean mask array `df[(df["ensemble_member"] == ensemble_member) & (df["param"] == var.param)]`) but who knows what's faster unless we test and its not worth your time to test :)

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config_test.py
```diff
@@ -76,7 +76,7 @@ def test_dimension_coordinates_shapes_and_values() -> None:
     assert lat[-1] == -90.0
     assert len(lat) == 721
 
-    # longitude: from -180 to +179.75, 0.25° steps
+    # longitude: from -180 to +179.50, 0.25° steps
```
aldenks: this comment now doesn't match the test below, but i believe the comment is wrong:

0.25 deg spacing, lon[0] == -180, -180 - 0.25 == +179.75

aldenks (approved):
Approved after addressing the last 2 comments below!

tests/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config_test.py
mosegontar: ah good catch thanks, left over from the back and forth on this

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
mosegontar: Hm yeah I am going to revert this to be consistent with our general pattern of reprocessing the last ingested forecast. But pinging @laurengulland explicitly if we know the behavior is ECMWF doesn't update the a grib after its available.

PR #297, opened 2025-10-24
ECMWF IFS ENS 15 day forecast: add additional variables

laurengulland:
Builds on #283 and #289. 

Adds additional variables to the ECMWF IFS dataset, including:
* Downward long-wave radiation flux
* Downward short-wave radiation flux
* Precipitation type (categorical variable)
* Pressure at mean sea level
* Pressure at surface
* 100m u & v components of wind

This PR also includes a fix for the `precipitation_surface` variable, which handles converting from `m` to `mm` by including a `scaling_factor` on the variable.

------------

#### Quick reference list of available variables:
<details><summary>Expand to view list of available variables</summary>
<p>
Note that the `param` labels included below correspond to the param in the _index_ file, not the short names in the grib files.

```
Surface variables:
    Convective available potential energy [J/kg] (param: "mucape")
    Deviation of sea level from mean [m] (param: "zos")
    ✅ Downward long-wave radiation flux [W/(m^2)] (param: "strd")
    ✅ Downward short-wave radiation flux [W/(m^2)] (param: "ssrd")
    Dew point temperature [C] (param: "2d")
    Eastward surface water velocity [m s-1] (param: "sve")
    Ice thinkness [m] (param: "sithick")
    Land cover (1=land; 2=sea) [Proportion] (param: "lsm")
    Net long wave radiation flux [W/(m^2)] (param: "str")
    Net short wave radiation flux [W/(m^2)] (param: "ssr")
    Northward surface water velocity [m s-1] (param: "svn")
    ✅ Precipitation type [0=No precipitation; 1=Rain; 2=Thunderstorm; 3=Freezing rain; 4=Mixed/ice; 5=Snow; 6=Wet snow; 7=Mixture of rain and snow; 8=Ice pellets; 9=Graupel; 10=Hail; 11=Drizzle; 12=Freezing drizzle; 13-191=Reserved; 192-254=Reserved for local use; 255=Missing] (param: "ptype")
    ✅ Pressure [Pa] (param: "msl")
    ✅ Pressure [Pa] (param: "sp")
    Reserved [] (param: "ewss") --> guess: eastward turbulent surface stress?
    Reserved [] (param: "nsss") --> guess: northward turbulent surface stress?
    Skin temperature [C] (param: "skt")
    ✅ Temperature [C] (param: "2t")
    Temperature [C] (param: "mx2t3")
    Temperature [C] (param: "mn2t3")
    Total column integrated water vapour [kg/m^2] (param: "tcwv")
    Total column water [kg/m^2] (param: "tcw")
    Total precipitation rate [kg/(m^2s)] (param: "tprate")
    ✅ u-component of wind [m/s] (param: "10u")
    ✅ u-component of wind [m/s] (param: "100u")
    ✅ v-component of wind [m/s] (param: "10v")
    ✅ v-component of wind [m/s] (param: "100v")
    Wind speed (gust) [m/s] (param: "10fg")
    ✅ (prodType 0, cat 1, subcat 193) [-] (param: "tp") --> this one is total precipitation
    (prodType 0, cat 19, subcat 192) [-] (param: "asn") --> ???
    (prodType 2, cat 0, subcat 201) [-] (param: "ro") --> this one is runoff

"Surface" variables:
    Net long wave radiation flux [W/(m^2)] (param: "str") --> listed as a surface var, but actually measured at top of atmosphere.

Pressure level variables (available at a variety of pressure levels, depending on the var):
    Geopotential height [gpm] (param: "gh")
    Relative divergence [1/s] (param: "d")
    Relative humidity [%] (param: "r")
    Relative vorticity [1/s] (param: "vo")
    Specific humidity [kg/kg] (param: "q")
    Temperature [C] (param: "t")
    u-component of wind [m/s] (param: "u")
    v-component of wind [m/s] (param: "v")
    Vertical velocity (pressure) [Pa/s] (param: "w")

Soil Variables:
    Soil Temperature [C] (param: "sot")
    Volumetric soil moisture [m^3/m^3] (param: "vsw")
```

</p>
</details> 

#### Full metadata for all available variables in the source dataset, including those that we have not opted to include yet:

The full set of compiled variable metadata for the control member, including what pressure levels certain variables are available at, all the grib metadata for each variable, etc., can be found in this text file:
[available_ecmwf_variables.txt](https://github.com/user-attachments/files/23110634/available_ecmwf_variables.txt)

Hopefully in the future, if we want to add more variables to this dataset, this provides a much easier jumping off point for what's available, and also includes the information we'd need to add to the InternalAttrs to start ingesting those new vars.

<details><summary>How this list was generated</summary>
<p>

I used the grib and associated index file for  2025-03-01's T00 forecast, lead time 0h.

* For each entry in the index file that corresponds to one ensemble member (the control), I extracted the param short names, the level information for the variable, and the grib byte windows it corresponded to.
* For each set of byte windows, I read just that direct window of the grib to grab the band information, short names, comments, descriptions, and more.
* I combined the index + grib metadata for each variable into one entry, since you need both halves to understand what any given variable really is. (There are a handful of variables in the grib that have nearly zero useful information in the band metadata, except when cross-referenced to short name listed in the index file, which is why I did it this way.)

Note: for most variables, the short param name in the index & the short variable name in the grib itself do __not__ match.
</p>
</details>


src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -392,4 +446,92 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+                name="precipitable_water_atmosphere",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="pwat",
+                    long_name="Precipitable water",
+                    units="kg/(m^2)",
+                    step_type="instant",
+                ),
+                internal_attrs=EcmwfInternalAttrs(
+                    grib_comment="Total column water [kg/m^2]",
```
laurengulland: Confirming: this is a different name, but is the equivalent of precipitable water in the whole atmosphere, correct?

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -392,4 +446,92 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+                    grib_index_param="tcw",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                ),
+            ),
+            EcmwfDataVar(
+                name="categorical_precipitation_type_surface",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="ptype",
+                    long_name="Categorical precipitation type at surface",
```
laurengulland, line 459: we don't have the same kinds of categorical snow/rain/etc vars that gefs has, but we do have this. I made up var names for this, but not sure if there's a precedent for this in other datasets that I missed, and/or if these seem reasonable!

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -392,4 +446,92 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+                    units="[0=No precipitation; 1=Rain; 2=Thunderstorm; 3=Freezing rain; 4=Mixed/ice; 5=Snow; 6=Wet snow; 7=Mixture of rain and snow; 8=Ice pellets; 9=Graupel; 10=Hail; 11=Drizzle; 12=Freezing drizzle; 13-191=Reserved; 192-254=Reserved for local use; 255=Missing",
+                    step_type="avg",
```
laurengulland: gefs had avg rather than instantaneous steps for these categorical vars, but not sure if that's also expected to be true here -- expecting to verify in validation?

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -392,4 +446,92 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+            ),
+            EcmwfDataVar(
+                name="downward_long_wave_radiation_flux_surface",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="sdlwrf",
+                    long_name="Surface downward long-wave radiation flux",
+                    units="W/(m^2)",
+                    comment="Average value in the last 6 hour period (00, 06, 12, 18 UTC) or 3 hour period (03, 09, 15, 21 UTC).",
+                    step_type="avg",
```
laurengulland: directly pulled this from gefs, but again, waiting for validation to see if it's at all reasonable or not 

(same for short wave flux below)

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
laurengulland, line 1: Regarding what other variables we may want to include:

Here's a summary of the available variables, which I assembled by combining the index & grib file information. Green checks means they should be represented already, Red question marks are the biggest ones I'm wondering if are worth also including.

```
Surface variables:
    Convective available potential energy [J/kg] (param: "mucape")
    Deviation of sea level from mean [m] (param: "zos")
    ✅ Downward long-wave radiation flux [W/(m^2)] (param: "strd")
    ✅ Downward short-wave radiation flux [W/(m^2)] (param: "ssrd")
    Dew point temperature [C] (param: "2d")
    Eastward surface water velocity [m s-1] (param: "sve")
    Ice thinkness [m] (param: "sithick")
    Land cover (1=land; 2=sea) [Proportion] (param: "lsm")
    Net long wave radiation flux [W/(m^2)] (param: "str")
    Net short wave radiation flux [W/(m^2)] (param: "ssr")
    Northward surface water velocity [m s-1] (param: "svn")
    ✅ Precipitation type [0=No precipitation; 1=Rain; 2=Thunderstorm; 3=Freezing rain; 4=Mixed/ice; 5=Snow; 6=Wet snow; 7=Mixture of rain and snow; 8=Ice pellets; 9=Graupel; 10=Hail; 11=Drizzle; 12=Freezing drizzle; 13-191=Reserved; 192-254=Reserved for local use; 255=Missing] (param: "ptype")
    ✅ Pressure [Pa] (param: "msl")
    ✅ Pressure [Pa] (param: "sp")
    Reserved [] (param: "ewss") --> guess: eastward turbulent surface stress?
    Reserved [] (param: "nsss") --> guess: northward turbulent surface stress?
    Skin temperature [C] (param: "skt")
    ✅ Temperature [C] (param: "2t")
    Temperature [C] (param: "mx2t3")
    Temperature [C] (param: "mn2t3")
    Total column integrated water vapour [kg/m^2] (param: "tcwv")
    ✅ Total column water [kg/m^2] (param: "tcw")
    Total precipitation rate [kg/(m^2s)] (param: "tprate")
    ✅ u-component of wind [m/s] (param: "10u")
    ✅ u-component of wind [m/s] (param: "100u")
    ✅ v-component of wind [m/s] (param: "10v")
    ✅ v-component of wind [m/s] (param: "100v")
    Wind speed (gust) [m/s] (param: "10fg")
    ✅ (prodType 0, cat 1, subcat 193) [-] (param: "tp") --> this one is total precipitation
    (prodType 0, cat 19, subcat 192) [-] (param: "asn") --> ???
    ❓(prodType 2, cat 0, subcat 201) [-] (param: "ro") --> this one is runoff
"Surface" variables:
    Net long wave radiation flux [W/(m^2)] (param: "str") --> at top of atmosphere.
Pressure level variables:
    ❓Geopotential height [gpm] (param: "gh")
    Relative divergence [1/s] (param: "d")
    ❓Relative humidity [%] (param: "r")
    Relative vorticity [1/s] (param: "vo")
    Specific humidity [kg/kg] (param: "q")
    Temperature [C] (param: "t")
    u-component of wind [m/s] (param: "u")
    v-component of wind [m/s] (param: "v")
    Vertical velocity (pressure) [Pa/s] (param: "w")
Soil Variables:
    Soil Temperature [C] (param: "sot")
    ❓Volumetric soil moisture [m^3/m^3] (param: "vsw")
```

If you'd like the full compilation of grib + index info (e.g. what pressure levels those pressure vars are available at, etc.), check out this file. 
[available_ecmwf_variables.txt](https://github.com/user-attachments/files/23110634/available_ecmwf_variables.txt)


My main questions:
* GEFS includes relative humidity & geopotential height, but at the surface/at 2m, which we don't have. Would choosing the lowest pressure level be the most reasonable equivalent? Any guidance here?
* We've got soil moisture and runoff if we want them?? thoughts?
* Any other vars that stand out to you @aldenks?

_note for handoff if someone other than me picks this up in the morning: if we want to include any pressure-level variables, we'd need to add another internalattr depicting `levelist` + a couple lines to the index parsing to get the byte ranges that includes what level those vars are at._

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
mosegontar: I defer to @aldenks re how to proceed, but I think we could reproduce the categorical variables by parsing out this variable with a mask using the units.

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: this one we need to look up in their docs and see. in the gdalinfo you can also sometimes figure it out based on timestep info in the var metadata

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: yes this looks good

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: lets skip all the red question marks for now. it's not hard to add more later

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -392,4 +446,92 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+                ),
+            ),
+            EcmwfDataVar(
+                name="downward_long_wave_radiation_flux_surface",
+                encoding=encoding_float32_default,
+                attrs=DataVarAttrs(
+                    short_name="sdlwrf",
+                    long_name="Surface downward long-wave radiation flux",
+                    units="W/(m^2)",
+                    comment="Average value in the last 6 hour period (00, 06, 12, 18 UTC) or 3 hour period (03, 09, 15, 21 UTC).",
```
aldenks: double check this, i think this setup is odd and probably unique to gefs

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: it's both reasonable to be inst or avg, we need to look it up

aldenks (commented):
looks reasonable so far!

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
aldenks: Let's leave as-is. It's quite simple to use and we wouldn't want to create as many new variables as there are categories here

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
mosegontar:
```
In [10]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "cf", "typeOfLevel": "surface"}, decode_timedelta=True).ptype.GRIB_stepType
Out[10]: 'instant'
```

Updating

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
mosegontar: Hm seems like for both downward long/short wave, what we have here is wrong:

```
In [25]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "cf", "typeOfLevel": "surface"}, decode_timedelta=True).strd.units
Out[25]: 'J m**-2'

In [26]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "cf", "typeOfLevel": "surface"}, decode_timedelta=True).strd.GRIB_stepType
Out[26]: 'accum'

In [27]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "cf", "typeOfLevel": "surface"}, decode_timedelta=True).ssrd.GRIB_stepType
Out[27]: 'accum'

In [28]: cfgrib.open_dataset("20250202000000-6h-enfo-ef.grib2", filter_by_keys={"dataType": "cf", "typeOfLevel": "surface"}, decode_timedelta=True).ssrd.GRIB_units
Out[28]: 'J m**-2'
```

Here are docs:
- https://codes.ecmwf.int/grib/param-db/169
- https://codes.ecmwf.int/grib/param-db/175

I think this means we'll need to deaccumulate as well, which should get us the proper units (W/m^2).

There docs say:
> This parameter is [accumulated over a particular time period which depends on the data extracted](https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation#ERA5:datadocumentation-Meanrates/fluxesandaccumulations).

And link to an ERA5 doc, which is a bit confusing though I assume it implies variables are treated the same.

This says:

> 
> Minimum/maximum since the previous post processing
> The short forecasts of ERA5 contain some surface and single level parameters that are the minimum or maximum value since the previous post processing (archiving), see Table 5 below. So, for:

> reanalysis: the minimum or maximum values are in the hour (the processing period) ending at the validity date/time
> ensemble: the minimum or maximum values are in the 3 hours (the processing period) ending at the validity date/time

So we could do 3 hours for the deaccumulation window, but the fact that our lead time switches to 6 hours in the latter half gives me pause.

aldenks: ah that sounds like what they do with precip (accumulate over the whole forecast period), but the best way to confirm is plot a time series of it along the lead time dimension for a single init at a point.

if that is the case we probably do want to deaccumulate. that's an opinionated choice, but on the scale of them it seems pretty minor to me, the # of cases where the value you actually want to use is accumulated since start of period seems a lot less frequent than where you want an instantaneous or average over a short window

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
```diff
@@ -109,19 +124,23 @@ def generate_source_file_coords(
...
-        return [
-            EcmwfIfsEnsForecast15Day025DegreeSourceFileCoord(
+        coords = []
+        for init_time, lead_time, ensemble_member in itertools.product(
+            processing_region_ds["init_time"].values,
+            processing_region_ds["lead_time"].values,
+            processing_region_ds["ensemble_member"].values,
+        ):
+            if not all_variables_available(data_var_group, init_time):
+                continue
```
laurengulland, line 140: Just to check my understanding: we're assuming that we shouldn't add any SourceFileCoords if some/any of the vars are unavailable, because we assume there's one var per data var group, yes?

If so, at the very least I'd want to add a comment here in case that assumption ever changes. Or maybe better (but unnecessary for now), make some more generalizable logic like "if not all_vars_available:  ->  iterate through and find available vars -> make subset_data_var_group from the vars that are available for that init time -> coords.append(SourceFileCoord(subset_data_var_group))". 

Again, not that that's actually necessary for now, but perhaps at least a comment about "this is presuming one data var per group", given that `all_variables_available` seems generalizeable?

mosegontar: It's not because there is a data var per group (we have 11 in one group and 1 in the other in our case), but we should add  an assert before the `continue` that all the group's `date_available` values are  the same, so we don't skip a variable because a variable was accidentally included in the group but has a different data available value and it accidentally gets skipped. I'll make this change.

src/reformatters/common/deaccumulation.py
```diff
@@ -5,13 +5,21 @@
...
+DEFAULT_INVALID_BELOW_THRESHOLD_RATE = -2e-5
+
 
 def deaccumulate_to_rates_inplace(
     data_array: xr.DataArray,
     *,
     dim: str,
     reset_frequency: pd.Timedelta,
     skip_step: Array1D[np.bool] | None = None,
+    invalid_below_threshold_rate: float | None = None,
```
aldenks: Why not set the default here vs below in the function body?

src/reformatters/common/deaccumulation.py
```diff
@@ -5,13 +5,21 @@
 
 from reformatters.common.types import Array1D, ArrayFloat32
 
+# OK to add units to this list if you believe they are reasonable output units to deaccumulate to.
+# We typically expect these to be per-second rates.
+VALID_OUTPUT_UNITS_FOR_DEACCUMULATION = ["mm/s", "m/s", "W/(m^2)"]
+
+
+DEFAULT_INVALID_BELOW_THRESHOLD_RATE = -2e-5
```
aldenks: Add comment like a "a small value that is reasonable for precipitation rates in mm/s"

src/reformatters/common/deaccumulation.py
```diff
@@ -57,7 +65,12 @@ def deaccumulate_to_rates_inplace(
         np.prod(data_array.shape[time_dim_index + 1 :] or 1),
     )
 
-    _deaccumulate_to_rates_numba(values, seconds, reset_after, skip_step)
+    if invalid_below_threshold_rate is None:
+        invalid_below_threshold_rate = DEFAULT_INVALID_BELOW_THRESHOLD_RATE
+
+    _deaccumulate_to_rates_numba(
+        values, seconds, reset_after, skip_step, invalid_below_threshold_rate
```
aldenks, line 70: lets remove the default value on the internal _deaccumulate_to_rates_numba and make it required

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
```diff
@@ -109,19 +124,29 @@ def generate_source_file_coords(
...
+            processing_region_ds["ensemble_member"].values,
+        ):
+            if not all_variables_available(data_var_group, init_time):
+                dates_available = {
+                    v.internal_attrs.date_available for v in data_var_group
+                }
+                assert len(dates_available) == 1, (
+                    f"Expected all variables in the group to have the same date_available, found {dates_available}"
+                )
+                continue
```
aldenks, line 140: this check and the all_variables_available util function feel a little more complex than necessary. what about doing the list comprehension as it was and then something like

```assert all(len(set(d for v in coord.var_group if (d := v.internal_attrs.date_available) is not None)) <= 1 for coord in coords)```

File: src/reformatters/common/deaccumulation.py (hunk 7 above)
Comment (aldenks) [reply src/reformatters/common/deaccumulation.py]: Ah, i see because the value can be set to None in the data var internal attrs. What about explicitly setting it in the template config for precip and then we can be more standard here?

Now that there isnt' a single value to use here I'm wanting to have us be explicit with it on every data var but don't want to increase this scope.

File: src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py [hunk 11]
```diff
@@ -388,8 +441,91 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
                     grib_element="unknown",
                     grib_index_param="tp",
                     deaccumulate_to_rate=True,
+                    scale_factor=1000,  # The raw data is in meters so we will need to scale to mm
                     keep_mantissa_bits=default_keep_mantissa_bits,
                     window_reset_frequency=pd.Timedelta.max,  # accumulate over the full dataset, never resetting
```
aldenks: "... full lead time dimension"

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
```diff
@@ -388,8 +441,91 @@ def data_vars(self) -> Sequence[EcmwfDataVar]:
...
+                    grib_description='0[-] SFC="Ground or water surface"',
+                    grib_element="DSWRF",
+                    grib_index_param="ssrd",
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                    deaccumulate_to_rate=True,
+                    window_reset_frequency=pd.Timedelta.max,
+                    # Downward short and long wave radiation flux surface values need to be deaccumulated,
+                    # but their raw values are much larger than precip, so we need to use a different threshold
+                    # checking overly-negative values.
+                    deaccumulation_invalid_below_threshold_rate=-8e3,
```
aldenks: What about putting -8e3 in a constant in deaccumulation.py (so we have PRECIPITATION_RATE_INVALID_BELOW_THRESHOLD, the default, and RADIATION_INVALID_BELOW_THRESHOLD.) Then the code is self explanatory without comments and we make it easier to use the right value in new cases

src/reformatters/ecmwf/ecmwf_utils.py
```diff
@@ -0,0 +1,15 @@
+from collections.abc import Sequence
+
+from reformatters.common.types import Timestamp
+from reformatters.ecmwf.ecmwf_config_models import EcmwfDataVar
+
+
+def all_variables_available(
```
aldenks, line 7: as above, i think we can remove this file

aldenks (approved):
Approved after comments!

src/reformatters/common/deaccumulation.py
aldenks: actually per later comment, lets rename constant to PRECIPITATION_INVALID_BELOW_THRESHOLD_RATE

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/region_job.py
aldenks: ahhh i see, we need to skip the coord. NVM

src/reformatters/ecmwf/ifs_ens/forecast_15_day_0_25_degree/template_config.py
laurengulland: added this information to the description so it's easier to start from later!

src/reformatters/ecmwf/ecmwf_utils.py
laurengulland: (resolving without addressing -- see above comment https://github.com/dynamical-org/reformatters/pull/297/files#r2474154407 )

PR #300, opened 2025-10-31
Reduce char length of ecmwf operational update job name

laurengulland:
The length of this job name is too long! We were getting the following errors during deploys.

`The CronJob "ecmwf-ifs-ens-forecast-15-day-0-25-degree-operational-update" is invalid: metadata.name: Invalid value: "ecmwf-ifs-ens-forecast-15-day-0-25-degree-operational-update": must be no more than 52 characters`

The new name (`ecmwf-ifs-ens-forecast-15-day-0-25-degree-update`) should slide right in under the limit.

We may want to do this for all operational update jobs in the future, but this unblocks deploys without messing with our operational resources for now.


aldenks (approved):
Approved after updating the test!

PR #301, opened 2025-11-03
Ensure that kubernetes_job.job_name computes the same name every time it is called

mosegontar:
As a `property`, `job_name` has been returning a different name each time it is invoked, because we randomly generate a suffix for each name. For each instantiated k8s job object, the name should be consistent across invocations.

The bug's impact thus far as has been minimal: we call `job_name` in only two places  currently: setting the `metadata` name in the k8s yaml, and when logging the job name. The current problematic behavior is that we log the incorrect job name when we do:

```
        log.info(f"Submitted kubernetes job {kubernetes_job.job_name}")
```

This fix caches the result, so the name is consistent.

### DIFF
--- a/tests/common/test_kubernetes.py
+++ b/tests/common/test_kubernetes.py
@@ -178,6 +178,7 @@ def test_kubernetes_job_name() -> None:
 
     k8s_obj: dict[str, Any] = job.as_kubernetes_object()
     assert job.job_name == k8s_obj["metadata"]["name"]
+    assert job.job_name == job.job_name  # quick explicit check that result is cached
 
 
 def test_as_kubernetes_object_with_custom_values() -> None:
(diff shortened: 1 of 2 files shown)


tests/common/test_kubernetes.py
```diff
@@ -163,6 +163,23 @@ def test_as_kubernetes_object_comprehensive() -> None:
...
+        cpu="500m",
+        memory="1Gi",
+        workers_total=4,
+        parallelism=2,
+        secret_names=["aws-creds", "db-creds"],
+    )
+
+    k8s_obj: dict[str, Any] = job.as_kubernetes_object()
+    assert job.job_name == k8s_obj["metadata"]["name"]
+
```
aldenks, line 182: quick test they random part is actually cached? `assert job.job_name == job.job_name`

aldenks (approved):

PR #302, opened 2025-11-04
Allow up to 5 days lag for NASA SMAP

aldenks:


aldenks:
this was not actually the validation error. closing

PR #312, opened 2025-11-10
Update ECMWF IFS ENS dataset name and description

aldenks:


laurengulland:
Thank you! I like this change :)

PR #315, opened 2025-11-14
Factor out common HRRR RegionJob and TemplateConfig

aldenks:
Lay the groundwork for HRRR analysis.

No new functionality introduced in this PR. 

- template for hrrr 48 hour has no changes as expected
- hrrr 48 hour dataset integration tests continue to pass with no changes as expected

### DIFF
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
@@ -40,6 +40,7 @@ class NoaaHrrrForecast48HourTemplateConfig(NoaaHrrrTemplateConfig):
     append_dim_start: Timestamp = pd.Timestamp("2018-07-13T12:00")  # start of HRRR v3
     append_dim_frequency: Timedelta = pd.Timedelta("6h")
 
+    #  here's an example dataset attributes AI
     @computed_field  # type: ignore[prop-decorator]
     @property
     def dataset_attributes(self) -> DatasetAttributes:
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
@@ -40,7 +40,6 @@ class NoaaHrrrForecast48HourTemplateConfig(NoaaHrrrTemplateConfig):
     append_dim_start: Timestamp = pd.Timestamp("2018-07-13T12:00")  # start of HRRR v3
     append_dim_frequency: Timedelta = pd.Timedelta("6h")
 
-    #  here's an example dataset attributes AI
     @computed_field  # type: ignore[prop-decorator]
     @property
     def dataset_attributes(self) -> DatasetAttributes:
--- a/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
+++ b/src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
@@ -22,7 +22,7 @@ from reformatters.common.zarr import (
 from reformatters.noaa.hrrr.hrrr_config_models import (
     NoaaHrrrDataVar,
 )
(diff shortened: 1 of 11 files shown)


aldenks:
I weighed having hrrr analysis classes just subclass their hrrr forecast counterparts.. that spreads logic over fewer places, this approach is probably more obvious to new readers. I could go either way

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -345,399 +229,4 @@ def data_vars(self) -> Sequence[NoaaHrrrDataVar]:
...
-            (
-                -2699020.142521929,
-                -1588806.152556665,
-                2697979.857478071,
-                1588193.847443335,
-            ),
-            (3000.0, -3000.0),
-            "+proj=lcc +lat_0=38.5 +lon_0=-97.5 +lat_1=38.5 +lat_2=38.5 +x_0=0 +y_0=0 +R=6371229 +units=m +no_defs=True",
-        )
+        return self.get_data_vars(encoding_float32_default)
```
mosegontar, line 232: I wonder if `var_chunks`, `var_shards` and `var_encoding` should be defined as properties or methods on the TemplateConfig. That would allow you to avoid needing to know to call this `self.get_data_vars` and pass things through, since the parent `data_vars` could reference `self.var_encoding`

src/reformatters/noaa/hrrr/template_config.py
```diff
@@ -0,0 +1,562 @@
...
+    BLOSC_4BYTE_ZSTD_LEVEL3_SHUFFLE,
+    BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE,
+)
+from reformatters.noaa.hrrr.hrrr_config_models import (
+    NoaaHrrrDataVar,
+    NoaaHrrrInternalAttrs,
+)
+
+
+class NoaaHrrrTemplateConfig(TemplateConfig[NoaaHrrrDataVar]):
```
mosegontar: This is a bit of nit since it's just a naming thing, but I wonder if it would be clearer to name this "base" class `NoaaHrrrTemplateConfigMixin`. I think that probably more accurately captures its use as a utility class that cannot be used standalone (required fields are not fully defined), inheritance still works, etc.

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
```diff
@@ -119,7 +102,10 @@ def coords(self) -> Sequence[Coordinate]:
         dim_coords = self.dimension_coordinates()
         append_dim_coordinate_chunk_size = self.append_dim_coordinate_chunk_size()
 
+        spatial_coords = super().coords
```
mosegontar: If we were going the Mixin route, maybe I'd call `self.spatial_coords` rather than trying to reason about why `super().coords` only has spatial coordinates.

mosegontar:
> I weighed having hrrr analysis classes just subclass their hrrr forecast counterparts.. that spreads logic over fewer places, this approach is probably more obvious to new readers. I could go either way

I think I generally prefer the approach you ended up with as being more intuitive (both inheriting from some shared common HRRR configuration), though maybe if there is some conceptually sensible relationship (where HRRR Analysis is seen as a variation on Forecast) directly inheriting from Forecast might make sense. I do agree with the potential upside (logic remains consolidated) and that one doesn't really have to reason about what the common/shared/base thing(s) represent.

src/reformatters/noaa/hrrr/forecast_48_hour/template_config.py
aldenks: mmm i do like how that has a more clear place to define encodings and avoids this get_data_vars function. the downside is that makes it awkwards to have different encodings per variable (which is a pretty reasonable thing to do, eg. categorical vars, or larger chunking for really compressible variables.) Maybe there's some way around that..

PR #316, opened 2025-11-14
Speed up common RegionJob test and mark it slow + AGENTS.md tweak

aldenks:


mosegontar (approved):
Seems reasonable to me!

Issue #258, opened 2025-09-23
Save ICON-EU GRIB files from DWD's FTP server to cloud object storage


JackKelly:
## Motivation

It seems that people are generally in favour of the idea of us archiving ICON-EU GRIB files on cloud object storage. For example, in [this comment](https://github.com/dynamical-org/reformatters/issues/176#issuecomment-3069883223), @aldenks says:

> I like the idea of archiving the complete gribs (e.g. on Source Coop) and then starting with a more limited surface variable/level set in the Zarr and seeing what people ask to add in. For AI WP model training I'd imagine almost everyone wants the whole domain at each timestep, meaning the native "chunking" of the grib files is actually optimal and we could expose it as a virtualzarr/virtual icechunk dataset and get equivalent read performance.

And, TBH, it's turning out to be quite hard to find a high-quality archive of ICON-EU data. (DWD delete GRIBs from their HTTP server after 24 hours). So archiving the GRIBs would be a useful contribution in its own right. And, before we know it, a year will have elapsed and we'll have a reasonable-sized archive.

And, if I've understood correctly, `RegionJob.operational_update_jobs` kind of expects to have access to an archive of GRIB files that spans from whenever `operational_update_jobs` was _last_ run until _now_. Which could be more than 24 hours if, for example, the ICON-EU `reformatters` code fails on a Friday night, and I don't fix it until Monday. So I'm pausing work on PR #226, until we have an operational service that copies GRIBs and stores them on object storage.

## Implementation

How would you like me to implement the script that grabs GRIBs from DWD's HTTP server and copies them to Source Coop?

I'm imagining a fairly simple script Python that you could run using cron (or similar). The script would:

* Take command line arguments for the source and destination. (e.g. source = `https://opendata.dwd.de/weather/nwp/icon-eu/grib/`, destination = `s3://sourcecoop/dynamical/dwd/icon-eu`)
* Check what's already in the destination
* Copy anything that's in the source that's not already in the destination
* Check the filesizes of the files in the source, and compare with the filesizes in the destination, and re-copy if any filesizes are too small (which indicates a partial transfer).
* Log using Python's standard logging mechanism

I'd probably use [`obstore`](https://github.com/developmentseed/obstore) instead of `fsspec`.

Does that sound like something I should build? Or would you like it implemented differently? (No worries either way!)

aldenks:
[...]

PR #318, opened 2025-11-19
Copy files from FTP to an `ObjectStore`

JackKelly:
This PR implements a simple, concurrent strategy for copying many files from an FTP server to an `ObjectStore`.

The main use-case for this code is copy GRIB files from DWD's FTP server to dynamical's source co-op bucket (see issue #258). To keep this PR small(ish), the code in this PR doesn't know anything about DWD. This PR just implements a general-purpose utility for downloading many FTP files.

A follow-on PR will implement logic to find which files to download.

Note that this PR relies on PR #319 to bump the Python version from 3.12 to 3.13 so we can use [`asyncio.Queue.shutdown`](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.shutdown) (which was only introduced in 3.13). Once PR #319 is merged into `main`, this current PR will only show 3 files as changed.

Still TODO before this PR is ready for review:

- [x] Implement retry logic in `obstore_worker`.
- [x] Implement mock FTP server for testing.
- [x] Save to a temporary directory for testing.
- [x] Fix Ruff failures in main code due to updating to Python 3.13 (see comment below) (implemented in PR #319)
- [x] Convert `print` statements to `log` statements.
- [x] Or, actually, maybe don't return any list of files, but just log failures. And rely on the code that figures out what files to download to find missing files. 
- [x] Describe function params in docstrings.
- [x] Tidy up `log` statements: remove some; change log levels on some.
- [x] Test failed FTP connection.
- [x] Test failed obstore write.
- [x] Test missing FTP file.


JackKelly:
@aldenks would you be comfortable with me adding `pytest-asyncio` to the `dev` dependencies in `pyproject.toml`? It's not a biggie. We can test async functions _without_ `pytest-asyncio`, it's a just a little prettier with `pytest-asyncio`:

### Without `pytest-asyncio`:

```python
@pytest.fixture
def sleep_time() -> float:
    return 0.5

async def _foo(sleep_time: float):
    await asyncio.sleep(sleep_time)

def test_foo(sleep_time: float):
    # Fixtures aren't applied to helper functions, so we have to pass in `sleep_time`:
    asyncio.run(_foo(sleep_time))
```

### With `pytest-asyncio` 

And after [setting](https://pytest-asyncio.readthedocs.io/en/latest/concepts.html#test-discovery-modes) `asyncio_mode = "auto"` in `[tool.pytest]` in `pyproject.toml`:

```python
@pytest.fixture
def sleep_time() -> float:
    return 0.5

async def test_foo(sleep_time: float):
    await asyncio.sleep(sleep_time)
```

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "zarr~=3.1",
     "kubernetes~=33.1",
     "icechunk~=1.1",
+    "aioftp>=0.27.2",
```
aldenks: lets do `~=0.27` and avoid accidentally pulling in a 1.0

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+    src_ftp_path: PurePosixPath
+    dst_obstore_path: str
+    n_retries: int = field(default=0, init=False)
+
+
+@dataclass
+class _ObstoreFile:
+    ftp_file: _FtpFile
+    data: bytes
+    n_retries: int = field(default=0, init=False)
```
aldenks: we use pydantic for most of our data classes, main benefit here is runtime checking that our types are actually respected

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+    # Put _FtpFile objects on the ftp_queue:
+    ftp_queue: Queue[_FtpFile] = Queue()
+    for src, dst in zip(src_ftp_paths, dst_obstore_paths, strict=True):
+        ftp_queue.put_nowait(_FtpFile(src, dst))
+
+    # Set maxsize of Queue to apply back-pressure to the FTP workers.
+    obstore_queue: Queue[_ObstoreFile] = Queue(maxsize=n_obstore_workers * 2)
+
+    # --- Start the Pipeline ---
+    async with asyncio.TaskGroup() as tg:
```
aldenks: slight preference for spelling out task_group local var name vs tg

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+        max_retries: The maximum number of times to try downloading each FTP file before giving up.
+            max_retries must be >= 1.
+
+    Raises:
+        A `TimeoutError`, `OSError`, or `StatusCodeError` if the FTP connection fails after
+        `max_retries` attempts.
+    """
+    worker_id_str: str = f"ftp_worker {worker_id}:"
+    log.info("%s Starting up...", worker_id_str)
+    _sanity_check_value_of_max_retries(max_retries)
```
aldenks: maybe just `assert max_retries >= 0` and save the reader having to go find the function def (even though it is used in two places i think)

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+                    worker_id_str,
+                )
+                await ftp_queue.put(ftp_file)
+            else:
+                log.error("%s ERROR: Giving up on ftp_file", worker_id_str)
+            ftp_queue.task_done()
+        else:
+            log.info("%s Finished downloading %s", worker_id_str, ftp_file.src_ftp_path)
+            await obstore_queue.put(_ObstoreFile(ftp_file=ftp_file, data=data))
+            ftp_queue.task_done()
```
aldenks, line 288: i like how you use the task_done() call in the finally in the other method to ensure we always call task_done() vs forgetting in some branch in a later refactor, maybe do that here too

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+            )
+            if obstore_file.n_retries < max_retries:
+                obstore_file.n_retries += 1
+                log.warning(
+                    "%s Putting obstore_file back on queue to retry later.",
+                    worker_id_str,
+                )
+                await obstore_queue.put(obstore_file)
+            else:
+                log.error(
```
aldenks: Will you make this a log.exception? sentry error tracking will then create an alert in that case, but it still won't block other processing

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+                obstore_file,
+                e,
+            )
+            if obstore_file.n_retries < max_retries:
+                obstore_file.n_retries += 1
+                log.warning(
+                    "%s Putting obstore_file back on queue to retry later.",
+                    worker_id_str,
+                )
+                await obstore_queue.put(obstore_file)
```
aldenks, line 343: lets add a sleep before putting back on queue so we don't exhaust retries quickly

src/reformatters/common/ftp_to_obstore.py
```diff
@@ -0,0 +1,353 @@
...
+            _log_ftp_exception(ftp_file, e, worker_id_str)
+            if ftp_file.n_retries < max_retries:
+                ftp_file.n_retries += 1
+                log.warning(
+                    "%s WARNING: Putting ftp_file back on queue to retry later.",
+                    worker_id_str,
+                )
+                await ftp_queue.put(ftp_file)
+            else:
+                log.error("%s ERROR: Giving up on ftp_file", worker_id_str)
```
aldenks: log.exception here, that will create an issue / notification in our error reporting and attach the raised exception so we have it all together without parsing adjacent log lines

aldenks (approved):
Nice, and thanks for breaking up into manageable PRs! you're welcome to add `pytest-asyncio` to dev deps

My comments are all small, merge to main yourself when you've addressed whichever of them make sense

src/reformatters/common/ftp_to_obstore.py
JackKelly: Good point. I've swapped from `from dataclasses import dataclass` to `from pydantic.dataclasses import dataclass`, which I believe should give us runtime type checking, but also makes it super-easy to switch back to stdlib dataclass if Pydantic turns out to be too slow (which is unlikely... but each NWP init is about 70,000 files... so we'll be creating hundreds of thousands of objects each time this code is run).

71bbe98

src/reformatters/common/ftp_to_obstore.py
JackKelly: Good idea. Done in f846360

src/reformatters/common/ftp_to_obstore.py
JackKelly: Good idea! Done in 305175a

(TBH, the reason I hadn't done this already is because I mistakenly believed that a `finally` block wouldn't be called if an Exception is re-raised from a `try... except` block! I've learnt something new! Thanks!)

src/reformatters/common/ftp_to_obstore.py
JackKelly: Good idea. Done in fb7cf26. Following your suggestion, I also made the same change (replacing `log.error` with `log.exception`) in `_process_ftp_queue`.

PR #326, opened 2025-11-28
Replace `@pydantic_dataclass` with `_StrictBaseModel` in `ftp_to_obstore.py`

JackKelly:
This PR started as a fix for a spelling mistake :slightly_smiling_face:: Replace `pydatic_dataclass` with `pydantic_dataclass` in `ftp_to_obstore.py`.

But we also took the opportunity to change from using the pydantic `@dataclass` decorator to using a `_StrictBaseModel`.

### DIFF
--- a/src/reformatters/common/ftp_to_obstore.py
+++ b/src/reformatters/common/ftp_to_obstore.py
@@ -18,28 +18,31 @@ Each `ftp_worker` keeps an `aioftp.Client` alive until there are no more `_FtpFi
 import asyncio
 from asyncio import Queue
 from collections.abc import Iterable
-from dataclasses import field
 from pathlib import PurePosixPath
 
 import aioftp
+import pydantic
 from obstore.store import ObjectStore
-from pydantic import BaseModel
 
 from reformatters.common.logging import get_logger
 
 log = get_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


aldenks:
Even (very slightly) better, subclass FrozenBaseModel instead of using the data class decorator https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/pydantic.py#L15C7-L15C22

JackKelly:
Good idea! Unfortunately, the we can't (easily) use `FrozenBaseModel` in this context because we _do_ modify the `n_retries` field (and `FrozenBaseModel` forbids updates). So I've defined a `_StrictBaseModel`, which is identical to `FrozenBaseModel` but with `frozen` left as its default setting of `False`. Is that OK? Should I move `_StrictBaseModel` to `reformatters.common.pydantic.StrictBaseModel`?

aldenks (approved):
In practice we aren't getting much bug protection from setting those `strict=True, revalidate_instances="always"` here (i threw those on FrozenBaseModel because we needed a new base class and why not). I'd be fine just subclassing pydantic.BaseModel. Since you already have it may as well leave it, but i wouldn't add the additional mental overhead of adding StrictBaseModel to common.pydantic

PR #329, opened 2025-12-01
Add 2m dew point temperature to ECMWF IFS ENS

aldenks:
- I temporarily added this variable to the integration tests and they produced reasonable values but im not keeping it in to avoid slowing down the integration test
- This will cause a new, empty variable to be added to the dataset after it deploys and the next update runs. I'll then run a backfill for this variable specifically. While we could backfill and then make the new variable visible to readers only after that was complete when writing only to a standard zarr, Icechunk's validation is more strict (you can't write to vars that don't exist in the store) and we haven't yet implemented Icechunk's cooperative concurrent writes to do the backfill before exposing the new variable to readers. We won't document this new variable until it's backfilled and validated and the backfill period will only take a couple hours, so im expecting limited/no impact of this temporary period with an incomplete variable.


aldenks:
cursor review

PR #331, opened 2025-12-16
Calculate which FTP files to transfer to object storage for ICON-EU

### LINKED ISSUE #258: Save ICON-EU GRIB files from DWD's FTP server to cloud object storage

JackKelly:
## Outline of the classes

- `ObstoreManager`: Contains the `ObjectStore` and `dst_root_path` and the logic for calculating paths on the destination object store and listing files on object store.
- `FtpManager`: Knows how to list just the files on the FTP server for a given NWP initialisation (e.g. 00z), and filter those FTP files. And extract NWP variable name and init datetime from the FTP path.
- `DwdFtpManager` subclasses `FtpManager` and implements the DWD-specific logic.
- `FtpTransferCoordinator`: Contains an `ObstoreManager` and `FtpManager` and coordinates them :slightly_smiling_face:. It computes the set difference of the ftp_files - files_already_on_object_storage to only download new files. It considers the filesize.
- `DwdIconEuForecastDataset` now has an `archive_grib_files` method, and `operational_kubernetes_resources` returns an `ArchiveGribFilesCronJob` which, if we're very lucky :slightly_smiling_face: will run 4 times per day, and copy grib files to Source Co-Op. (To code should know how to get the credentials to upload to Source CoOp).

You can run the new code locally like this:

```
uv run main \
    dwd-icon-eu-forecast 
    archive-grib-files \
    --nwp-init-hour=6 \
    --filename-filter=single-level_.*_00[0-5]_ \ 
    --local-dst-root-path=/home/jack/data/ICON-EU
```

## Related

- Follows on from PR #318 
- Closes #258 
- #173 

## Tasks:

- [x] Write `class FtpTransferCalculator(ABC)`
- [x] Implement `DwdIconEuForecastDataset.archive_grib_files` (following [Alden's instructions](https://github.com/dynamical-org/reformatters/issues/258#issuecomment-3325224315))
- [x] Implement new `DwdFtpTransferCalculator` for the FTP -> Source CoOp transfers
- [x] More logging!
- [x] Allow users to massively limit the number of files downloaded for testing locally (e.g. only get the first hour step?) maybe allow users to provide a regex filename filter? This is just for testing. 
- [x] Only define `ftp_host` and `dst_store` once (instead of when calling `copy_files_from_ftp_to_obstore` and in `DwdFtpTransferCalculator`.
- [x] Write brief explanation of how the calculator works.
- [x] Test locally (on Jack's laptop)
- [x] remove paths to Jack's laptop used for hacking early testing! 
- [x] Refactor: Pass the `object_store` and `dst_path` into the calculator (don't define it in the subclass).
- [x] Allow the user to pass in local paths at the command line (for testing / running locally).
- [x] Test locally (passing in a local path).
- [x] pass in the ftp_host into the constructor
- [x] Write unit tests:
    - [x] save an actual FTP listing and an actual obstore listing to disk as text files (for testing) and compute which files to download
    - [x] more tests...
- [x] Check TODOs.


JackKelly:
OK, I think this is _finally_ ready for review! BUT it's Christmas. So please don't feel ANY pressure to look at this code any time soon. I'm gonna be on holiday until 5th Jan 2026 and plan to not look at any work stuff until then.

I've tested the code on my laptop for several days. It seems to work. I've been using this command to run on my laptop (to only download the first 5 steps):

```shell
uv run main \
    dwd-icon-eu-forecast 
    archive-grib-files \
    --nwp-init-hour=6 \
    --filename-filter=single-level_.*_00[0-5]_ \
    --local-dst-root-path=/home/jack/data/ICON-EU/grib/download_and_compress_and_concat_script/icon-eu/regular-lat-lon
```

I'll update the comment at the top of this PR with a quick overview of how the code works.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dev = [
     "pytest-asyncio~=1.3",
     "pytest-sugar~=1.0",
     "pytest-xdist~=3.8",
+    "pytest-mock~=3.14", # Added for mocker fixture
```
aldenks, line 49: you can remove this comment, it's clear enough why

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -1,55 +1,151 @@
...
+
+    # When running in prod, `secret` will be {'key': 'xxx', 'secret': 'xxxx'}.
+    # When not running in prod, `secret` will be empty. `S3Store()` won't
+    # accept `None` for `access_key_id` or `secret_access_key`.
+    if secret:
+        return S3Store(
+            bucket=bucket,
+            region=region,
+            access_key_id=secret["key"],
+            secret_access_key=secret["secret"],
```
aldenks, line 148: I'd add retries to this.
- Example structure: https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/download.py#L96-L104

- docs https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3Store.retry_config

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -1,55 +1,151 @@
...
+            schedule="0 4,10,16,22 * * *",
+            pod_active_deadline=timedelta(
+                minutes=120
+            ),  # TODO(Jack): Check how long it *actually* takes!
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="14",
+            memory="30G",
+            shared_memory="12G",
+            ephemeral_storage="30G",
```
aldenks, line 125: since this will be doing very little cpu work lets try making this smaller. fine to increase if we need. we also don't need shared memory (we just use that for sharing an array w/ multiprocessing)
```suggestion
            cpu="3",
            memory="7G",
            ephemeral_storage="30G",
```

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -1,55 +1,151 @@
...
+            dst_root_path = PurePosixPath(
+                "dynamical/dwd-icon-grib/icon-eu/regular-lat-lon/"
+            )
+            store = get_source_coop_s3_store()
 
-        # return [operational_update_cron_job, validation_cron_job]
-        raise NotImplementedError(
-            f"Implement `operational_kubernetes_resources` on {self.__class__.__name__}"
+        ftp_manager = DwdFtpManager(
+            ftp_host="opendata.dwd.de", filename_filter=filename_filter
```
aldenks, line 55: Setting a default ftp host on the DwdFtpManager makes sense to me

src/reformatters/common/kubernetes.py
```diff
@@ -251,12 +251,19 @@ class ValidationCronJob(CronJob):
     parallelism: int = 1
 
 
+class ArchiveGribFilesCronJob(CronJob):
+    command: Sequence[str] = ["archive-grib-files"]
+    workers_total: int = 1
+    parallelism: int = 1
```
aldenks, line 257: In the interest of creating fewer things for readers to keep track of, lets remove this and operational_kubernetes_resources use CronJob directly, passing the command, workers, etc there

src/reformatters/common/ftp_manager.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def extract_init_datetime_from_ftp_path(ftp_path: PurePosixPath) -> datetime:
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def extract_nwp_variable_name_from_ftp_path(ftp_path: PurePosixPath) -> str:
+        pass
```
aldenks, line 108: This might not be as general as we hope. I know many NWPs that don't store the variable name in the file path. Init in the filepath is pretty dang common, that feels necessary to drive logic and i've yet to find a NWP that doesn't do that.

I'm not sure how much work it is, but if instead of trying to parse out var names we preserved the directory / path structure of the source when copying into the destination that'd remove needing to have this understand variable name as a concept.

src/reformatters/common/ftp_manager.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @staticmethod
+    def skip_ftp_item(ftp_path: PurePosixPath, ftp_info: UnixListInfo) -> bool:
+        """Skip FTP items that we don't need."""
+        return (
+            ftp_info["type"] == "dir"
+            or "pressure-level" in ftp_path.name
+            or not ftp_path.name.endswith("grib2.bz2")
+        )
```
aldenks, line 117: these belong in the DWD specific one yeah?

src/reformatters/common/ftp_to_obstore_types.py
```diff
@@ -0,0 +1,22 @@
...
+class TransferJob(BaseModel):
+    src_ftp_path: PurePosixPath
+    src_ftp_file_size_bytes: int
+    dst_obstore_path: PurePosixPath
+    nwp_init_datetime: datetime
+
+
+class PathAndSize(NamedTuple):
+    # We have to use NamedTuple because we want to use PathAndSize objects in a `set`.
+    path: str  # No slash at the start of this string!
```
aldenks, line 18: if you subclass our FrozenBaseModel the frozen aspect will make the object hashable. You could then have pydantic enforce the no leading slash for you:

```
from typing import Annotated
from pydantic import BaseModel, StringConstraints

from reformatters.common.pydantic import FrozenBaseModel

class PathAndSize(FrozenBaseModel):
     path: Annotated[str, StringConstraints(pattern=r"^[^/].*")]
```

src/reformatters/common/ftp_transfer_coordinator.py
```diff
@@ -0,0 +1,116 @@
...
+
+from reformatters.common.ftp_to_obstore_types import (
+    PathAndSize,
+    TransferJob,
+)
+from reformatters.common.logging import get_logger
+
+if TYPE_CHECKING:
+    from reformatters.common.ftp_manager import FtpManager
+    from reformatters.common.obstore_manager import ObstoreManager
```
aldenks, line 15: I don't expect any slow nor cyclic imports, you can import this without a TYPE_CHECKING guard

src/reformatters/common/ftp_manager.py
aldenks: oh, do we need to control the structure in the dest so we can efficiently know what we've already copied? we might just need to leave

src/reformatters/common/ftp_manager.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, ftp_host: str, filename_filter: str = "") -> None:
+        self.ftp_host = ftp_host
+        self.filename_filter = filename_filter
+
+    async def list_and_filter_for_init_hour(
+        self,
+        init_hour: int,
+    ) -> Sequence[FtpPathAndInfo]:
+        ftp_path = self.convert_nwp_init_hour_to_ftp_path(init_hour)
+        ftp_host_and_path = f"ftp://{self.ftp_host}{ftp_path}"
```
aldenks, line 33: What happens if the paths at the source don't begin with an init _hour_ only? e.g. a lot of nwps do something like "2025-12-01/00/..."?

src/reformatters/common/ftp_transfer_coordinator.py
```diff
@@ -0,0 +1,116 @@
...
+        self,
+        ftp_manager: "FtpManager",
+        obstore_manager: "ObstoreManager",
+    ) -> None:
+        self.ftp_manager = ftp_manager
+        self.obstore_manager = obstore_manager
+
+    async def calc_new_files_for_multiple_nwp_init_hours(
+        self,
+        nwp_inits_to_transfer: Sequence[int] = (0, 6, 12, 18),
```
aldenks, line 39: I'd have this be a required argument (no default) so we don't bake in assumptions about init frequency

src/reformatters/common/ftp_manager.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+
+    @staticmethod
+    @abstractmethod
+    def convert_nwp_init_hour_to_ftp_path(init_hour: int) -> PurePosixPath:
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def sanity_check_ftp_path(ftp_path: PurePosixPath) -> None:
```
aldenks, line 97: slight preference for "validate" or "check" vs sanity

aldenks (changes requested):
Thanks for all this!

Two overarching comments
* There's a couple assumptions about the file path structure at the source that are likely to not be general. It's nice to aim for generality but we don't technically need this to be general out of the box. I know it'll work well for any DWD dataset so what about moving all the common/ftp_* and common/obstore_manager.py files into reformatters/dwd/ftp_obstore_sync for now and then we can generalize if/when we need to? The two main assumptions that are likely to hold this back from being general that I see are
  * Directories at the source starting with the init _hour_ only
  * One variable per file and the variable name in the file name
  * If you see a way to remove those assumptions without much work, even better, but lets still move them all into their own directory (e.g. reformatters/common/ft_obstore_sync/).

* There's a lot of related classes to keep track of here. will you add a brief readme in the new dir that lists each module/class and how they relate?

JackKelly:
This PR may be replaced by the much simpler approach proposed in PR #360 :slightly_smiling_face:

aldenks:
Closing in favor of #360 !

src/reformatters/common/ftp_manager.py
JackKelly: Interesting point! For DWD, my understanding is that we do want to use a different directory structure in Source Co-Op than DWD use on their FTP server. DWD's FTP paths look like this: `https://opendata.dwd.de/weather/nwp/icon-eu/grib/00/t_2m/`. If we copied DWD's directory structure to Source Co-Op then we'd end up with literally millions of objects in each bucket leaf directory, which I think will kill performance when reading from the bucket. (I believe that Dexter Energy originally copied DWD's directory structure to Dexter's GCS bucket, but it ended up being far too slow)

And, as you say, it's probably easier and more efficient to figure out what we've already copied if we use a directory structure more like `/2026-01-12T00Z/t_2m/`.

That said, if we did just copy DWD's directory structure then we could do the copy with almost a single line of `rclone` (although we'd probably still want a bunch of Python to parse `rclone`'s logs etc.)

PR #359, opened 2026-01-14
Dataset CF compliance

aldenks:
Adopt CF Conventions across all datasets by updating coordinate and data variable attributes. Adds validation to check all datasets against CF standard name + units table and `cf_xarray`.

Shout out Deepak Cherian for his earlier review of our datasets and all `cf_xarray` maintainers.

### Unit Changes Review

These are all standard CF convention formatting changes or equivalent representations (with one exception, note in final row):

| Old | New | Reason |
|-----|-----|--------|
| `degrees_north` / `degrees_east` | `degree_north` / `degree_east` | CF convention singular form |
| `m³/m³` | `1` | Dimensionless ratio |
| `unitless` | `1` | CF convention for dimensionless |
| `categorical` | `1` | Dimensionless |
| `%` | `percent` | CF convention |
| `m/s` | `m s-1` | CF convention formatting |
| `C` | `degree_Celsius` | CF convention |
| `W/(m^2)` | `W m-2` | CF convention formatting |
| `kg/(m^2)` | `kg m-2` | CF convention formatting |
| `realization` | `1` | Ensemble member is dimensionless |
| `0=no; 1=yes` | `1` | Categorical → dimensionless. Categories now in `comment` attribute. |
| `gpm` | `m` | Geopotential meters ≈ meters (under standard gravity) |
| `mm/s` | `kg m-2 s-1` | Equivalent for water (1 mm/s = 1 kg m⁻² s⁻¹) |
| `kg m**-2`  (snow_depth_water_equivalent) | `mm` | Equivalent for water (1 kg/m² = 1 mm depth) |
| `mm h20` / `mm snow thickness` | `mm` | Simplified but equivalent |
| `kg m**-2`  (DWD ICON EU precipitation_surface) | `kg m-2 s-1` | Data is now deaccumulated to rates; units/standard_name updated to match **This is a real change, but to an unpublished and in progress dataset.** Aligns with conventions in our datasets.  |


tests/common/common_template_config_subclasses_test.py
```diff
@@ -203,3 +204,233 @@ def test_coordinates_not_sharded(
...
+    template_path = template_config.template_path()
+
+    ds = xr.open_zarr(template_path)
+
+    # Get all standard names from cf_xarray
+    recognized_standard_names = ds.cf.standard_names
+
+    # Check each data variable that has a standard_name configured
+    for var_config in template_config.data_vars:
+        if var_config.attrs.standard_name is not None:
```
aldenks: @cursor Search for all DataVarAttrs instantiations that don't have a standard_name. For each case, fill in the standard name if CF Conventions defines one. In this test, create an explicit list of `var_config.name`s in our codebase that CF conventions does not define standard names for. This test should fail if the ds[var].attrs doesn't have a standard name defined and the var_config.name is not in the allowed missing list

aldenks: @copilot Search for all DataVarAttrs instantiations that don't have a standard_name. For each case, fill in the standard name if CF Conventions defines one. In this test, create an explicit list of var_config.names in our codebase that CF conventions does not define standard names for. This test should fail if the ds[var].attrs doesn't have a standard name defined and the var_config.name is not in the allowed missing list

src/reformatters/noaa/gefs/forecast_35_day/template_config.py
```diff
@@ -158,7 +160,9 @@ def coords(self) -> Sequence[Coordinate]:
                     shards=None,
                 ),
                 attrs=CoordinateAttrs(
-                    units="realization",
+                    long_name="Ensemble member",
+                    standard_name="realization",
+                    units="1",
```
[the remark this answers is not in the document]
aldenks: https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

aldenks:
I've manually reviewed every one of these changes and had AIs review them as well. The table in the PR description above summarizes the changes to existing attributes: format changes to effectively equivalent units.

This PR may break users who are doing string matching against unit strings, however it does _not_ remove any attributes, change the type of any attributes, change the keys of data variables in the dataset, nor change any values in the dataset. Additionally, new attributes are being added but that is unlikely to cause issues for existing code. I think the changes are minimal and easy enough to fix in the unlikely case they should break any users code and combined with the benefits of standardizing relatively early on in our existence make this worthwhile.

PR #360, opened 2026-01-14
Copy files from DWD FTP using `rclone`

### LINKED ISSUE #258: Save ICON-EU GRIB files from DWD's FTP server to cloud object storage

JackKelly:
I'm optimistic this is a concise and general solution to #258.

This PR, if accepted, almost completely replaces my two previous PRs:

1. #331 (+674 lines)
2. #318 (+619 lines)

Compared to my previous two PRs, this PR should be more concise, more general, more robust, and easier to maintain :slightly_smiling_face:. No more fiddly `async` Python code.

The "trick" in this PR is to use [`rclone`](https://rclone.org) for the heavy lifting of [listing](https://rclone.org/commands/rclone_lsf/) and [copying](https://rclone.org/commands/rclone_copy/) new FTP files. 

The only downside of this PR is that we'll need to install `rclone` (which is just `sudo snap install rclone` on Ubuntu). Please let me know if that's acceptable.

In a subsequent PR I'll implement the `DynamicalDataset` code to call this new FTP copying code.

## Why am I suggesting deleting a bunch of code I've already written?!

Last year, before I wrote PR #331 & #318, I briefly looked into whether an `rclone` one-liner could copy DWD's files for us. I concluded the answer is "no" due to a limitation that `rclone` processes path parts one-by-one, left-to-right, and so cannot transform an earlier path segment based on a later path segment (this limitation is described in more detail [here](https://github.com/dynamical-org/reformatters/blob/3e219af92b65dd28c585e528916378522ff8749e/src/reformatters/dwd/copy_files_from_dwd_ftp.py#L19-L34)). This is a deal-breaker for us because we want to copy files into paths on Source Co-Op of the form `<NWP init datetime>/<NWP variable name>/<filename>`, but the NWP init datetime only exists in the leaf filename.

But then, over the Christmas break, whilst tinkering with `rclone` for a personal project, I found out that it _is_ possible to get `rclone` to do what we want, by combining `rclone` with a bit of Python pre-processing. This means that we no longer have to write our own code for listing FTP files & concurrent downloading! Yay!

I'm really sorry to have taken up your time reviewing my previous two PRs #331 & #318! But I hope this current PR will be a significant improvement.

## Test locally

`uv run src/reformatters/dwd/run_copy_test.py /path/to/local/data/directory/ --max-files 10`

See the README in this PR for more info.

## TODO: 

- [ ] Handle errors like this when the file disappears from FTP... maybe re-check the FTP listing if we get a 550 error?

```
2026-01-14T20:56:49Z INFO:reformatters.dwd.copy_files_from_dwd_ftp:Batch [88/93]: Asking rclone to copy 6975 file(s) to /home/jack/data/ICON-EU/grib/rclone_with_python/2026-01-13T18Z/w (if they don't already exist)...
2026-01-14T21:01:20Z ERROR:reformatters.dwd.copy_files_from_dwd_ftp:rclone: 2026/01/14 20:57:58 ERROR : w/icon-eu_europe_regular-lat-lon_model-level_2026011318_036_11_W.grib2.bz2: Failed to copy: failed to open source object: open: 550 Failed to open file.
2026-01-14T21:01:20Z ERROR:reformatters.dwd.copy_files_from_dwd_ftp:rclone: 2026/01/14 20:57:58 ERROR : w/icon-eu_europe_regular-lat-lon_model-level_2026011318_036_13_W.grib2.bz2: Failed to copy: failed to open source object: open: 550 Failed to open file.
2026-01-14T21:01:20Z ERROR:reformatters.dwd.copy_files_from_dwd_ftp:rclone: 2026/01/14 20:57:58 ERROR : w/icon-eu_europe_regular-lat-lon_model-level_2026011318_036_14_W.grib2.bz2: Failed to copy: failed to open source object: open: 550 Failed to open file.
```


pyproject.toml
JackKelly, line 1: These changes to `pyproject.toml` simply reverse the changes I made in PR #318.

aldenks:
sweet! very supportive of a simpler approach. let me think about if we should install rclone in our main docker image or run this on a specific one (or just install it on the fly?). I'll give the code itself a review soon too

JackKelly:
Follow-on PR: https://github.com/JackKelly/reformatters/pull/1

src/reformatters/dwd/copy_files_from_dwd_ftp.py
```diff
@@ -0,0 +1,369 @@
...
+    total_files = 0
+    for file_to_be_copied in file_list:
+        if "pressure-level" in file_to_be_copied.path.name:
+            continue
+
+        if len(file_to_be_copied.path.parts) != n_expected_path_parts:
+            log.warning("Unexpected path structure: %s", file_to_be_copied.path)
+            continue
+
+        match = date_regex.search(file_to_be_copied.path.name)
```
aldenks: I would .findall rather than .search and then assert there is no more than one match so we start erroring loudly up if they ever put another 10 digit string in their file paths.

src/reformatters/dwd/copy_files_from_dwd_ftp.py
```diff
@@ -0,0 +1,369 @@
...
+
+    Args:
+        ftp_host: The FTP host (e.g. 'opendata.dwd.de').
+        ftp_path: The root source path on the FTP server (including NWP init hour).
+        dst_path: The specific destination directory (including NWP init timestamp).
+        files_to_be_copied: List of file path and sizes relative to ftp_path to be copied.
+        transfers: Number of parallel transfers to use for this batch.
+    """
+    # Modern Linux platforms often install `rclone` as a sandboxed snap, which does not have access
+    # to `/tmp`, so we store the temporary file in the current working directory.
```
aldenks, line 298: in docker we can/would apt install it but this is all good as is

src/reformatters/dwd/copy_files_from_dwd_ftp.py
```diff
@@ -0,0 +1,369 @@
...
+    log_entries = parse_and_log_rclone_json(result.stderr)
+    return result.stdout, log_entries
+
+
+def _set_death_signal() -> None:
+    """Linux-specific: Ensure the child process dies if the parent dies."""
+    pr_set_pdeathsig: Final[int] = 1  # For Linux prctl (PRocess ConTroL)
+    libc = ctypes.CDLL("libc.so.6")
+    # Send SIGTERM to the child if the parent terminates.
+    libc.prctl(pr_set_pdeathsig, signal.SIGTERM)
```
aldenks, line 377: Do we need this? I'm pretty sure that in kubernetes if the entry point process dies the pod would exit with a failure exit code and the whole container would be terminated

src/reformatters/dwd/README.md
aldenks, line 1: I'd clarify in this file that, unlike many other datasets, we are both copying complete files as-is to build an archive of the source files, and (to be implemented), transforming those into zarr.

aldenks (approved):
Much simpler! Good call.

We can add rclone to our docker image. I tested and it will only increase the size by ~3%. To do that add `rclone \` below this line https://github.com/dynamical-org/reformatters/blob/main/deploy/Dockerfile#L121 (that can happen in a follow up PR when we're ready to start running this)

src/reformatters/dwd/copy_files_from_dwd_ftp.py
JackKelly: Yeah, good question! I definitely see your point that we probably don't need this on kubernetes. But I think this might be useful for testing locally. After each tweak to the code, I found myself running the code to copy data from DWD to my laptop, and then hitting `ctrl-c` to kill Python after it had successfully downloaded a handful of files. Before I implemented `_set_death_signal`, `rclone` would just keep chugging away silently in the background as a zombie even after I killed Python. Which was very confusing until I figured out what was going on!

aldenks: ahh got it, lets keep -- we want local testing to be easy

JackKelly:
Thanks loads for the quick review! Exciting stuff! As you suggested, I've added `rclone \` to `Dockerfile` in follow-on PR #370

PR #370, opened 2026-01-19
Production code to copy ICON-EU grib files from FTP to Source Co-Op

### LINKED ISSUE #258: Save ICON-EU GRIB files from DWD's FTP server to cloud object storage

JackKelly:
This PR follows on from PR #360.

This PR will implement the code necessary to copy gribs from DWD's FTP server to Dynamical's Source Co-Op bucket.

TODO:
- [x] Copy & adapt relevant `DwdIconEuForecastDataset` code from PR #331
- [x] Test!

If we're lucky then this code _should_ be ready to start copying DWD's ICON-EU grib files to Source Co-Op! I've tested locally (see README for how).

A follow-on PR will add more unit-tests. And another PR will finish the Zarr conversion code :slightly_smiling_face: 

## Related issues

- #176 
- #173

This PR closes #258


JackKelly:
@aldenks would you like me to send metrics to Sentry? (using Sentry's new ["metrics" feature](https://docs.sentry.io/product/explore/metrics/), which is still in beta). e.g. send the number of files copied, and the total number of bytes copied, and the transfer speed?

aldenks:
@JackKelly oo that looks useful. I'd say leave them out of this first pass and add them as a follow up

.gitignore
```diff
@@ -166,3 +166,5 @@ cython_debug/
 .aider*
 
 depot.json
+
+.rclone_files_*
```
aldenks, line 171: Add the same to .dockerignore too please. Right? These are local state not something we'd want to build in?

aldenks (changes requested):
I'll give a closer look tomorrow but from a quick skim this looks great!

.gitignore
JackKelly: Great point, thanks! Done!

aldenks:
I haven't forgot about this but it might be this weekend when i can review

JackKelly:
No big rush! Please don't feel you have to work at the weekend on this PR :slightly_smiling_face:

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -13,36 +19,36 @@ class DwdIconEuForecastDataset(
 ):
     template_config: DwdIconEuForecastTemplateConfig = DwdIconEuForecastTemplateConfig()
     region_job_class: type[DwdIconEuForecastRegionJob] = DwdIconEuForecastRegionJob
+    grib_archive_path: str = "s3:us-west-2.opendata.source.coop/dynamical/dwd-icon-grib/icon-eu/regular-lat-lon/"
```
aldenks, line 25:
```suggestion
    grib_archive_path: str = "s3://us-west-2.opendata.source.coop/dynamical/dwd-icon-grib/icon-eu/regular-lat-lon/"
```

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -13,36 +19,36 @@ class DwdIconEuForecastDataset(
...
+            name=f"{self.dataset_id}-archive-grib-files",
+            # We want the 00, 06, 12, and 18 ICON-EU runs. DWD's transfer to their FTP server starts
+            # about 2 hours 15 mins after the init time, and finishes about 3 hours 45 minutes after
+            # the init hour. So, to avoid copying incomplete files, we fetch the files 4 hours after
+            # each init. But, every time the cron job runs, the script checks all 4 NWP inits, to
+            # keep the code simple, especially when recovering if the script hasn't run for a while.
+            # It only takes 4 minutes to check an NWP run that we've already transferred.
+            schedule="0 0 4,10,16,22 * *",
+            # Copying 1 NWP run takes 45 minutes on a 1 Gbps internet connection. But, when this
+            # script first runs, or if it hasn't run for >6 hours, then it'll transfer up to 4 NWP runs.
```
aldenks, line 45: very helpful thank you

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -53,3 +59,52 @@ def validators(self) -> Sequence[validation.DataValidator]:
...
             f"Implement `validators` on {self.__class__.__name__}"
         )
+
+    def archive_grib_files(
+        self,
+        dst_root: str = grib_archive_path,
+        # The `type: ignore` on the line below is because Typer doesn't understand the type hints
+        # `tuple[int, ...]` or `Sequence[int]`, so we have to use `list[int]`.
+        nwp_init_hours: list[int] = (0, 6, 12, 18),  # type: ignore[assignment]
+        transfers: int = 10,
```
aldenks: transfer_parallelism?

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -53,3 +59,52 @@ def validators(self) -> Sequence[validation.DataValidator]:
         raise NotImplementedError(
             f"Implement `validators` on {self.__class__.__name__}"
         )
+
+    def archive_grib_files(
+        self,
+        dst_root: str = grib_archive_path,
+        # The `type: ignore` on the line below is because Typer doesn't understand the type hints
+        # `tuple[int, ...]` or `Sequence[int]`, so we have to use `list[int]`.
+        nwp_init_hours: list[int] = (0, 6, 12, 18),  # type: ignore[assignment]
```
aldenks, line 74: if someone actually uses the arg we'll get a list and loose the immutability anyways. I think you're safe to make the default a list, or get fancy and keep the arg here nwp_init_hours and then in the body do

```
init_hours = tuple(nwp_init_hours)
del nwp_init_hours
```
and use init_hours below

src/reformatters/dwd/copy_files_from_dwd_ftp.py
```diff
@@ -94,17 +95,24 @@ def copy_files_from_dwd_ftp(
...
     max_files_per_nwp_variable: int = sys.maxsize,
+    env_vars: dict[str, Any] | None = None,
 ) -> TransferSummary:
     """Restructure DWD GRIB files from FTP to a timestamped directory structure.
 
     Args:
         ftp_host: The FTP host, e.g. 'opendata.dwd.de'
         ftp_path: The source path on the FTP host, e.g. '/weather/nwp/icon-eu/grib/00'
-        dst_root: The destination root directory.
+        dst_root: The destination root directory. e.g. for S3, the dst_root could be: 's3:bucket/foo/bar'
```
aldenks, line 105:
```suggestion
        dst_root: The destination root directory. e.g. for S3, the dst_root could be: 's3://bucket/foo/bar'
```

aldenks (approved):
A couple very minor suggestions, but looks great. Once this merges is there anything else before I should unsuspend it start it running in prod? (I'll watch the first couple runs closely in addition to our usual sentry alerting)

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
JackKelly: Ah, yes, sorry, you're right that it's surprising that `rclone` doesn't want a double slash after `s3:`! In commit 8415d93 I've added a comment to explain that `rclone` wants `s3` paths in the form `s3:<bucket>/<path>` (with no double slash).

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
JackKelly: Yeah, I 100% agree that `transfer_parallelism` is a better name! I've changed the name throughout our code to `transfer_parallelism` in commit d5c541d. (`transfers` is the arg name that `rclone` uses. But I totally agree that it's not an ideal name!)

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
JackKelly: Ah, yes, this is a fun design mistake deep within Python :slightly_smiling_face:. The boring reason not to use a list as a default argument is that Ruff complains "Do not use mutable data structures for argument defaults [B006]". The deeper problem - and the _reason_ that Ruff complains - is that Python re-uses the default argument value across function calls, so you get truly horrible behaviour like this:

```python
def add_to_list(item, some_list=[]):
    some_list.append(item)
    return some_list

l1 = add_to_list(0)  # [0]
l2 = add_to_list(1)  # [0, 1]
```

The Ruff docs for the B006 error explain the issue: https://docs.astral.sh/ruff/rules/mutable-argument-default/

src/reformatters/dwd/copy_files_from_dwd_ftp.py
JackKelly: In commit 7fcc275 I've added a comment to explain the surprising format that `rclone` uses for its `s3` paths!

JackKelly:
Thanks for the review!

> Once this merges is there anything else before I should unsuspend it start it running in prod?

I _think_ it should be good to go! (It works locally on my machine.)

Exciting!

But I'm fully expecting there will be some glitches to iron out once it starts running. If it makes life easier for you then feel free to add me to Sentry and I can keep an eye on it if you'd like. But absolutely no pressure to do that!

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
aldenks: oo yes right, good point and thank you for not listening to me :)

PR #375, opened 2026-01-21
ICON-EU: Try downloading from Source Co-Op first, then try DWD's HTTP server

JackKelly:
Short PR implementing Alden's suggestion in [this comment](https://github.com/dynamical-org/reformatters/issues/173#issuecomment-3778649201):

First, try to download the `.grib2.bz2` file from Dynamical's Source Co-Op ICON-EU archive. If that fails then try downloading from DWD's HTTP server. If that fails then raise `FileNotFoundError`.

Unlike [in the GEFS code](https://github.com/dynamical-org/reformatters/blob/1d2d35fd689f2c69bb01e7632c609265f14b5751/src/reformatters/noaa/gefs/utils.py#L45), when we try downloading from DWD, I've decide not to explicitly check if `pd.Timestamp.now()` is within 24 hours of the requested init time, and instead just let `download_http_to_disk` raise `FileNotFoundError` if it can't find the file on DWD's HTTP server. I figure there's little harm to trying to fetch an old file. We'll still noisily come crashing to a halt. But LMK if you'd like me to implement a time check.


src/reformatters/dwd/icon_eu/forecast/region_job.py
```diff
@@ -126,12 +140,27 @@ def generate_source_file_coords(
...
+            if isinstance(e, GenericError):
+                # GenericError contains multiple lines that aren't informative. So, log gory details
+                # at the debug log level, and shorten the error message for the warning log level.
+                error_msg = str(e).splitlines()[0]
+                log.debug("%s", e)
+            else:
+                error_msg = str(e)
+            log.warning(
+                f"{error_msg}. Failed to download '{url}'. Attempting to download from {fallback_url=}"
+            )
```
aldenks: You can actually skip any warn/info logging in this method and let the exception from the fallback source bubble up and the RegionJob code will handle logging. Ideally catch a GenericError and raise a FileNotFound error if that's what it generally represents -- we [log more succinctly](https://github.com/dynamical-org/reformatters/blob/main/src/reformatters/common/region_job.py#L626-L633) in a FileNotFound case as that's a usual part of updating very recent data.

To get lower latency on our ICON-EU zarr updates I expect we'll want to run the zarr update asap rather than waiting until we're confident the dwd ftp -> source coop has completed. That means we'll likely be downloading directly from dwd often and therefore i wouldn't want to spew warning lines in an expected case. 

debug logging is fine to keep, but honestly we will probably never see it because our default level is info :)

aldenks (approved):
Awesome thanks! Approved after comment

src/reformatters/dwd/icon_eu/forecast/region_job.py
JackKelly: Great point. I feel uncomfortable about this bit of code, too. I've simplified the logging in commit 402691e.

JackKelly: (I haven't yet converted the `GenericError` to a `FileNotFoundError` because I'm not yet 100% certain that all `GenericError`s represent `FileNotFoundError`, and I'm assuming we'd prefer verbose debug logs instead of outright _misleading_ log messages!)

PR #378, opened 2026-01-24
Add shared memory estimate to chunk_shard_size tool

aldenks:
Adds a new section to the diagnostic output that estimates the shared
memory required for processing a single slice along the append dimension.

The shared memory estimate calculates the total uncompressed data that
needs to be held in memory when processing one slice along the append
dimension (init_time for forecast mode, time for analysis mode).

Calculation:
- Count shards touched along all dimensions except append_dim
- Multiply by uncompressed shard size
- Display result in GB

Example: For a dataset with 4 shards along longitude, 2 along latitude,
2 along lead_time, and 1 along ensemble_member, with init_time as the
append dimension, the estimate is:
  4 × 2 × 2 × 1 = 16 shards touched
  If uncompressed shard size is 2GB, shared memory = 16 × 2GB = 32GB

The estimate appears in the ACCESS PATTERN COSTS section of the
diagnostic output and helps users plan memory requirements for their
zarr processing workflows.

### DIFF
--- a/docs/chunk_shard_layout_tool.md
+++ b/docs/chunk_shard_layout_tool.md
@@ -152,6 +152,7 @@ The tool outputs a comprehensive ASCII table with:
 1. **Spatial Slice**: Chunks/data needed to read full spatial domain for single time
 2. **Shard Touch**: Total shard data opened for spatial slice
 3. **Full Time Series**: Shards/data needed for complete time series at single location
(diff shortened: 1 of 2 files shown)


aldenks:
@claude fix lints: https://github.com/dynamical-org/reformatters/actions/runs/21315481767/job/61357602684#step:8:1

PR #382, opened 2026-01-26
Fix typo in comment: Replace March with May

JackKelly:
This is a tiny, one-line PR, which I _think_ fixes a typo in the comments in ECMWF `region_job_test`. But I may have misunderstood!


aldenks (approved):
You're right, thank you!

PR #389, opened 2026-01-28
Improve test suite speed by ~25%

aldenks:
- Reuse ProcessPoolExecutor across _write_shards calls within a single RegionJob.process() call
- Reduce lead_time dimensions in tests to minimum needed for assertions
- Reduce backfill time ranges where possible
- Reduce test_region_job_empty_chunk_writing from 4 to 2 variables

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -36,23 +36,24 @@ jobs:
       - name: Install Arm Build Dependencies
         if: matrix.arch == 'arm64'
         run: |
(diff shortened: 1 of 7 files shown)


aldenks:
follow ons
- update to rasterio 1.5.0 with prebuilt arm wheels and skip gdal install step
- in heaviest backfill + update integration tests limit shards written by monkeypatching shared_memory_utils's shard_slice_indexers (from common.iterating) to filter down to just shards needed by the values checked in the test.

PR #394, opened 2026-01-28
Update large swath of deps, including zarr to 3.1.5

aldenks:


aldenks:
befuddled by mypy failure i can't reproduce locally. trying fresh branch #399

PR #401, opened 2026-01-30
Replace pre-commit with prek

samn:
It's a drop-in replacement that's faster.
See https://prek.j178.dev/


aldenks:
ooo sick! thank you

Issue #390, opened 2026-01-28
Research into speeding up `rclone copy` from DWD's FTP server to Source Co-Op


JackKelly:
Follows on from #258.

This issue is just a dumping ground for my raw notes. Don't worry about following along 🙂 . Once I've come to some solid conclusions, I'll close this issue and start a new one with a concise summary & a list of options.

Alden found that `rclone` is being super-slow (2.3 MiB/s) when copying grib files from DWD's FTP server to our Source Co-Op bucket on S3 (complete [raw logs here](https://gist.github.com/JackKelly/23ba8e0e742f26905667931d4b07c23b)):

```
2026-01-27T20:32:12Z INFO:reformatters.dwd.copy_files_from_dwd_ftp:Batch [45/93] starting: Asking rclone to copy 6882 file(s) totalling 2.650 GiB to :s3:us-west-2.opendata.source.coop/dynamical/dwd-icon-grib/icon-eu/regular-lat-lon/2026-01-27T00Z/qi (if they don't already exist)...
2026-01-27T20:53:42Z INFO:reformatters.dwd.copy_files_from_dwd_ftp:Batch [45/93] complete: 6882 files transferred, 0.000 MiB/sec, 2.650 GiB total transferred, 0 files checked, 0 errors, 1290.375 seconds rclone runtime, 0.000 seconds transfer time, 0 directories listed.
```

This is despite [our code using the flags](https://github.com/dynamical-org/reformatters/blob/6e93151aba1454f4cffcb935e859528c24e8aeb6/src/reformatters/dwd/copy_files_from_dwd_ftp.py#L319-L332) for `rclone copy` that the `rclone` documentation recommends, specifically: 

- `--transfers=10`  # Probably the max number of concurrent FTP requests DWD can support
- `--ignore-checksum`  # DWD's FTP server does not support hashing.
- `--update`   # Skip files that are newer on the destination.
- `--fast-list`  # Use less API calls to S3, in exchange for using more RAM.

My plan is to:

- [x] Set up a temporary S3 bucket for testing. (I plan to use a real AWS S3 bucket instead of a mock local bucket because a local mock S3 bucket would hide all the latencies we see with a real S3 bucket)
- [x] Save a `.rclone_files_*` file to disk.
- [x] Run `rclone copy --files-from-raw` at the command line, with `--progress -vv --dump headers` which _should_ list every network request `rclone` makes.
- [x] Fiddle with `rclone`'s command line arguments to see if I can speed things up.
[...]

PR #406, opened 2026-01-31
Transfer files from DWD HTTPs server to Source Co-Op using `rclone copyurl --urls`

### LINKED ISSUE #392: Speed up `rclone` copy from DWD's FTP server to Source Co-Op
## Summary of the problem
Alden found that `rclone` is being super-slow (~2.3 MiB/s) when copying grib files from DWD's FTP server to our S3 Source Co-Op bucket (complete [raw logs here](https://gist.github.com/JackKelly/23ba8e0e742f26905667931d4b07c23b)). This compute job ran in `us-west-2`.

There are several issues & causes:

- Many of the files on DWD's FTP server are _tiny_ (some of the CLC files are just a few hundred bytes). (I plan to email DWD to raise this issue with them).
- This isn't necessarily a deal-breaker, but we need lots (>=128) of concurrent uploads to S3 to achieve acceptable speeds.
- But `rclone` doesn't allow users to set a different number of threads for FTP and S3. So, if we try to transfer from FTP to S3 with one command to `rclone`, we're stuck with using 10 concurrent threads for FTP _and_ 10 for S3. That's far too little concurrency for S3. We're stuck with 10 because DWD's FTP server doesn't allow more than 10 concurrent connections from one IP address.
- As well as all this, FTP _hates_ the geographical distance from Germany to the US. I can get 21 MiB/s transferring from DWD's FTP in Germany to my laptop in London on a domestic gigabit internet connection. In stark contrast, an EC2 VM in `us-east-1` only gets 7 MiB/s from DWD's FTP server to the VM's fast SSD. Gemini says:
> "the drop from 21 MiB/s (London) to 7 MiB/s (Virginia) is a classic example of how latency kills TCP-based protocols like FTP. Because FTP requires an acknowledgement for every small "window" of data sent, increasing the distance (adding ~80-100ms of round-trip time across the Atlantic) creates a bottleneck that no amount of bandwidth can fix."
- As a result, transferring the model-level directories (with 6882 tiny files in each directory) from DWD's FTP to S3 currently takes about 28 minutes, when the compute is running in `us-west-2`. (In contrast, I can comfortably download an entire NWP run to my laptop's SSD in 45 minutes).

### Move compute to Germany 
- An EC2 VM in Frankfurt (`eu-central-1`) gets 64 MiB/s from DWD's FTP server to the VM's local SSD with `--transfers=10` (compared to 7 MiB/s to a VM in `us-east-1`). Can we run our FTP transfer job in `eu-central-1`? Even if it's just a single VM on its own? (not in kubernetes?)

### Split the transfers into two: 1) FTP-to-local-disk; 2) local-disk-to-S3
- Even on the VM in Frankfurt, I'm only getting 7.3 MiB/s when transferring from FTP to S3 in one go, presumably because S3 wants many more concurrent connections than the 10 we're limited to.
- We get much higher throughput by splitting the transfer: The VM in Frankfurt gets 64 MiB/s from FTP to local disk (with `--transfers=10`); and about 55 MiB/s from local disk to S3 in `us-east-1` (with `--transfers=256`).
- So I'd propose that our code could have two threads: a download thread and an upload thread, connected by a queue. The download thread tells `rclone` to fetch files from FTP and save to local disk (with --transfers=10). When it finishes downloading an entire directory, it puts a message in a queue to tell the upload thread that it's safe to start uploading that local directory to S3, using a huge amount of concurrency (256 seems to work well, but I haven't rigorously tested different settings). The queue would have max size of, say, 2 items, to apply back-pressure to the FTP downloading thread. (Note that the `aws` cli seems to forget to transfer some files when using a high concurrency limit on thousands of files. In contrast, `rclone` is much more reliable.)

## If the fixes above don't provide enough speed:
- We could zip the files per directory after downloading to local disk, and upload the zip to S3. Uncompressed zip contains a "table of contents" at the end of the zip, so people can still lazily download only the bytes they need, using standard tools. This would also massively reduce the number of S3 HTTP requests when reading and writing these files (which could save Jed some money?!). And putting all the gribs into a zip per directory would make it super-fast to figure out what still needs to be transferred. And it should be safe because, IIUC, S3's uploads are atomic; so if the VM gets pre-empted while it's uploading a zip then the zip will simply not exist on S3; no need to worry about partial uploads. And zips should make our Zarr-creation code faster and more efficient (it can just grab 1 zip per NWP variable, instead of looping round 93 little gribs).
- `rclone` currently takes 7 seconds to figure out that a single-level directory already has all the files it needs. That's 10 minutes per NWP run. This is an issue because the VM gets pre-empted. We could implement our own code to list FTP and S3 once, and figure out which files need to be uploaded.
- run `rclone` as a daemon (in "remote control" mode) for the duration of the run, so it doesn't re-establish its FTP clients for every directory, and so it doesn't re-run FTP LIST for every directory. (Although Gemini says that running rclone as a daemon won't cache the file listings; only running rclone as a mount will cache the file listing)

## Questions

- Can we move the FTP transfer job to `eu-central-1`?
- Are you happy with my proposed plan for splitting the transfers into two stages?

## Related

- #390 - my raw notes from my research
- #370 
- #360 
- #258 
- #173

JackKelly:
This PR _should_ be ready for testing in production. I've tested it on an EC2 VM, copying to an S3 bucket, and it was getting 50 MiB/s, and it only takes about 30 secs to check an entire NWP run :slightly_smiling_face: 

## Aims
- Speed up every part of the file transfer from DWD to Source Co-Op.
- Provide a solution that can handle DWD's planned directory format.

## Implementation notes
The approach in this PR differs from the previous approach (PR #360 & #370) in that this PR:
- **Downloads files from DWD's HTTPS server instead of the FTP server**.
    - DWD's FTP server doesn't allow more than 10 concurrent connections from one IP address. This lack of concurrency was killing transfer performance given the ~150ms latency over the 8,000 km from Germany to `us-west-2`, and given the tiny size of the grib files, and given the "chatty" nature of the FTP protocol. FTP requires about 7 round trips per file, which, over 8,000 km, is 1 second of overhead per file, even though the actual payload might only take a fraction of a second to transfer.  In contrast, DWD's HTTPS server doesn't appear to have an upper limit on concurrent connections. And HTTP is less "chatty" than FTP: once the connection is established, HTTPS only needs 1 back-and-forth per file.
- **Uses `rclone copyurl --urls src_and_dst_paths.csv` instead of `rclone copy`**.
    - [`rclone copyurl`](https://rclone.org/commands/rclone_copyurl/) is a gem hidden deep within `rclone`'s docs, and I only discovered it by chance a few days ago (and the docs don't make it clear that `copyurl` is capable of solving our problems, to the extent that Gemini firmly believes that `copyurl` cannot do what this PR does with `copyurl`! (UPDATE: I've submitted a [PR to `rclone` to update the docs](https://github.com/rclone/rclone/pull/9141))). AFAICT, `copyurl` is the only `rclone` command that allows us to provide a list of mappings from full source URL to full destination path. In contrast, `rclone copy --files-from` only allows us to provide a list of _source_ files; `rclone copy` can't modify the structure of the destination path (e.g. extracting the NWP init datetime from the base filename of the grib, and using that datetime in an early part of the destination path). `rclone copyurl` is wonderful because it allows us to provide a list of arbitrary mappings from source URL to destination path.
- **Uses Python to figures out which files have already been transferred**
    - The previous design relied on `rclone copy` to figure out which files already exist on the destination. This was slow.
    - In contrast `rclone copyurl` doesn't check the destination before copying. It just blindly copies.
    - So, the Python code in this PR first asks `rclone` to list all the files on the HTTPS server, and list the relevant files in the destination path, and then our Python code computes the set difference. So we can guarantee that this check only happens once per run. And it's fast. It only takes about 30 seconds to check an entire NWP run, whereas the old approach took several minutes to check the larger NWP variable directories.
- **Simplified logging**
    - Instead of writing fiddly Python code to parse JSON logs from `rclone`, this PR instead takes a simpler approach: We just persuade `rclone` to output useful logs as strings, and log those strings. No more JSON parsing; and no more adding up `TransferSummary` objects. We still get a regular log saying how much data has been transferred, and the throughput. We do lose a _tiny_ bit of functionality, in that the current version doesn't say how much is _left_ to transfer. We can add that back in if needs be.

## TODO
- [x] Put into a `src/reformatters/dwd/archive_gribs` folder, with an `__init__.py`.
- [x] Split into several Python files.
- [x] Benchmark on an OCF AWS EC2 VM, copying from DWD HTTPS to S3. DONE. It's fast! (~50 MiB/s). Logs [here]([archive-grib-files.log](https://gist.github.com/JackKelly/685a319adc01189d1905f8a402572ac1)).
- [x] Write unit tests

## Related

- This PR closes: #392 
- My raw research notes: #390
- [More research notes](https://docs.google.com/document/d/1NL-mixEJe1aR3gnoexiYXEoLSBah77Y8BPBgaLG1w68/edit?tab=t.0) (in Google docs)
- Previous approach PR 2/2: #370 
- Previous approach PR 1/2: #360 
- #258 
- #173


aldenks:
:tada: Will review soon!

JackKelly:
Thank you! No huge rush - I'll be busy this week with other stuff. And sorry for this being the _third_ re-write of this functionality! "Third time's a charm", right?! I'm feeling optimistic that this approach is the simplest yet :slightly_smiling_face:

src/reformatters/dwd/archive_gribs/copy_files_from_dwd.py
```diff
@@ -0,0 +1,171 @@
...
+    1. The full source path, e.g. `https://opendata.dwd.de/.../filename.grib2.bz2`.
+    2. The destination path, from the NWP init datetime onwards.
+
+    This is the format required by `rclone copyurls`.
+    """
+    csv_of_files_to_transfer: list[str] = []  # Each list item is one line of the CSV.
+    for src_path in src_paths_starting_with_nwp_var:
+        dst_path = convert_src_path_to_dst_path(src_path)
+        if dst_path not in files_already_on_dst:
+            full_src_path = f"{src_host_and_root_path}/{src_path}"
```
JackKelly, line 168: I'd be tempted to use `urlpath.URL` here, to ensure we join URLs correctly, if you're happy for me to add [`urlpath`](https://pypi.org/project/urlpath/) as a dependency?

aldenks: would the benefit be to make sure we don't have double slashes? i think id rather ensure that with a unit test or assert since we don't have many cases to deal with here if that's the specific benefit, but maybe there's more?

src/reformatters/dwd/archive_gribs/list_files.py
```diff
@@ -0,0 +1,163 @@
...
+        if (
+            e.returncode == 3
+            and isinstance(e.stderr, str)
+            and "directory not found" in e.stderr.lower()
+        ):
+            log.info("Directory not found: '%s'", path)
+            return []
+        else:
+            _log_error_from_called_process_error(e)
+            raise
```
aldenks, line 101: what about a

try:
   _log_err...
finally:
  raise

to make sure the nice logging doesn't get in the way of raising?

src/reformatters/dwd/archive_gribs/path_conversion.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class DateExtractionError(Exception):
+    pass
+
+
+def extract_nwp_init_datetime_from_grib_filename(
+    grib_filename: str,
+) -> datetime:
+    dwd_nwp_init_date_regex: Final[re.Pattern[str]] = re.compile(r"_(\d{10})(?=_)")
```
aldenks, line 47: its probably not material perf wise, but the .compile is happening on each call. We could put in a constant

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
```diff
@@ -69,22 +68,32 @@ def archive_grib_files(
         self,
         # It would've made more sense for `dst_root` to be a `PurePosixPath` but Typer doesn't
         # handle `PurePosixPath`, so we use a `str` to keep Typer happy.
-        dst_root: str = grib_archive_path,
+        dst_root_path: str = grib_archive_path,
```
aldenks, line 71: Will you make the grib_archive_path `Final`? and maybe it should be called dynamical_grib_archive_rclone_root to make it clear its our archive and rclone specific

aldenks (approved):
Awesome! Thanks for the iterations on this -- "easy" things often aren't so easy at some scale 🙃 

If you want to merge and do follow ups in a separate PR thats good too, they are all minor. I'll let you know when i can get this up and testing, not sure exactly but in the next week for sure.

aldenks:
Actually, @JackKelly im going to merge this so i can wrap up the `ty` switch

JackKelly:
Awesome, thanks! I'll implement your suggestions in a follow-up PR!

src/reformatters/dwd/archive_gribs/copy_files_from_dwd.py
JackKelly: Good idea! I've implemented some checks within the body of `compute_which_files_still_need_to_be_transferred` in eca70f78ae9e0cf8451d14cb6db63f448724d13c

src/reformatters/dwd/archive_gribs/list_files.py
JackKelly: Implemented in 241d0a1

src/reformatters/dwd/archive_gribs/path_conversion.py
JackKelly: Implemented in 4eef095

src/reformatters/dwd/icon_eu/forecast/dynamical_dataset.py
JackKelly: Oops, good catch about me forgetting to make it `Final`! And thanks for the naming suggestion - I agree! Implemented in a9a4792

PR #425, opened 2026-02-06
Add src/scripts/check_resource_provisioning.py

aldenks:
Idea: over provision initially to avoid guess and check for resource requests, then run this regularly to tune down based on actual usage.


aldenks:
we should do this, and im not right now. will resurrect when ready

PR #432, opened 2026-02-10
DWD archive grib unsuspend, adjust resources, and add internal retries

aldenks:
## Resources
I saw cpu usage of ~0.4 and memory of about 1GB (respect, rclone)

It took about 2h17 mins to do 4 inits. that was with 2 pod restarts. leaving pod timeout at 3h, although it probably could be a little lower.

I'm not sure if the http availability timing is the same as ftp but im assuming it is and left the schedule the same.

## Retries
I did see one timeout of the 120s limit on listing files on DWD https. I'm inclined to leave that timeout as is and add the retry.

### DIFF
--- a/src/reformatters/dwd/archive_gribs/copy_files_from_dwd.py
+++ b/src/reformatters/dwd/archive_gribs/copy_files_from_dwd.py
@@ -53,6 +53,7 @@ from pathlib import PurePosixPath
 from typing import Any
 
 from reformatters.common.logging import get_logger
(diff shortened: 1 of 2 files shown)


aldenks:
@JackKelly im going to merge this so the job is unuspended and we start building our uninterrupted archive! but would love a retroactive review

JackKelly (commented):
Looks good to me! It's very exciting to see this up-and-running! Thanks for all your work on this :)

PR #510, opened 2026-03-14
Add virtual zarr + icechunk prototype

### LINKED ISSUE #509: Virtual datasets exploration
We want 
- very low latency updates
- efficient map oriented access
- complete variables + levels for datasets which already have public grib2s or similar

aldenks:
Explores creating virtual zarr / icechunk datasets from gribs

### DIFF
--- a/prototypes/virtual_grib_gribberish.py
+++ b/prototypes/virtual_grib_gribberish.py
@@ -1,0 +1,184 @@
+"""
+Prototype: Direct GRIB virtualization using gribberish + VirtualiZarr + Icechunk.
+
+Unlike the NetCDF4-based prototype, this virtualizes real NOAA GFS GRIB2 files
+*directly* using gribberish's zarr v3 codec (GribberishCodec). No format
+conversion needed.
+
+Pipeline:
+  1. gribberish scans GRIB file → byte offsets per message
+  2. VirtualiZarr ManifestArray built with GribberishCodec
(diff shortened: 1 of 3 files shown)


aldenks:
This has served its exploration purpose

aldenks:
This looks quite possible with icechunk + gribberish codec.

Virtualizarr helps automate some of the creating the structure of a virtual dataset and gribberish can scan gribs for us, but in our cases its probably simpler to not use virtualizarr and only use GribberishCodec as the read time codec (not needed at write time).

As long as we
- have a template -- for consistency, use our existing dataset templates with specific changes to encoding (use GribbershCodec vs our usual zstd) and chunk updates (no shards, chunk shape == grib shape)
- can read the byte offset and length from a grib index
- can calculate the chunk coords -- trivial if you have the template dataset with coordinates labels and can parse the times from the url which we can in all cases so far (if coords are not in the url, you can download and read the grib to get them from internal metadata)

we're good:

```
# After setting up standard icechunk zarr repo and writing metadata, set virtual refs on the store

for grib_url, grib_index_url in urls:
    for var in vars:
      init_time_chunk_i, lead_time_chunk_i = parse_chunk_coords(grib_url, template_ds)
      offset, length = parse_grib_index(grib_index_url, var)
      store.set_virtual_ref(
          f"{var.name}/c/{init_time_chunk_i}/{lead_time_chunk_i}/0/0",
          grib_url,
          offset=offset,
          length=length,
      )
```

PR #584, opened 2026-04-28
Use amend instead of commit for intermediate writes on temp branches

### LINKED ISSUE #567: Use icechunk amend within a backfill/update to reduce snapshot counts
Icechunk 2 adds an `amend` operation we can use instead of `commit` to record partial progress within a distributed backfill or update job. Those intermediate snapshots were always done in a backfill before a dataset is released or on a temp branch and never intended to be visible to end users, amend will actually keep them out of the history.

aldenks:
## Summary
This PR optimizes the temporary branch workflow for parallel writes by using icechunk's `amend` operation for intermediate writes instead of `commit`. This collapses per-worker and intermediate metadata writes into a single snapshot in the main branch's history, rather than creating separate snapshots for each write.

## Key Changes

- **New `amend_if_icechunk()` function** in `storage.py`: Mirrors `commit_if_icechunk()` but uses `amend` to replace the branch tip instead of appending. Includes automatic rebase-on-conflict handling since icechunk's amend doesn't accept a `rebase_with` argument.

- **Updated parallel coordination flow** in `parallel_coordination.py`:
  - Initial setup still uses `commit_if_icechunk()` to preserve the original main snapshot as the parent
  - Finalization now uses `amend` (with manual conflict handling) instead of `commit` to fold final metadata into the existing tip

- **Worker chunk writes** in `dynamical_dataset.py`: Changed from `commit_if_icechunk()` to `amend_if_icechunk()` so per-worker writes collapse into a single snapshot

- **Documentation updates** in `parallel_processing.md`: Clarified the three-phase workflow and explained why each phase uses commit vs. amend, plus the net effect on main's history

## Implementation Details

The `amend_if_icechunk()` function handles `ConflictError` by manually rebasing before retrying the amend, since icechunk's amend API doesn't support the `rebase_with` parameter that commit does. This maintains the same retry semantics (max 10 attempts) as the existing commit function.

The workflow now ensures that a backfill or operational update produces exactly one new snapshot in main's history, regardless of how many workers participate or how many intermediate metadata writes occur.

https://claude.ai/code/session_01Evsd8LQV37jLSF5e7Z4EGR


aldenks:
Needs a close review from myself

src/reformatters/common/parallel_coordination.py
```diff
@@ -185,9 +189,13 @@ def finalize(
...
-            new_snapshot = session.commit(
-                commit_message, rebase_with=icechunk.ConflictDetector()
-            )
+            # Amend folds the final metadata into the existing temp-branch tip
+            # so it does not become an extra snapshot in main's history.
+            try:
+                new_snapshot = session.amend(commit_message)
+            except icechunk.ConflictError:
+                session.rebase(icechunk.ConflictDetector())
+                new_snapshot = session.amend(commit_message)
```
aldenks: can this use amend_if_icechunk? otherwise, lets add retrying like that function does -- we're limited to one retry here but if we've gotten this far we'd like to rebase and try again up to 10 times

docs/parallel_processing.md
```diff
@@ -26,9 +26,11 @@ For backfills, metadata is written before workers start (the dataset is being cr
...
+2. **All workers** — open sessions on the temp branch, write chunk data, **amend** the temp branch tip (via `amend_if_icechunk`, which falls back to `rebase` + `amend` on `ConflictError`). Each amend replaces the previous tip, so per-worker writes collapse into a single snapshot rather than each becoming its own snapshot.
+3. **Last worker finalization** — if `update_template_with_results` produced a template that differs from the input (e.g. some reads failed and the time dim was trimmed), writes the new metadata on the branch and **amends** it into the existing tip (via `amend_if_icechunk`). If the template is unchanged, the temp tip already has the right metadata, so the rewrite + amend is skipped. Either way, finalize then atomically resets `main` to the temp branch tip using `reset_branch("main", snapshot, from_snapshot_id=original)`. This branch reset is what makes all writes visible to readers. The `from_snapshot_id` check ensures no concurrent process moved main.
```
aldenks:
```suggestion
3. **Last worker finalization** — if `update_template_with_results` produced a template that differs from the input (e.g. the time dim was trimmed or a coordinate or attribute was updated), writes the new metadata on the branch and **amends** it into the existing tip (via `amend_if_icechunk`). If the template is unchanged, the temp tip already has the right metadata, so the rewrite + amend is skipped. Either way, finalize then atomically resets `main` to the temp branch tip using `reset_branch("main", snapshot, from_snapshot_id=original)`. This branch reset is what makes all writes visible to readers. The `from_snapshot_id` check ensures no concurrent process moved main.
```

src/reformatters/common/parallel_coordination.py
```diff
@@ -82,6 +83,10 @@ def parallel_setup(
             ]
             for ic_store in ic_stores:
                 copy_zarr_metadata(template_ds, tmp_store, ic_store)
+            # Must be commit, not amend: this is the first write on the fresh
+            # temp branch, and amend would replace the original main snapshot
+            # in the branch's view of history. Subsequent writes (worker chunks
+            # and finalize) use amend so they collapse into this snapshot.
```
aldenks:
```suggestion
            # Must be commit, not amend: this is the first write of the update, becoming the snapshot all future changes in this update are amended into
```

src/reformatters/common/parallel_coordination.py
```diff
@@ -164,10 +169,17 @@ def finalize(
     else:
         updated_template = template_ds
 
+    # If update_template_with_results didn't actually change anything (backfill,
+    # or operational update with no read failures), the temp branch tip already
+    # has the right metadata from setup. Skip the redundant copy + amend and
+    # just reset main to the existing tip. This keeps a no-op finalize from
+    # adding a manifest-only "rewrite" snapshot.
```
aldenks: remove whole comment, this is documented and code is clear already
```suggestion
```

src/reformatters/common/parallel_coordination.py
```diff
@@ -181,13 +193,27 @@ def finalize(
...
+                session = repo.writable_session(branch_name)
+                copy_zarr_metadata(
+                    updated_template, tmp_store, session.store, icechunk_only=True
+                )
+
+                # Amend folds the final metadata into the existing temp-branch tip
+                # so it does not become an extra snapshot in main's history.
+                # Inlined rather than via amend_if_icechunk because we already
+                # iterate one repo at a time; amend_if_icechunk's primary/
+                # replicas split would be misleading here.
```
aldenks: remove comment

src/reformatters/common/parallel_coordination.py
```diff
@@ -181,13 +193,27 @@ def finalize(
...
+                def _amend(s: icechunk.Session = session) -> str:
+                    try:
+                        return s.amend(commit_message)
+                    except icechunk.ConflictError:
+                        s.rebase(icechunk.ConflictDetector())
+                        return s.amend(commit_message)
+
+                new_snapshot = retry(_amend, max_attempts=10)
             repo.reset_branch("main", new_snapshot, from_snapshot_id=original_snapshot)
```
aldenks, line 219:
```suggestion
            # Make our update visible to readers of main.
            # from_snapshot_id=original_snapshot ensures we don't overwrite another uncoordinated update
            repo.reset_branch("main", new_snapshot, from_snapshot_id=original_snapshot)
```

src/reformatters/common/parallel_coordination.py
```diff
@@ -53,41 +54,53 @@ def parallel_setup(
...
-        # Zarr v3: do NOT expand (readers would see empty holes)
+
+        setup_info: SetupInfo
+        if existing_setup:
+            # ready.json exists, meaning a prior worker 0 attempt completed
+            # setup (branch + "Expand dataset" commit + ready.json) on every
+            # icechunk repo. Redoing the metadata copy + commit here would
+            # stack a second "Expand dataset" snapshot on the temp branch
+            # that subsequent worker amends keep as their parent, leaking
+            # an extra snapshot into main's ancestry at finalize.
```
aldenks:
```suggestion
            # ready.json exists, meaning a prior worker 0 attempt completed
            # setup (branch + "Expand dataset" commit + ready.json) on every icechunk repo.
```

PR #625, opened 2026-05-29
Fix/silence harmless warnings

aldenks:
## Summary
This PR silences open_zarr efficiency warnings during validation of Icechunk stores by using consolidated=False. Icechunk ignores consolidated metadata, so we now conditionally pass `consolidated=False` when opening Icechunk stores to avoid unnecessary warnings and fallback behavior.

## Key Changes

- **validation.py**: Added logic to detect Icechunk stores and pass `consolidated=False` to `xr.open_zarr()` since Icechunk ignores consolidated metadata. Zarr v3 stores always carry consolidated metadata, so we require it there.

- **dynamical_dataset.py**: Updated the replica store opening in `validate_dataset()` to use the same consolidated metadata handling as the primary validation path, ensuring consistent behavior across replica and primary stores.

- **validation_test.py**: Updated test to write consolidated metadata (matching production stores) and suppress the expected Zarr format 3 spec-compatibility warning that production also suppresses.

- **dynamical_dataset_test.py**: Updated mock to accept the new `consolidated` parameter in `xr.open_zarr()` calls.

- **region_job_test.py (GEFS analysis & forecast)**: Added `@pytest.mark.filterwarnings` decorators to suppress expected `NotGeoreferencedWarning` from rasterio that leaks under ThreadPoolExecutor (warnings.catch_warnings is not thread-safe).

## Implementation Details

The key insight is that Icechunk and Zarr v3 handle consolidated metadata differently:
- Icechunk ignores consolidated metadata and always reads unconsolidated, triggering a warning if `consolidated=True` is passed
- Zarr v3 stores always carry consolidated metadata and should be opened with `consolidated=True`

By checking `isinstance(store, IcechunkStore)`, we can conditionally set the appropriate value and avoid spurious warnings during validation.

https://claude.ai/code/session_01JGZrrRJyKu8KUK2Ri4VrAn

### DIFF
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -499,7 +499,6 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
                         xr.open_zarr(
                             replica_store,
                             chunks=None,
-                            # Icechunk ignores consolidated metadata; see validation.validate_dataset.
                             consolidated=not isinstance(replica_store, IcechunkStore),
                         ),
                     )
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -14,6 +14,7 @@ import sentry_sdk
 import sentry_sdk.crons
 import typer
 import xarray as xr
+from icechunk.store import IcechunkStore
 from pydantic import Field, computed_field
 
 from reformatters.common import (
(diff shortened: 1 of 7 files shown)


src/reformatters/common/validation.py
```diff
@@ -63,10 +64,15 @@ def validate_dataset(
     """
     log.info(f"Validating zarr {store}")
 
+    # Icechunk ignores consolidated metadata, so reading it always falls back to
+    # the (slower) unconsolidated path with a warning; pass consolidated=False.
+    # Our zarr3 stores always carry consolidated metadata, so require it there.
```
aldenks: remove comment

src/reformatters/common/dynamical_dataset.py
```diff
@@ -495,7 +496,12 @@ def validate_dataset(
                     partial(
                         validation.compare_replica_and_primary,
                         self.template_config.append_dim,
-                        xr.open_zarr(replica_store, chunks=None),
+                        xr.open_zarr(
+                            replica_store,
+                            chunks=None,
+                            # Icechunk ignores consolidated metadata; see validation.validate_dataset.
```
aldenks: remove comment

PR #632, opened 2026-05-31
Fix zero-filled easternmost longitude column in GEFS a/b GRIB upsampling

aldenks:
## Problem

When upsampling 0.5°/1.0° GEFS a/b GRIB data to the 0.25° grid, `rasterio.warp.reproject` (bilinear) has no longitude wraparound. The easternmost destination column (longitude 179.75°) falls just outside the source grid's eastern edge — the source has no column for longitude 180° — so `reproject` fills those pixels with `0.0`. That value coincides with the encoding `fill_value=0`, so the column reads back as a zeroed strip.

This affects every variable sourced from a/b files at lead times **> 240h** (where the higher-res s-files no longer apply), on every latitude row, for ~99/179 lead times of each 00Z init. For physically-positive variables (pressure, RH, longwave radiation, precipitable water, geopotential heights) it shows up directly as `val_spatial_min = 0`; for signed variables it's hidden below the real spatial min.

## Fix

`read_data._reproject_bilinear_longitude_wrap` pads the global source grid with a wrap-around column on each side (longitude 180° == −180°) and shifts the source transform west by one source pixel, so every destination pixel falls strictly inside the source extent and interpolates correctly across the antimeridian. The approach is longitude-convention-agnostic and asserts the source spans the full globe.

The existing pixel-center alignment invariant (`raw == result[::step, ::step]`) still holds — the wrap columns only affect the seam pixels that don't align with source centers.

## Why this slipped through

The slow integration test asserted only `np.all(np.isfinite(data))`. The buggy column is `0.0`, which is finite, so the check passed despite the corruption. This PR adds a wraparound assertion: the easternmost column must equal the antimeridian interpolation of its neighbors (`(data[..., -2] + data[..., 0]) / 2` for 0.5° source), which fails on the old behavior and passes on the fix.

## Tests

- New fast unit test `tests/noaa/gefs/read_data_test.py` covering 0.5° and 1.0° sources: no NaN, exact center alignment preserved, and correct antimeridian interpolation at the seam.
- Tightened slow test `test_download_and_read_all_vars_current` (lead 252h) with the wraparound assertion above.
- 84 fast GEFS tests pass; ruff/ty clean.

## Note on existing data

This fixes the read/processing path going forward. Data already written to the archive still has the zeroed column and needs a reprocess of the affected init/lead ranges to correct historically (being bundled with other edits separately).

https://claude.ai/code/session_01AP8RhtctgUqDKfyzsMTHtB

---
_Generated by [Claude Code](https://claude.ai/code/session_01AP8RhtctgUqDKfyzsMTHtB)_

### DIFF
--- a/src/reformatters/noaa/gefs/read_data.py
+++ b/src/reformatters/noaa/gefs/read_data.py
@@ -114,14 +114,13 @@ def read_rasterio(
                         assert reader.transform == out_transform
                         assert reader.crs.to_dict() == out_crs
                         return raw
-                    result, _ = rasterio.warp.reproject(
+                    result = _reproject_bilinear_longitude_wrap(
                         raw,
-                        np.full(out_spatial_shape, np.nan, dtype=np.float32),
-                        src_transform=reader.transform,
-                        src_crs=reader.crs,
-                        dst_transform=out_transform,
-                        dst_crs=out_crs,
-                        resampling=rasterio.warp.Resampling.bilinear,
+                        reader.transform,
+                        reader.crs,
+                        out_spatial_shape,
+                        out_transform,
+                        out_crs,
                     )
                     if not Config.is_prod:
                         # Because the pixel centers are aligned we exactly retain the source data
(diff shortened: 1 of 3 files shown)


aldenks:
## Visual gut-check

Downloaded a real a-file (`pressure_surface`, init=2024-01-01T00, lead=252h, member 0 → `gec00.t00z.pgrb2a.0p50.f252`, 0.5°), read the raw band, and ran our exact `_reproject_bilinear_longitude_wrap`. Compared against the old (no-wrap) `reproject` for contrast.

**Counts**

| Array | Shape | Zeros | NaNs | Min (Pa) | Max (Pa) |
|---|---|---|---|---|---|
| Original 0.5° source | (361, 720) | 0 | 0 | 49062.8 | 103939.6 |
| Reprojected — our fix | (721, 1440) | **0** | 0 | 49062.8 | 103939.6 |
| Reprojected — old behavior | (721, 1440) | **721** | 0 | **0.0** | 103939.6 |

The old behavior zeros out exactly the easternmost column (lon 179.75°) — all 721 latitude rows — dropping the field min to 0.0. With the fix the easternmost column has 0 zeros and the min returns to the physical 49062.8 Pa.

**Plots** (reviewed in the working session; can't attach binaries to a GH comment here):
- *Maps* — original 0.5° vs fixed 0.25° are visually identical, continents clearly resolved, no zeroed edge.
- *East-edge zoom (178°–180°)* — old behavior shows a dark zero stripe at 179.75°; the fix is continuous to the edge.
- *Histograms* — identical distributions, both bottoming out at ~49000 Pa with no spike at 0 (the old behavior would add a 721-count spike at 0).

---

PR #636, opened 2026-06-01
Add NASA IMERG Early and Late V07 analysis datasets

### LINKED ISSUE #462: IMERG analysis - early and late
Early, late and final -- probably each in their own dataset because they update at different times and frequencies

aldenks:
## Summary

Adds two global half-hourly precipitation analysis datasets from NASA GPM IMERG V07:

- `nasa-imerg-analysis-early-v7`
- `nasa-imerg-analysis-late-v7`

(IMERG Final is intentionally deferred to V08 — V07 Final stopped 2025-09-30.)

The two runs differ only in their source product/URL and update latency, so they share `NasaImergRegionJob` and a `NasaImergAnalysisTemplateConfig` base under `src/reformatters/nasa/imerg/`, mirroring the NOAA GFS forecast/analysis shared-code layout. A single `NasaImergSourceFileCoord` carries a `run` field; thin per-run `RegionJob` subclasses set it via a `ClassVar`.

## Details

- **Source:** GES DISC HTTPS with NASA Earthdata Login (reuses the SMAP `earthdata_auth` pattern). `download_file` tries the date-appropriate `V07C`/`V07B` filename label and falls back to the other.
- **Reading:** rasterio over the HDF5 `/Grid` subdatasets, reorienting the source `(time, lon, lat)` layout to `(latitude, longitude)`, converting fill values (`-9999.9` float / `-9999` int16) to NaN. `precipitation` is converted mm/hr → `kg m-2 s-1` to match the existing `precipitation_surface` / `precipitation_flux` convention.
- **Variables:** `precipitation_surface`, `probability_of_liquid_precipitation_surface`, `precipitation_quality_index_surface`.
- **Grid / time:** global 0.1° (1800 × 3600), record from **1998-01-01** (verified against the GES DISC archive listing).
- **Chunking:** chunk `(1440, 45, 45)` / shard `(1440, 450, 450)` = 30 days/shard (~37 GB per-worker buffer). The memory ↔ read-locality ↔ update-cost tradeoff and lean (720) / premium (2880) alternatives are documented in `template_config.py`.
- **Storage:** Icechunk on AWS Open Data (`s3://dynamical-nasa-imerg`).
- Added `"0.1 degrees (~10km)"` to the shared `SpatialResolution` literal; added the two non-CF/non-ECMWF variables to the compliance-test exemption lists.

## Testing

- Unit tests for all custom logic (URL building for both runs, V07B/V07C selection, the HDF5 reorientation, fill/scale, coord generation, version-fallback download, operational update jobs, K8s/validators).
- A **fully-mocked `backfill_local` → `update` integration test that runs offline** and exercises the whole download → read → write → append pipeline.
- **Slow tests that download and read a real granule from each run** (Early & Late) — require Earthdata credentials, so they don't run in plain CI.
- `ruff`, `ty`, and the full non-slow suite pass.

## Follow-ups before backfill

- **Verify lat/lon orientation against a real file:** the `(time,lon,lat)` → `(lat,lon)` transpose + latitude flip is implemented from the source spec but only confirmed by the slow tests. Run `uv run pytest tests/nasa -m slow` with Earthdata credentials before backfilling.
- Add the datasets to the `dynamical.org` catalog (separate repo).

Relates to #462


aldenks:
closing in favor of #781

PR #644, opened 2026-06-05
Fix multi-dimensional coordinate summarization in validation

aldenks:
## Summary
Fixed coordinate summarization in dataset validation to properly handle multi-dimensional coordinates (e.g., `valid_time` spanning `init_time × lead_time` dimensions) by flattening them before generating summary strings.

## Changes
- **`src/reformatters/common/validation.py`**: Modified `_summarize_coords()` to call `.ravel()` on coordinate values instead of `np.atleast_1d()`. This ensures multi-dimensional coordinates are flattened to 1D before summarization, producing concise "first..last" range summaries instead of attempting to index and dump entire nested arrays.

- **`tests/common/validation_test.py`**: Added `test_summarize_coords_multidimensional()` test case that verifies:
  - Multi-dimensional coordinates (e.g., `valid_time` over `init_time × lead_time`) are summarized as a range with element count
  - The summary output remains single-line (no newlines from nested array dumps)
  - Other coordinates in the dataset are still properly summarized

## Implementation Details
The change from `np.atleast_1d()` to `.ravel()` handles the case where a coordinate is defined over multiple dimensions. Previously, `np.atleast_1d()` would preserve the multi-dimensional structure, causing the summarization logic to fail when trying to format the nested array. The `.ravel()` method flattens the array to 1D while preserving all values, allowing the existing min/max range logic to work correctly.

https://claude.ai/code/session_01SDuw5TV65SzF3qVBPHefUe

### DIFF
--- a/src/reformatters/common/validation.py
+++ b/src/reformatters/common/validation.py
@@ -260,9 +260,6 @@ def _format_coord_value(value: object) -> str:
 def _summarize_coords(ds: xr.Dataset) -> str:
     parts = []
     for name in ds.coords:
-        # ravel so multi-dimensional coords (e.g. valid_time over init_time x
-        # lead_time) summarize as a concise first..last range instead of
-        # indexing whole sub-arrays and dumping the full nested array.
         values = ds.coords[name].values.ravel()
         if values.size == 0:
             parts.append(f"{name}=<empty>")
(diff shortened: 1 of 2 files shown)


src/reformatters/common/validation.py
```diff
@@ -260,7 +260,10 @@ def _format_coord_value(value: object) -> str:
 def _summarize_coords(ds: xr.Dataset) -> str:
     parts = []
     for name in ds.coords:
-        values = np.atleast_1d(ds.coords[name].values)
+        # ravel so multi-dimensional coords (e.g. valid_time over init_time x
+        # lead_time) summarize as a concise first..last range instead of
+        # indexing whole sub-arrays and dumping the full nested array.
```
aldenks: remove this comment

Issue #513, opened 2026-03-16
Virtual dataset implementation + first dataset


aldenks:
### Motivation

3 fold:
* Low latency (p50 < 5s, p99 < 10s), incremental updates after each source file is made available
* Fast and efficient support for spatial / map / single time step data access patterns
* Allow us to offer all variables + model levels in cases where existing grib files are publicly available (e.g. all NOAA NODD data, ECMWF IFS ENS, our DWD ICON-EU archive)

### Work 

1. Figure out architecture #621 
2. Implement common + first dataset. Mechanics explored and validated in https://github.com/dynamical-org/reformatters/issues/509#issuecomment-4059310744

### Implementation checklist

PR sequence from the [design plan](https://github.com/dynamical-org/reformatters/blob/main/docs/plans/virtual_icechunk_datasets.md#implementation-plan):

- [x] **PR 1** — Extract `MaterializedRegionJob` (mechanical base-class swap; tests pass unchanged) — #643
- [x] **PR 2** — Prep: rename ECMWF index parser to `grib_message_byte_ranges_from_index`, add `Encoding.serializer`, add `gribberish` dependency, add `IcechunkVirtualConfig` + `StoreFactory` wiring + virtual `DynamicalDataset` validator — #647
- [x] **PR 3a** — Worker-processing seam: abstract `RegionJob.process_worker_jobs` + extract `MaterializedRegionJob.process_worker_jobs` (behavior-preserving, materialized only) — #649
- [x] **PR 3b** — `VirtualRegionJob` on the seam (virtual write loop, `process_virtual_refs`, `filter_already_present`, `sync_dims_to`, `chunk_key`) + single-writer operational fork in `update()` + integration tests — #650 _(supersedes #648, now closed)_
- [x] **PR 3c** — Move-only refactor: split `region_job.py` into `virtual_region_job.py` + `materialized_region_job.py` (base `RegionJob` stays) — #653 _(byte-identical move)_
- [x] **PR 3d** — Move `process()` onto `MaterializedRegionJob` (drop the base stub; materialized `process_worker_jobs` narrows its jobs) — #654
- [x] **PR 3e** — Consolidate processing-loop docs (seam, coordination lifecycles, commit cadence) into `docs/`, linked from code — #655
- [x] **PR 4** — First concrete virtual dataset, end to end: `noaa-gefs-forecast-10-day-spatial-dev` (virtual encodings, `*Spatial*` classes, minimal validators) — #656
[...]

PR #648, opened 2026-06-08
VirtualRegionJob + virtual driver fork (virtual datasets PR 3/6)

aldenks:
The virtual write path for [virtual Icechunk datasets](docs/plans/virtual_icechunk_datasets.md). `VirtualRegionJob` writes virtual chunk references (byte ranges into public GRIB files) into icechunk instead of materializing rechunked data. No behavior change for existing materialized datasets.

## What's here

- **`VirtualRegionJob`** (sibling of `MaterializedRegionJob`):
  - `process_virtual` runs the per-batch write loop. Each yielded batch (whole source files) commits on its **own fresh writable session** — a committed icechunk session is read-only, so the per-batch loop reopens a session from the repo.
  - `process_virtual_refs` — abstract generator; yields `VirtualRefBatch`es (one commit's worth of whole files).
  - `filter_already_present` — manifest probe via `store.exists` (no chunk fetch/decode), fanned out concurrently.
  - `sync_dims_to` — grows the append dim by appending the new slice of the already-derived template with `compute=False`, so data vars stay dask-lazy and the decode-only `GribberishCodec` is never invoked. No-op on a pre-sized backfill branch.
  - `chunk_key` — derives chunk indices from the full-size template (so it can't drift from what readers resolve and never re-reads the growing store); shared by filter and emit.
  - Emit via `set_virtual_refs` (explicit `VirtualChunkSpec` per ref — handles the common sparse one-message-per-cell packing; `set_virtual_refs_arr` is a later optimization).
- **Driver fork** in `_process_region_jobs`: virtual + operational → single-writer-to-`main` (no temp branch / `parallel_setup` / `finalize`) for ~5 s per-file visibility; virtual backfill → the existing temp-branch flow, routed through `process_virtual`.
- **`get_jobs` partitioning**: shards if present, else chunks (virtual arrays have no shards). Materialized partitioning is byte-for-byte unchanged.
- **storage**: `_virtual_repository_config_and_credentials` also accepts local-filesystem containers (no credentials) so dev/test virtual datasets can point at local files.

## On the commit guard

The plan called for guarding commits on `session.has_uncommitted_changes` (empty icechunk commits raise). In implementation the generator contract makes that unnecessary: every yield is a non-empty batch and emitting refs always dirties the session, so the commit is never empty; the empty-poll / all-present cases yield no batch and never reach the commit. The loop instead **asserts** `batch.refs` (loud, early failure on a contract violation) and `commit_if_icechunk` stays unconditional. Plan doc updated to match.

## Tests

A synthetic virtual dataset (local-filesystem container, raw-float64 "messages" file, default `BytesCodec`) drives:
- unit tests — `chunk_key` (index mapping, unknown-label → `None`, unaligned-position assert), `_needed_append_dim_size`, empty-batch contract;
- integration — emit → commit → reopen → **value** read-back (catches bad chunk keys *and* byte ranges), filter skips present, `sync_dims_to` grow-then-noop, yield-is-the-commit-unit, two-worker disjoint backfill on a branch, operational single-writer expansion + idempotent second fire;
- a `GribberishCodec` test pins that the serializer threads through expansion without the decode-only codec being invoked.

All of `tests/common` + `tests/example` (1230 tests) and `tests/noaa` pass; `ruff` + `ty` clean.

Part of #513.

### DIFF
--- a/src/reformatters/common/region_job.py
+++ b/src/reformatters/common/region_job.py
@@ -1,6 +1,7 @@
+import asyncio
 import concurrent.futures
 import os
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Iterator, Mapping, Sequence
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from contextlib import suppress
 from copy import deepcopy
@@ -9,13 +10,26 @@ from http import HTTPStatus
 from itertools import chain, pairwise
 from multiprocessing.shared_memory import SharedMemory
 from pathlib import Path
-from typing import Annotated, Any, ClassVar, Generic, Self, TypeVar, get_args
+from typing import (
+    Annotated,
+    Any,
+    ClassVar,
+    Generic,
+    NamedTuple,
+    Self,
+    TypeVar,
+    get_args,
+)
 
 import httpx
+import icechunk
 import numpy as np
 import pandas as pd
 import pydantic
 import xarray as xr
+import zarr
+from icechunk import VirtualChunkSpec
+from icechunk.store import IcechunkStore
 from pydantic import (
     AfterValidator,
     Field,
@@ -24,6 +38,7 @@ from pydantic import (
     field_validator,
 )
 from zarr.abc.store import Store
+from zarr.core.metadata import ArrayV3Metadata
 
 from reformatters.common.binary_rounding import round_float32_inplace
 from reformatters.common.config_models import DataVar
@@ -35,6 +50,7 @@ from reformatters.common.shared_memory_utils import (
     make_shared_buffer,
     write_shards,
 )
+from reformatters.common.storage import commit_if_icechunk
 from reformatters.common.types import (
     AppendDim,
     ArrayND,
@@ -342,8 +358,12 @@ class RegionJob(pydantic.BaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         else:
             data_var_groups = [data_vars]
 
-        # Regions along append dimension
-        regions = dimension_slices(template_ds, append_dim, kind="shards")
+        # Regions along append dimension. Materialized arrays partition by physical
+        # shard; virtual arrays have no shards (one ref per GRIB message), so they
+        # fall back to chunks. dimension_slices stays strict about which key it reads.
+        sample_encoding = next(iter(template_ds.data_vars.values())).encoding
+        kind = "shards" if sample_encoding.get("shards") is not None else "chunks"
+        regions = dimension_slices(template_ds, append_dim, kind=kind)
 
         # Filter regions by time
         if (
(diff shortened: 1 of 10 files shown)


src/reformatters/common/region_job.py
```diff
@@ -342,8 +358,12 @@ def get_jobs(
         else:
             data_var_groups = [data_vars]
 
-        # Regions along append dimension
-        regions = dimension_slices(template_ds, append_dim, kind="shards")
+        # Regions along append dimension. Materialized arrays partition by physical
+        # shard; virtual arrays have no shards (one ref per GRIB message), so they
+        # fall back to chunks. dimension_slices stays strict about which key it reads.
```
```suggestion
        # Regions along append dimension.
        # Materialized arrays partition by shard; virtual arrays use chunks.
```

src/reformatters/common/region_job.py
```diff
@@ -746,6 +766,296 @@ def _cleanup_local_files(
...
+    # icechunk store rather than materializing rechunked data. Base class for
+    # spatial/map-optimized virtual datasets. The driver routes virtual jobs to
+    # process_virtual() (not process()), because each yielded batch is committed
+    # on its own writable session - a committed icechunk session is read-only, so
+    # the per-batch commit loop must reopen a session from the repository.
+
+    # Operational poll deadline. None for backfills, which exhaust their candidate
+    # set and return; set by operational_update_jobs for operational updates, which
+    # poll for newly published files until the deadline approaches.
+    deadline: Timestamp | None = None
```
aldenks: I don't think this gets us much, lets remove the concept.

src/reformatters/common/region_job.py
```diff
@@ -746,6 +766,296 @@ def _cleanup_local_files(
...
+
+class VirtualRegionJob(
+    RegionJob[DATA_VAR, SOURCE_FILE_COORD], Generic[DATA_VAR, SOURCE_FILE_COORD]
+):
+    # Writes virtual chunk references (byte ranges into public GRIB files) into an
+    # icechunk store rather than materializing rechunked data. Base class for
+    # spatial/map-optimized virtual datasets. The driver routes virtual jobs to
+    # process_virtual() (not process()), because each yielded batch is committed
+    # on its own writable session - a committed icechunk session is read-only, so
+    # the per-batch commit loop must reopen a session from the repository.
```
```suggestion
    """Base class for processing a region of virtual Icechunk datasets that point to external files."""
```

src/reformatters/common/region_job.py
```diff
@@ -746,6 +766,296 @@ def _cleanup_local_files(
...
+
+    def process(
+        self,
+        primary_store: Store,
+        replica_stores: list[Store],
+    ) -> Mapping[str, Sequence[SOURCE_FILE_COORD]]:
+        raise NotImplementedError(
+            "VirtualRegionJob is driven via process_virtual(); the dataset driver "
+            "routes virtual jobs there so each batch can commit on its own session."
+        )
```
aldenks: this is gnarly, lets remove `process` from RegionJob and on to MaterializedRegionJob so we can remove thsi

aldenks (commented):
first batch of comments

aldenks:
Superseded by #650 (PR 3b), which reworks this onto the `process_worker_jobs` seam from #649 (PR 3a). The driver fork was redesigned to B-method per the design review; closing in favor of #650.

PR #650, opened 2026-06-09
VirtualRegionJob on the worker-processing seam (virtual datasets PR 3b)

aldenks:
**PR 3b of the Virtual Icechunk Datasets effort** (`Part of #513`). Builds on the seam from #649 (PR 3a). **Supersedes #648**, which I'll close in favor of this.

## What this adds

`VirtualRegionJob` — virtual chunk references (byte ranges into public GRIB files) written into icechunk instead of materialized data — slotted onto 3a's `process_worker_jobs` seam.

### Driver (B-method)
- `VirtualRegionJob.process_worker_jobs` opens the icechunk repos and drives each job's per-batch write loop (`process_virtual`): filter → `process_virtual_refs` batches → **fresh writable session per batch** (a committed icechunk session is read-only) → `sync_dims_to` + emit + commit. Many commits per worker vs the materialized variant's one — same seam, the only difference is batch count.
- `_process_region_jobs` (the temp-branch coordinator) stays **variant-free** — materialized backfill/update and virtual backfill all flow through `process_worker_jobs`.
- The **one** driver fork is in `update()`: a virtual operational update routes to `_run_virtual_operational_update` (single-writer streaming straight to `main`, reusing `process_worker_jobs` on `branch="main"` with no `parallel_setup`/`finalize`), so readers see each source file within seconds.

### Mechanics
- `VirtualRef(data_var, out_loc, location, offset, length)`; `chunk_key` derives indices from the full-size template (shared by filter + emit, so they can't disagree); emit via `set_virtual_refs` (explicit `VirtualChunkSpec`s).
- `filter_already_present`: async `store.exists` manifest probe — no decode.
- `sync_dims_to` sizes each store from its own committed size (replicas commit first, so a partial commit can leave one ahead; appending the primary's slice to it would duplicate positions).
- No commit guard — the generator contract forbids empty yields, so the loop `assert`s `refs` and `commit_if_icechunk` stays unconditional.
- `_processing_region_ds` asserts `get_processing_region() == region` (buffering is meaningless for virtual refs and would make adjacent backfill workers emit into each other's regions).
- `get_jobs` partitions shards-else-chunks (virtual arrays have no shards).
- `StoreFactory.icechunk_repos()` / `all_icechunk_repos(sort=)` split; `_virtual_repository_config_and_credentials` accepts local-filesystem containers (dev/test).
- Validator: a `VirtualRegionJob` `region_job_class` requires a non-null `icechunk_virtual_config`.

## Testing

`tests/common/virtual_region_job_test.py` (new, synthetic local-filesystem-container dataset, raw-float64 messages, default `BytesCodec` → real value read-back): `chunk_key`/sizing units, two-worker disjoint backfill on a branch, operational single-writer expansion + idempotent second fire, per-store `sync_dims_to` after a partial commit, the config validator, the buffered-region guard, and a serializer-threads-through-expansion test. Full non-slow suite green (1291), `ruff` + `ty` clean.

Plan doc (`docs/plans/virtual_icechunk_datasets.md`) refreshed to the B-method design throughout.

### DIFF
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -468,6 +468,12 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         whole source files straight to the "main" icechunk branch as they arrive
         (no parallel_coordination), so readers see each file within seconds."""
         assert workers_total == 1, "Virtual operational updates run single-writer"
+        # A single active-window job whose generator polls the union of all
+        # still-missing files; multiple jobs would run sequentially and the first
+        # one's polling could consume the pod's deadline and starve the rest.
+        assert len(all_jobs) == 1, (
+            f"Virtual operational updates run a single active-window job, got {len(all_jobs)}"
+        )
         self.region_job_class.process_worker_jobs(
             all_jobs, self.store_factory, "main", 0
         )
@@ -477,12 +483,17 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         reformat_job_name: Annotated[str, typer.Argument(envvar="JOB_NAME")],
     ) -> None:
         """Validate the dataset, raising an exception if it is invalid."""
+        # Virtual datasets have no shards (one ref per chunk) and intentionally
+        # have missing chunks for partially-published inits, so check_for_expected_shards
+        # does not apply. A manifest-aware virtual validator comes in the validation PR.
+        is_virtual = issubclass(self.region_job_class, VirtualRegionJob)
         with self._monitor(ValidationCronJob, reformat_job_name):
             primary_store = self.store_factory.primary_store()
             primary_store_validators = list(self.validators())
-            primary_store_validators.append(
-                partial(validation.check_for_expected_shards, primary_store)
-            )
+            if not is_virtual:
+                primary_store_validators.append(
+                    partial(validation.check_for_expected_shards, primary_store)
+                )
 
             validation.validate_dataset(
                 primary_store,
@@ -492,9 +503,10 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
 
             for replica_store in self.store_factory.replica_stores():
                 replica_store_validators = list(self.validators())
-                replica_store_validators.append(
-                    partial(validation.check_for_expected_shards, replica_store)
-                )
+                if not is_virtual:
+                    replica_store_validators.append(
+                        partial(validation.check_for_expected_shards, replica_store)
+                    )
                 replica_store_validators.append(
                     partial(  # ty: ignore[invalid-argument-type]
                         validation.compare_replica_and_primary,
@@ -622,16 +634,25 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
 
     @model_validator(mode="after")
     def _validate_virtual_storage(self) -> Self:
-        # A virtual region job emits chunk refs into icechunk and needs the source
-        # containers registered, so the virtual config is mandatory for it.
-        if (
-            issubclass(self.region_job_class, VirtualRegionJob)
-            and self.icechunk_virtual_config is None
-        ):
-            raise ValueError(
-                f"{self.region_job_class.__name__} is a VirtualRegionJob but no "
-                "icechunk_virtual_config was provided; virtual datasets require it."
-            )
+        if issubclass(self.region_job_class, VirtualRegionJob):
+            # A virtual region job emits chunk refs into icechunk and needs the
+            # source containers registered, so the virtual config is mandatory.
+            if self.icechunk_virtual_config is None:
+                raise ValueError(
+                    f"{self.region_job_class.__name__} is a VirtualRegionJob but no "
+                    "icechunk_virtual_config was provided; virtual datasets require it."
+                )
+            # max_vars_per_job splits a source file's variables across separate jobs,
+            # each of which commits independently — breaking the one-file-per-commit
+            # atomicity virtual readers rely on. Forbid it; virtual jobs are tiny
+            # (byte-range refs) and parallelize along the append dim, not variables.
+            if self.region_job_class.max_vars_per_job is not None:
+                raise ValueError(
+                    f"{self.region_job_class.__name__} sets max_vars_per_job="
+                    f"{self.region_job_class.max_vars_per_job}, but virtual region "
+                    "jobs must keep each source file's variables in one job for "
+                    "per-file commit atomicity; leave max_vars_per_job unset (None)."
+                )
         # Virtual datasets store icechunk metadata + virtual chunk refs, so every store must be icechunk.
         if self.icechunk_virtual_config is not None:
             non_icechunk = [
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -66,46 +66,6 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
     replica_storage_configs: Sequence[StorageConfig] = Field(default_factory=tuple)
     icechunk_virtual_config: IcechunkVirtualConfig | None = None
 
-    @model_validator(mode="after")
-    def _validate_virtual_storage(self) -> Self:
-        # A virtual region job emits chunk refs into icechunk and needs the source
-        # containers registered, so the virtual config is mandatory for it.
-        if (
-            issubclass(self.region_job_class, VirtualRegionJob)
-            and self.icechunk_virtual_config is None
-        ):
-            raise ValueError(
-                f"{self.region_job_class.__name__} is a VirtualRegionJob but no "
-                "icechunk_virtual_config was provided; virtual datasets require it."
-            )
-        # Virtual datasets store icechunk metadata + virtual chunk refs, so every store must be icechunk.
-        if self.icechunk_virtual_config is not None:
-            non_icechunk = [
-                config
-                for config in (
-                    self.primary_storage_config,
-                    *self.replica_storage_configs,
-                )
-                if config.format != DatasetFormat.ICECHUNK
-            ]
-            if non_icechunk:
-                raise ValueError(
-                    "icechunk_virtual_config requires every storage config to use "
-                    f"the ICECHUNK format, but found: {non_icechunk}"
-                )
-        return self
-
-    @computed_field
-    @property
-    def store_factory(self) -> StoreFactory:
-        return StoreFactory(
-            primary_storage_config=self.primary_storage_config,
-            replica_storage_configs=self.replica_storage_configs,
-            dataset_id=self.dataset_id,
-            template_config_version=self.template_config.version,
-            icechunk_virtual_config=self.icechunk_virtual_config,
-        )
-
     def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
         """
         Return the kubernetes cron job definitions to operationally
@@ -170,6 +130,17 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         """Generate and persist the dataset template using the template_config."""
         self.template_config.update_template()
 
+    @computed_field
+    @property
+    def store_factory(self) -> StoreFactory:
+        return StoreFactory(
+            primary_storage_config=self.primary_storage_config,
+            replica_storage_configs=self.replica_storage_configs,
+            dataset_id=self.dataset_id,
+            template_config_version=self.template_config.version,
+            icechunk_virtual_config=self.icechunk_virtual_config,
+        )
+
     def num_variable_groups(self) -> int:
         """Number of variable groups for parallel updates."""
         return self.region_job_class.num_variable_groups(self.template_config.data_vars)
@@ -202,10 +173,7 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
             )
 
             if issubclass(self.region_job_class, VirtualRegionJob):
-                # Virtual operational updates are single-writer streaming (whole
-                # files committed straight to main as they arrive) — a different
-                # lifecycle from the parallel temp-branch coordinator, so they
-                # don't go through _process_region_jobs.
+                # Virtual operational updates are single-writer streaming (see docs/parallel_processing.md)
                 self._run_virtual_operational_update(all_jobs, workers_total)
             else:
                 self._process_region_jobs(
@@ -496,10 +464,9 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
         all_jobs: Sequence[RegionJob[DATA_VAR, SOURCE_FILE_COORD]],
         workers_total: int,
     ) -> None:
-        """Single-writer virtual operational update: commit whole source files
-        straight to "main" as they arrive (no temp branch, no parallel_setup, no
-        finalize), so readers see each file within seconds. Reuses the variant's
-        per-batch write loop via process_worker_jobs on branch "main"."""
+        """Single-writer virtual dataset operational update: commit one or more
+        whole source files straight to the "main" icechunk branch as they arrive
+        (no parallel_coordination), so readers see each file within seconds."""
         assert workers_total == 1, "Virtual operational updates run single-writer"
         self.region_job_class.process_worker_jobs(
             all_jobs, self.store_factory, "main", 0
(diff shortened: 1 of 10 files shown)


src/reformatters/common/dynamical_dataset.py
```diff
@@ -190,15 +201,22 @@ def update(
...
-                reformat_job_name=reformat_job_name,
-                template_ds=template_ds,
-                tmp_store=tmp_store,
-                update_template_with_results=True,
-            )
+            if issubclass(self.region_job_class, VirtualRegionJob):
+                # Virtual operational updates are single-writer streaming (whole
+                # files committed straight to main as they arrive) — a different
+                # lifecycle from the parallel temp-branch coordinator, so they
+                # don't go through _process_region_jobs.
```
aldenks: update docs/parallel_coordination.md with this exception and reduce comment here


```suggestion
                # Virtual operational updates are single-writer streaming (see docs/parallel_coordination.md)
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -67,6 +68,16 @@ class DynamicalDataset(FrozenBaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
 
     @model_validator(mode="after")
     def _validate_virtual_storage(self) -> Self:
```
aldenks: move this validator down to the bottom of the class def (we want the important methods up top and this is not that)

src/reformatters/common/dynamical_dataset.py
aldenks: move this down below `def update_template` (it's not something a subclass would override)

src/reformatters/common/dynamical_dataset.py
```diff
@@ -473,6 +491,20 @@ def _process_region_jobs(
...
 
+    def _run_virtual_operational_update(
+        self,
+        all_jobs: Sequence[RegionJob[DATA_VAR, SOURCE_FILE_COORD]],
+        workers_total: int,
+    ) -> None:
+        """Single-writer virtual operational update: commit whole source files
+        straight to "main" as they arrive (no temp branch, no parallel_setup, no
+        finalize), so readers see each file within seconds. Reuses the variant's
+        per-batch write loop via process_worker_jobs on branch "main"."""
```
```suggestion
        """Single-writer virtual dataset operational update: commit one or more 
        whole source files straight to the "main" icechunk branch as they arrive
        (no parallel_coordination), so readers see each file within seconds."""
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,283 @@ def _cleanup_local_files(
                     coord.downloaded_path.unlink()
 
 
+class VirtualRef(NamedTuple):
+    """A single virtual chunk reference: one GRIB message -> one zarr chunk.
+
+    `out_loc` is the output cell the message fills (e.g. {init_time, lead_time});
+    the chunk index it maps to is computed by VirtualRegionJob.chunk_key, so the
+    generator never reimplements zarr's chunk-key math. `data_var` identifies the
+    array the ref belongs to (and supplies the chunk geometry for chunk_key).
```
```suggestion
    """A single virtual chunk reference: one GRIB message -> one zarr chunk.
    `out_loc` is the output cell the message fills (e.g. {init_time, lead_time}).
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+        """Discover available source files among `remaining` and yield their refs.
+
+        Each yield is one commit's worth of refs. The contract is that a yield
+        holds *whole* source files — every ref of a file is emitted in the same
+        batch, never split across yields — so a reader on `main` sees all of a
+        file's data or none of it (see "Reader safety" in the design doc). This is
+        a discipline the generator must keep (and tests check); it is not encoded
+        in the type. The generator owns the batching policy: backfill yields
+        batches of ~N whole files to amortize commit overhead; operational yields
+        ~1 file at a time for per-file visibility.
```
```suggestion
        """Discover available source files among `remaining` and yield their virtual references.

        Each yield is one commit's worth of refs. The contract is that a yield
        holds *whole* source files — every ref of a file is emitted in the same
        batch, never split across yields — so a reader on `main` sees all of a
        file's data or none of it. This is a discipline the generator must keep (and tests check).
        The generator owns the batching policy: backfill yields batches of ~N whole
        files to amortize commit overhead; operational yields ~1 file at a time for per-file visibility.
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+        store_factory: storage.StoreFactory,
+        branch_name: str,
+        worker_index: int,  # noqa: ARG003 - virtual jobs commit per batch with their own messages
+    ) -> dict[str, list[SourceFileResult]]:
+        """Drive each job's per-batch virtual write loop on ``branch_name``.
+
+        Returns no results: virtual refs live in the icechunk manifest, so there is
+        nothing to thread back to finalize. Many commits per worker (one per yielded
+        batch) vs the materialized variant's single commit — same seam, the only
+        difference is batch count.
```
```suggestion
        """Drive each job's per-batch virtual write loop on ``branch_name``.

        Always returns an empty dict of results: virtual refs live in the icechunk manifest,
        so there is nothing to thread back to finalize.
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+        replica_repos: Sequence[icechunk.Repository],
+        branch: str,
+    ) -> None:
+        """Run the virtual write loop, committing each yielded batch atomically.
+
+        Same loop for backfill and operational; the only difference is the
+        `branch` the driver opened (a temp branch for backfill, "main" for the
+        single-writer operational update). sync_dims_to grows the store lazily (a
+        no-op on the pre-sized backfill branch). A fresh writable session is opened
+        per batch because committing an icechunk session makes it read-only.
```
```suggestion
        per batch because an icechunk session becomes read-only after commit.
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+            needed_size = self._needed_append_dim_size(refs)
+            if needed_size > current_size:
+                self.sync_dims_to(stores, needed_size)
+                current_size = needed_size
+
+            self._emit_refs(stores, refs)
+
+            now = pd.Timestamp.now(tz="UTC")
+            storage.commit_if_icechunk(
+                f"Virtual refs at {now.strftime('%Y-%m-%dT%H:%M:%SZ')}",
```
```suggestion
                f"Update at {now.strftime('%Y-%m-%dT%H:%M:%SZ')}",
```

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+            else:
+                position = 0  # dims absent from out_loc (e.g. lat/lon) are one chunk
+            chunk_index, remainder = divmod(position, chunk_size)
+            assert remainder == 0, (
+                f"{dim}={labels.get(dim)} maps to position {position}, which is not "
+                f"on a chunk boundary (chunk size {chunk_size}); a virtual chunk must "
+                "be exactly one GRIB message."
+            )
+            index.append(chunk_index)
+        return tuple(index)
```
aldenks, line 1063: Take another look at this method and see if it can be simplified, particularly by using more existing machinery inside the libraries we are emulating

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+        """Grow the append dim (and its dependent coords) to `needed_append_dim_size`.
+
+        Writes only the new positions by appending the corresponding slice of the
+        already-derived template; existing positions are untouched. Data vars stay
+        dask-lazy under compute=False, so the decode-only serializer is never
+        invoked. A no-op for any store already covering the size (so backfill's
+        pre-sized branch never resizes, and parallel workers never conflict). Each
+        store's missing slice is computed from its own committed size: replicas
+        commit before the primary, so a partial commit can leave a replica ahead,
+        and appending the primary's slice to it would duplicate positions.
```
aldenks: make this part a comment and leave only the first line ("Grow the append ...") in the docstring

src/reformatters/common/region_job.py
```diff
@@ -803,6 +821,285 @@ def _cleanup_local_files(
...
+    def chunk_key(
+        self, out_loc: Mapping[Dim, CoordinateValue], var: DataVar[Any]
+    ) -> tuple[int, ...] | None:
+        """Map a source message's coordinate labels to its zarr chunk index.
+
+        Geometry (dim order, chunk sizes, coord positions) is read from the
+        full-size template, so the index matches what readers resolve and does not
+        depend on how far the store has been expanded. Shared by the filter and the
+        emitter so they cannot disagree. Returns None if a label is not in the
+        template's coords, which the filter treats as "remaining".
```
aldenks: Move the first 2 sentences of this paragraph into a comment just below the docstring

src/reformatters/common/storage.py
```diff
@@ -139,6 +139,17 @@ def all_stores_exist(self) -> bool:
         return True
 
     def icechunk_repos(
+        self,
+    ) -> tuple[icechunk.Repository, tuple[icechunk.Repository, ...]]:
+        """Returns (primary_repo, (replica_repos, ...)) for the common case where
+        the caller just wants the primary and its replicas. Use all_icechunk_repos
+        when you also need roles or a specific primary-first/last ordering."""
+        repos = self.all_icechunk_repos(sort="primary-first")
+        primary = next(repo for role, repo in repos if role == "primary")
```
aldenks: add assert that only one primary is returned

src/reformatters/common/dynamical_dataset.py
```diff
@@ -245,8 +231,11 @@ def backfill_kubernetes(
                 "Not all stores exist, cannot run with overwrite_existing=True"
             )
             log.info("Writing to existing stores, skipping metadata write.")
-        else:
-            # Write metadata to final store. Required for Zarr v3 only, Icechunk metadata is written in parallel_setup.
+        elif not self._all_stores_icechunk():
+            # Zarr v3 needs metadata in the final store before workers run. Icechunk
+            # gets it from parallel_setup (local tmp store + temp branch), and
+            # pre-writing the full template to main here would publish a NaN-padded
+            # future to readers until finalize resets main.
```
```suggestion
            # Write metadata to final store. Required for Zarr v3 only, Icechunk metadata is written in parallel_setup.
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -331,8 +320,10 @@ def backfill_local(
         )
 
         template_ds = self._get_template(append_dim_end)
-        # Write metadata to final store. Required for Zarr v3 only, Icechunk metadata is written in parallel_setup.
-        template_utils.write_metadata(template_ds, self.store_factory)
+        # Zarr v3 needs metadata in the final store before workers run; icechunk gets
+        # it from parallel_setup (see backfill_kubernetes).
```
```suggestion
        # Write metadata to final store. Required for Zarr v3 only, Icechunk metadata is written in parallel_setup.
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -473,17 +464,42 @@ def _process_region_jobs(
...
+
     def validate_dataset(
         self,
         reformat_job_name: Annotated[str, typer.Argument(envvar="JOB_NAME")],
     ) -> None:
         """Validate the dataset, raising an exception if it is invalid."""
+        # Virtual datasets get their own manifest-aware validators. The materialized
+        # ones don't apply: check_for_expected_shards (no shards; missing chunks for
+        # partially-published inits are expected) and compare_replica_and_primary
+        # (would S3-fetch + GRIB-decode every compared chunk).
```
```suggestion
```

src/reformatters/common/region_job.py
```diff
@@ -343,8 +359,19 @@ def get_jobs(
         else:
             data_var_groups = [data_vars]
 
-        # Regions along append dimension
-        regions = dimension_slices(template_ds, append_dim, kind="shards")
+        # Regions along append dimension.
+        # Materialized arrays partition by shard; virtual arrays use chunks. All
+        # data vars must agree: a mix would partition by whichever var is first,
+        # and a chunk-sized region over a sharded var lets workers write into the
+        # same physical shard.
```
```suggestion
        # Materialized arrays partition by shard; virtual arrays use chunks.
        # All data vars must agree: a mix would partition by whichever var is first,
        # and a chunk-sized region over a sharded var would cause multiple workers
        # to write to the same shard.
```

src/reformatters/common/region_job.py
```diff
@@ -417,7 +444,9 @@ def process(
         Process this region, writing chunk data to primary_store (and replica_stores)
         and returning the per-variable source file coordinates with their final status.
 
-        Implemented by the MaterializedRegionJob and VirtualRegionJob subclasses.
+        Implemented by MaterializedRegionJob. (VirtualRegionJob writes via
+        process_virtual instead; this stub stays on the base so the materialized
+        process_worker_jobs loop, which receives base-typed jobs, type-checks.)
```
aldenks, line 449: add a todo to the plan doc to address `process` being defined on the base but not used by virtual.

src/reformatters/common/region_job.py
```diff
@@ -431,12 +460,16 @@ def process_worker_jobs(
...
         worker_index: int,
     ) -> dict[str, list[SourceFileResult]]:
-        """Process one worker's region jobs against ``branch_name`` and return the
-        per-variable source file results.
-
-        Each variant owns its own store/session lifecycle and commit cadence, so the
-        shared coordinator (``DynamicalDataset._process_region_jobs``) drives every
-        variant through this one call. Implemented by the MaterializedRegionJob and
+        """Process one worker's (non-empty) region jobs against ``branch_name`` and
+        return the per-variable source file results.
```
```suggestion
        """Process one worker's region jobs against ``branch_name`` and 
        return the per-variable source file results.
```

src/reformatters/common/region_job.py
```diff
@@ -431,12 +460,16 @@ def process_worker_jobs(
...
+
+        Each variant owns its own store/session lifecycle and commit cadence behind
```
```suggestion
        Materialized and virtual datasets each own their own store/session lifecycle and commit cadence behind ..
```

src/reformatters/common/region_job.py
```diff
@@ -431,12 +460,16 @@ def process_worker_jobs(
...
+        this one call. The parallel temp-branch coordinator
+        (``DynamicalDataset._process_region_jobs``) drives both variants through it;
+        a virtual *operational* update is the one exception, forking in ``update()``
+        to ``_run_virtual_operational_update`` (single-writer-to-``main``). Callers
```
aldenks: add a todo to the plan to pull documentation about how virtual and materialized datasets run their processing loops into docs and link that where needed rather than scattering notes around comments and docstrings

src/reformatters/common/region_job.py
```diff
@@ -803,6 +839,294 @@ def _cleanup_local_files(
...
+
+class VirtualRegionJob(
+    RegionJob[DATA_VAR, SOURCE_FILE_COORD], Generic[DATA_VAR, SOURCE_FILE_COORD]
+):
+    """Base class for processing a region of virtual Icechunk datasets that point to external files."""
+
+    # Locked to None: an integer max_vars_per_job would split one source file's
+    # variables across separate jobs that commit independently, breaking the
+    # per-file commit atomicity virtual readers rely on. Virtual jobs are tiny
+    # (byte-range refs) and parallelize along the append dim, not variables.
```
```suggestion
    # Locked to None: an integer max_vars_per_job would split one source file's
    # variables across separate jobs that commit independently, breaking the
    # per-file commit atomicity virtual readers rely on.
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -190,15 +172,19 @@ def update(
...
-                worker_index=worker_index,
-                workers_total=workers_total,
-                reformat_job_name=reformat_job_name,
-                template_ds=template_ds,
-                tmp_store=tmp_store,
-                update_template_with_results=True,
-            )
+            if issubclass(self.region_job_class, VirtualRegionJob):
+                # Virtual operational updates are single-writer streaming (see docs/parallel_processing.md)
+                self._run_virtual_operational_update(all_jobs, workers_total)
```
aldenks: add a worker_index argument to _run_virtual_operational_update, pass it in, in that method assert it's 0, and pass it into self.region_job_class.process_worker_jobs rather than hardcoding 0

src/reformatters/common/region_job.py
```diff
@@ -803,6 +834,293 @@ def _cleanup_local_files(
...
+
+    def representative_var(self, coord: SOURCE_FILE_COORD) -> DATA_VAR:  # noqa: ARG002
+        """The variable whose chunk presence means `coord`'s file is fully ingested.
+
+        Used by filter_already_present to probe a single representative cell per
+        file. Default: the first data var, valid when every file contains every
+        variable. Override for one-variable-per-file packings (e.g. GEFS v12
+        reforecast) to return the candidate's own variable.
+        """
+        return self.data_vars[0]
```
aldenks: make the default implementation prefer the first var with step_type == instant if it exists. those are the most likely to have data at every step

src/reformatters/common/region_job.py
```diff
@@ -803,6 +834,293 @@ def _cleanup_local_files(
...
+        candidates: Sequence[SOURCE_FILE_COORD],
+        store: IcechunkStore,
+    ) -> list[SOURCE_FILE_COORD]:
+        """Drop candidates whose representative chunk is already in the manifest.
+
+        Probes ref existence (store.exists) - never reads or decodes a chunk, which
+        would trigger an S3 fetch + GRIB decode. A false-negative (re-emitting done
+        work) is a harmless idempotent rewrite; the manifest probe has no
+        false-positives (a key is present iff we emitted it), so no work is ever
+        skipped undone.
```
```suggestion
        Probes ref existence (store.exists) - never reads or decodes a chunk, which
        would trigger a download + decode.
```
there's no cases where we could false-negative here either, we can drop that. also s3+grib is just one case

PR #656, opened 2026-06-10
GEFS 10-day spatial -dev virtual dataset (virtual datasets PR 4)

aldenks:
PR 4 of #513: the first concrete virtual icechunk dataset, to exercise the structure end to end. `noaa-gefs-forecast-10-day-spatial-dev` — 4 inits/day, 0–240h leads, the GEFS 35-day forecast's variables that exist in the 0.25° `s` files. The `-dev` suffix marks it as a deliberately throwaway operational test: it exists to iterate on the virtual-dataset approach (loop design, manifest splitting, latency) while dataset-structure questions stay open.

## Design

Chunks are byte-range references into NOAA's GRIB archive, decoded at read time by a per-variable `GribberishCodec`. The grid is therefore the **native 0.25° s-file grid** (latitude 90→-90, longitude 0→359.75, 721×1440) with one chunk per message — a byte-range ref can't be upsampled or longitude-rolled the way the materialized `read_data` pipeline does. Lead times stop at 240h where the s files end, and the three a/b-only variables drop out, leaving 19 vars all packed in **one s file per (init, member, lead)**.

Raw-GRIB values also mean raw-GRIB metadata: temperatures are Kelvin (`units: K`, exempted from cross-dataset consistency), and precipitation is the window accumulation — a different quantity than the materialized rate, so it gets a distinct name, **`total_precipitation_surface`** (ECMWF `tp`, `precipitation_amount`, kg m-2), rather than a consistency exemption.

- **Template config** — virtual encoding per the plan's encoding rules: float64 (gribberish's native decode dtype), NaN fill, `shards=None`, empty compressors/filters (empty, not None — `exclude_none` serialization would drop a None and zarr would apply its default zstd on top of the GRIB bytes), `GribberishCodec(var=<grib_element>)` serializer. Reuses the 35-day shared variable configs with encodings swapped via `replace`. These invariants are now also validated by `DynamicalDataset` at registration.
- **Region job: the tick loop.** `process_virtual_refs` polls: each tick lists the bucket (obstore via the cached `common/download.py::s3_store`), diffs against pending files, fetches newly available `.idx` files on a 64-wide threadpool, and yields everything found as one commit. A file is available once **both** its data and index objects are listed (the `.idx` lands 0–5s after the `.grib2`); listed sizes resolve the index's unknown last-message end byte, so there are no per-file HEAD requests. `processing_mode` selects the stopping rule: `"update"` polls until everything expected is ingested (the pod deadline is the only timeout — a `DeadlineExceeded` job failure is the right signal for a file that never published); `"backfill"` sweeps what exists once and exits. The coord's `get_url()` returns the `s3://noaa-gefs-pds/...` location directly — it is the ref location, matching the virtual chunk container — and index downloads go through a new common `s3_download_to_disk`, so all object-store access shares one obstore stack.
- **Dataset: one cron fire per model run.** Schedule `43 3,9,15,21 * * *` — 3 minutes before the earliest observed publication start (measured from S3 mtimes: starts init+3:46:29–3:47:57, a 90-second band; ends ≤init+5:39) — with a deadline of init+5:53, kept well under the 6h between fires so they never overlap. Manifest splits are **one week of inits** (28): every commit rewrites each touched array's active split, so split size caps per-commit cost; measuring this tradeoff at real cadence is an explicit goal of the -dev test. Minimal validators (manifest-aware validation is plan PR 7).

## Hardening from architecture review

- The operational driver asserts jobs are constructed with `processing_mode="update"` (forgetting it would silently degrade updates to a useless pre-publication sweep).
- `process_virtual_refs` yields `(coord, refs)` pairs and the write loop asserts each file's refs cover the cell `filter_already_present` probes — a violation would re-ingest the file forever, invisibly.
- A writer committing to `main` mid-backfill (e.g. this dataset's update cron) makes finalize skip publishing the backfill's branch with a warning; operational rule recorded in `docs/virtual_datasets.md`: suspend the `-update` cron during virtual backfills.
- `copy_zarr_metadata` ordering is per-format (icechunk needs `zarr.json` before coordinate chunks; the branch commit keeps readers safe).

## Verification

- 23 fast tests (URLs, coord generation incl. lead-0 and b22-transition var filtering, tick-loop sweep/poll behavior, discovery's data+index gating, raw-value attrs, coords/encodings, cron schedule/deadline/single-writer) plus the template/CF/registry suites.
- Two slow tests against real NOAA data: a **backfill_local + operational update round trip** over an immutable 2024 archive slice with exact snapshot-value assertions, and a **source contract test** (list the real bucket, resolve byte ranges for all 19 vars from a real index, decode a message, assert the raw grid matches the template lat/lon exactly).
- Real local backfill of the 4 most recent inits (10,044 files / ~190k refs): **3m02s** (was ~1h with the original serial single-pass loop), 7 commits (was 629), 8.1 MB of manifests; values verified through gribberish + anonymous S3.
- Measured `.idx` fetch throughput vs pool width: 8 → ~105 files/s, 64 → ~380 (chosen), 128 → ~435.
- An operational-cadence replay profiler (real listings/fetches/commits against one init's actual S3 publication mtimes on a simulated clock) — results in the PR conversation.

## Knowingly deferred

- Manifest-aware validation (plan PR 7) and `save_config` persistence (plan PR 5).
- Ref maintenance for re-published source files and NOMADS→archive repointing (plan open question 6).
- Generalizing the tick loop / listing discovery into `VirtualRegionJob` or shared utilities — deliberately deferred until the second dataset (AIFS ENS on the GCS mirror) shows what's actually general.


aldenks:
**Local backfill results** (latest 4 init times, 2026-06-09T12 → 2026-06-10T06):

- 189,596 virtual refs committed in 629 batches (16 whole files per commit); one transient HEAD failure mid-run validated crash recovery — the rerun's manifest filter skipped all committed work and resumed.
- Read-back through gribberish at lead 24h: t2m 210–322 K, precip 0–153 kg m-2, fully finite.

**Manifest measurements** (local icechunk store, manifests are one file per array per split per snapshot):

- ~18.9 bytes/entry on disk (largest tip manifest: 189,573 B for 10,044 refs).
- Live tip for 4 inits ≈ 3.6 MB across 19 data vars; the manifests dir totals ~1 GB because all 629 per-commit rewrites of the active split are retained (expiration/GC reclaims).

**Extrapolation at split_size=365×4 (one year):** per-array manifest grows to ~3.67M entries ≈ 69 MB by year end; the full 2020-to-present archive (≈8,300 inits, ~394M entries) totals ~7.4 GB of tip manifests. Entries-per-manifest is comfortably under the ~10M heuristic (which would allow ~2.7 years/split) — but each commit rewrites every touched array's active split, so per-file operational commits would rewrite up to 19 × 69 MB ≈ 1.3 GB at year end, and a reader's first chunk access fetches its array's whole split manifest (~69 MB). Split-size testing (PR conversation) should weigh write churn + first-read fetch against object count, not just entry count.

PR #671, opened 2026-06-15
Add ECMWF IFS ENS 15-day spatial virtual dataset (PR 6)

aldenks:
Second concrete virtual icechunk dataset (#513 PR 6): `ecmwf-ifs-ens-forecast-15-day-spatial-dev`. A different provider that proves the `VirtualRegionJob` abstractions generalize — no base-class changes — and stresses three things the GEFS dataset (PR 4) did not.

## What's new vs GEFS spatial

- **Different index format.** ECMWF `.index` is JSON-lines with explicit `_offset`/`_length` per message (no implied ends), so virtual `_file_refs` needs none of GEFS's "last message → file size" PAD fallback. Reuses `ecmwf_grib_index.grib_message_byte_ranges_from_index`, plus a new `…_by_member` that parses one index once and returns every member's ranges.
- **Multi-member-per-file packing.** One open-data file packs *all* members of a lead, so a single source file commits many member chunks in one atomic commit — exercising the base `chunk_key`'s "a multi-chunk dim must appear in `out_loc`" path. Post IFS-50r1 (2026-05-12 06z) the control member lives in `oper-fc` and perturbed members in `enfo-ef`, so each `(init, lead)` maps to two files; each is its own coord. `append_dim_start` is the first 00z fully in that regime, so there are no historical special cases.
- **Raw-value attrs.** Like GEFS, chunks serve raw GRIB values: Kelvin temperatures, 0-1 cloud fraction, and metres of whole-forecast accumulated `total_precipitation_surface`. The two accumulated-radiation vars are dropped (their raw form needs a bespoke time-integrated name/units; no value for a throwaway). Variable + CRS definitions are reused from the materialized `forecast_15_day_0_25_degree` config so metadata stays in sync.

## Validation (spiked against live data, then automated)

- **gribberish decodes ECMWF IFS GRIB2** (`oper-fc` and `enfo-ef`). The `GribberishCodec(var=…)` label is ignored on decode, so gribberish's occasional ECMWF param misnaming is cosmetic; `var` is set to `grib_element` for consistency with GEFS.
- **Native grid wraps**: longitude `180…359.75` then `0…179.75` (lat 90→-90); the template matches this exactly since chunks serve the raw message.
- **Real local backfill** through the actual icechunk virtual container + `GribberishCodec` zarr read path: control and perturbed members both ingest; ~298K at the equator.
- Full fast suite (1416) + ruff + ty green; slow real-source test resolves 17 vars × 50 members and decodes against live data.

## Operational note

ECMWF open data has **~4-day retention** and no permanent open-data archive, so inits before the retention window can never be backfilled. The store carries an empty init prefix from `append_dim_start` up to wherever operational ingest begins, then grows forward (lazy append-dim expansion). Acceptable for a throwaway `-dev`; a real dataset would pair open data with a permanent archive (MARS/source.coop) for history.

Registered with a primary-only (no replica) icechunk store; cron schedules and the validate threshold mirror the GEFS spatial conventions, tuned to ECMWF's ~07:40 UTC publication.


aldenks:
Closing without merging. PR 6 existed only to prove the `VirtualRegionJob` abstraction generalizes to a second provider and index format (ECMWF IFS ENS, JSON-lines index, multi-member-per-file packing). PR 7 drew the seam from it, so the throwaway `-dev` dataset has served its purpose and won't land on `main`.

PR 7's refactor has been rebased directly onto `main` with the ECMWF dataset excluded: #684. The `ifs-ens-spatial-dev` branch can be deleted.

PR #676, opened 2026-06-24
Store timedelta coords as float64 to fix NaT decode overflow

aldenks:
Forecast lead_time and the derived forecast-length coords were encoded as
int64 seconds. xarray writes NaT as the int64 sentinel (np.iinfo(int64).min),
and decoding that sentinel to a finer resolution (seconds -> us/ns) multiplies
it and overflows int64 on some xarray versions (OutOfBoundsTimedelta /
"Overflow in int64 * timedelta64 multiplication"). The int64[us] decode hint
added in the pandas 3 update made this surface for downstream readers.

No int64 fill sentinel can avoid this: the mask coder maps the fill to
int64.min before the timedelta range check runs. Storing the coords as float64
makes NaT a plain NaN that decodes correctly across xarray versions. CF places
no integer requirement on time/forecast_period encoding, and the in-memory
dtype for readers is unchanged (timedelta64[us], driven by the dtype attribute).
Regenerated every forecast template.

Tests: load every checked-in template's coords (regression for the overflow,
exercising stored NaT sentinels) and assert every timedelta coord is stored as
float64 (cross-template consistency guard). Allow all-fill coords (e.g. all-NaT
ingested_forecast_length) to write no chunk.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01VWewZqk6Z5EQb1efEATFvi


aldenks:
This fixes the timedelta coordinate overflow worked around in dynamical-org/dynamical-stac#28. Once this merges and each dataset runs its next operational update (rewriting the coord arrays from the new float64 template), the workaround there can be removed.

Planning to merge tomorrow.

---

PR #678, opened 2026-06-24
Trigger operational updates from wxopticon source-arrival webhooks

### LINKED ISSUE #682: Kick off runs asap based on new data in virtualize data products

mrshll:
## Context

Operational updates are cron-only today. Each dataset's `ReformatCronJob` runs `<dataset> update` on a fixed schedule timed for *after* source data is expected to land (e.g. GFS at `init+5h22m`). That margin makes updates later than necessary when data is early and failure-prone when it's late.

[wxopticon](https://status.dynamical.org) already watches the upstream sources and POSTs an HMAC-signed webhook the moment a run crosses a readiness boundary. This PR lets those webhooks trigger the update immediately, **keeping the existing cron as a backup** that still fires if a webhook is missed or the receiver is down. Lower latency normally, no reliability regression.

## Approach

The webhook just **fires the existing `<dataset>-update` CronJob early**: an in-cluster receiver clones the live CronJob's `jobTemplate` into a one-off Job (the same thing the "create-job-from-cronjob" ops action does), so image/resources/worker-count stay a single source of truth.

- **Opt-in:** `DynamicalDataset.source_arrival_triggers()` returns the wxopticon products+triggers that should fire a dataset's update. Pilot: `noaa-gfs-forecast` on `external-noaa-gfs-aws` `complete`. Datasets without an override stay cron-only.
- **Receiver** (`common/webhook/receiver.py`, FastAPI): verifies the HMAC signature + timestamp freshness, maps `(product_id, trigger)` → update CronJob, creates a Job named deterministically from `event_id` (redelivered event → 409 → ignored), and tags Jobs from `complete` events with `SOURCE_RUN_COMPLETE=true`.
- **Backup cron skips when already done / in progress:** both trigger paths run the same `update` and derive the same data-driven `run_key`, coordinating via marker files in the existing `_internal/` object-store layer (`common/operational_run_guard.py`). A run marked complete, or with a fresh in-progress lease from another job, is skipped; a stale lease is taken over (self-healing). Only `complete`-triggered runs mark a run complete, so cron/`progress` runs still reprocess the tail to catch late data when the webhook path is down. (This also makes a genuine no-update a clean no-op instead of the previous empty-jobs `finalize()` crash.)
- **Deploy:** `uv run main deploy` now also applies the receiver resources (ServiceAccount + Role/RoleBinding, Deployment, Service, ALB Ingress). New CLI: `serve-webhooks` (container entrypoint) and `register-webhook-subscription`.

## Files
- `common/webhook/{triggers,receiver,deploy,subscription}.py` (new), `common/operational_run_guard.py` (new)
- `common/dynamical_dataset.py` (opt-in method + guard in `update()`), `noaa/gfs/forecast/dynamical_dataset.py` (pilot)
- `common/deploy.py` (CLI + receiver deploy), `common/storage.py` (`delete_coordination_file`)
- `docs/webhooks.md` (new), `AGENTS.md`, `docs/ops_card.md`; `fastapi`/`uvicorn` deps
- Tests: `tests/common/operational_run_guard_test.py`, `tests/common/webhook/receiver_test.py`

## Testing
- `ruff format`, `ruff check`, and `ty check` all pass.
- ⚠️ **pytest could not be run locally:** the only available interpreter is Python `3.14.0b4`, under which the repo's pinned pydantic fails to import (`eval_type_backport` ForwardRef assertion) — reproducible on a clean checkout, unrelated to this change, and not fixable here because the package index is frozen by `exclude-newer`. The new tests follow existing patterns and are type-checked; relying on CI to execute them.

## Before merge / deploy (open items)
- [ ] Set `RECEIVER_HOST` + `ACM_CERTIFICATE_ARN` in `common/webhook/deploy.py` and create the DNS record (EKS auto-mode ALB).
- [ ] Create the `wxopticon-webhook` k8s secret (`WXOPTICON_WEBHOOK_SECRET`) from the subscription secret.
- [ ] Register the subscription: `WXOPTICON_ADMIN_TOKEN=… uv run main register-webhook-subscription https://<host>/webhooks/wxopticon` (deferred — needs the deployed URL + admin token).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,7 @@ This project contains code to reformat weather data into the Zarr v3 / Icechunk 
 Datasets are created in 3 phases:
 1. A template of the dataset, in the form of zarr metadata that is checked into the repo, is created with `uv run main <dataset_id> update-template`. This template (not in-code config) is loaded by steps 2 and 3 and drives processing and output in those steps. This approach of checking in the metadata allow us to review diffs if the structure or metadata of the dataset changes.
 2. A zarr backfill is run. The backfill uses kubernetes indexed jobs to run work in parallel. When the user runs a `uv run main <dataset-id> backfill-kubernetes ...` command the metadata for the zarr is first written by the local process to the final zarr store, then a kubernetes index job is kicked off with each job index responsible for writing a portion of the zarr chunk data into the zarr archive.
-3. Operational updates to the zarr are run using a kubernetes cronjob and validated by another kubernetes cronjob which runs after the update is expected to succeed. Updates use the same parallel worker model as backfills. To ensure the archive is valid to readers throughout the update, zarr v3 metadata is written only after all workers finish, and icechunk stores use a temporary branch that is atomically merged to main.
+3. Operational updates to the zarr are run using a kubernetes cronjob and validated by another kubernetes cronjob which runs after the update is expected to succeed. Updates use the same parallel worker model as backfills. To ensure the archive is valid to readers throughout the update, zarr v3 metadata is written only after all workers finish, and icechunk stores use a temporary branch that is atomically merged to main. A dataset can additionally opt in (via `source_arrival_triggers()`) to having wxopticon source-arrival webhooks trigger its update the moment data lands; the cron then acts as a backup. See [docs/webhooks.md](docs/webhooks.md).
 
 ## Repository structure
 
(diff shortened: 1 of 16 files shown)


aldenks:
We are going to close this and instead have the _virtual_ dataset kick off rechunked dataset processing as soon as a virtual update reaches certain lead time(s)

PR #680, opened 2026-06-26
Test-suite safety nets: structure guard, failure-injection, validator execution, deploy hardening

aldenks:
## Why

A holistic review of test coverage asked: when the suite passes, are we confident we won't ship unexpected crashes, reader-breaking structure changes, incorrect data, or silent holes? This PR closes the gaps that mattered most.

### Runtime structure guard (reader/archive protection)
Before any writes, worker 0 of an operational update now asserts the update template still matches the published store — variable presence, dims, dtype, and chunk/shard sizes (`template_utils.assert_no_structural_drift_from_existing_store`, called from `DynamicalDataset._process_region_jobs`). A drifted template (removed/renamed variable, changed dtype/dims/chunks/shards) would corrupt the existing archive or break readers, so the update now fails fast and leaves the live store untouched. Changing structure requires a backfill, which is exempt.

- All axes are compared in dev/prod. The **append axis** chunk/shard is skipped **only** under `Config.env == test`, where the integration auto-shrink helper sizes append-dim geometry to the (varying) template length; in production those sizes are fixed config and are also caught at PR time by the template-drift test.
- Adding new variables is always allowed (only variables already in the store are checked).

### Failure-injection e2e tests (silent holes)
`tests/common/test_failure_injection.py` (tiny 10×25 synthetic dataset) pins what happens when source files fail mid-update: a transient download failure and a lagging variable each leave NaN holes that resume-from-latest does not refill and that shard-presence validation does not catch. The module docstring documents the intentional design — we don't fail the whole job on individual unreadable files (so one corrupt file / transient error can't hold back otherwise-publishable data), and most `operational_update_jobs` reprocess the most recent existing chunk (one automatic follow-up attempt for a gap in the latest chunk; an older gap needs a targeted backfill).

### Validator execution in e2e tests
Each dataset's configured validators now run against the store its e2e test builds (`assert_configured_validators`), so the validator **configs** are actually exercised (previously `validate_dataset` was fully mocked). Every validator must run without raising, and the recency validators (`check_*_current_data`) must pass (we patch `now` to the store's latest append-dim coordinate). Wired into all 15 dataset e2e tests plus the example scaffold. Virtual datasets are opted out of `check_for_expected_shards` (mirroring `validate_dataset`); manifest-aware validators for them are a separate planned phase.

### Deploy hardening
- Code Quality now also runs on push to `main`, and the production deploy (`deploy-operational-updates.yml`) triggers via `workflow_run` after it succeeds — so the full test suite runs once and the deploy depends on its result, with no duplicated quality steps. The deploy builds/ships the exact commit that passed (`workflow_run.head_sha`).
- `uv sync --locked` on the deploy jobs so the shipped image can't drift from the lockfile that CI tested.
- The Dockerfile build-time smoke test now runs the packaged `main` CLI, so a broken image fails the build instead of shipping.

### Small fix
`iterating.chunk_slices` had an assert on a generator expression (always truthy); now asserts contiguity for real.

## Testing
Full suite green locally (1502 passed, 3 skipped). `ruff format`, `ruff check`, and `ty` all clean. Docs updated (`docs/parallel_processing.md`).

https://claude.ai/code/session_01XyGRKeEzahWb4S4J9dYynJ

### DIFF
--- a/.github/workflows/deploy-operational-updates.yml
+++ b/.github/workflows/deploy-operational-updates.yml
@@ -17,8 +17,47 @@ env:
   IMAGE_TAG: "${{ secrets.DOCKER_REPOSITORY }}:${{ github.sha }}"
 
 jobs:
+  quality:
+    name: Code Quality
+    runs-on: ubuntu-24.04
+
+    # Keep code-quality.yaml and these steps in sync
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+
+      - name: "Set up Python"
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version-file: ".python-version"
+
+      - name: Install the project
+        run: uv sync --all-extras --dev --locked
+
+      - name: Install this project
+        run: "uv pip install -e ."
+
+      - name: Run Ruff
+        run: uv run ruff check --output-format=github
+
+      - name: Check code format
+        run: uv run ruff format --check
+
+      - name: Run ty
+        run: uv run ty check
+
+      - name: Run Pytest
+        run: uv run pytest -p no:sugar --verbose
+
   deploy:
     name: Deploy
+    # Gate the production deploy on the full test suite passing on the merge commit.
+    needs: quality
     runs-on: ubuntu-24.04
     environment: prod
 
@@ -38,7 +77,7 @@ jobs:
           python-version-file: ".python-version"
 
       - name: Install the project
-        run: uv sync --all-extras --dev
+        run: uv sync --all-extras --dev --locked
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6
--- a/.github/workflows/deploy-operational-updates.yml
+++ b/.github/workflows/deploy-operational-updates.yml
@@ -1,8 +1,13 @@
 name: Deploy Operational Updates
 
+# Gate the production deploy on Code Quality passing on main: this fires only after
+# the "Code Quality" workflow completes for a push to main, so the full test suite
+# runs once (no duplicated steps here) and the deploy depends on its result.
 on:
-  push:
-    branches: [ main ]
+  workflow_run:
+    workflows: ["Code Quality"]
+    types: [completed]
+    branches: [main]
 
 # Prevent concurrent runs since this is deploying to the K8S cluster
 concurrency:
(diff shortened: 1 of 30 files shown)


.github/workflows/deploy-operational-updates.yml
```diff
@@ -14,17 +19,24 @@ permissions:
...
 env:
-  IMAGE_TAG: "${{ secrets.DOCKER_REPOSITORY }}:${{ github.sha }}"
+  # workflow_run's github.sha is the default-branch tip, so pin to the exact commit
+  # that passed Code Quality.
+  IMAGE_TAG: "${{ secrets.DOCKER_REPOSITORY }}:${{ github.event.workflow_run.head_sha }}"
 
 jobs:
   deploy:
     name: Deploy
+    # Only deploy if the Code Quality run succeeded.
```
aldenks: remove comment

.github/workflows/deploy-operational-updates.yml
```diff
@@ -14,17 +19,24 @@ permissions:
...
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubuntu-24.04
     environment: prod
 
     # Keep deploy-staging.yaml and these steps in sync
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          # Build/deploy the exact commit that passed Code Quality, not the branch tip.
```
aldenks: remvoe, duplicate of comment above

tests/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset_test.py
```diff
@@ -14,6 +14,9 @@
     NoaaNdviCdrAnalysisRegionJob,
 )
 from tests.chunk_utils import shrink_chunks_and_shards
+from tests.common.dynamical_dataset_test import (
+    assert_configured_validators,
+)
```
aldenks: fix uncessary multi line wrap in all dataset test imports 
`from tests.common.dynamical_dataset_test import assert_configured_validators`

tests/example/dynamical_dataset_test.py
```diff
@@ -52,6 +55,11 @@
 #         subset_ds["your_variable"].values, [190.0, 163.0, 135.0]
 #     )
 
+#     # Run this dataset's configured validators against the built store: every
+#     # validator must not raise, and the recency (check_*_current_data) validators
+#     # must pass. Catches validator config bugs that would crash the validation cron.
```
aldenks: also remove this comment

.github/workflows/deploy-operational-updates.yml
```diff
@@ -1,8 +1,13 @@
 name: Deploy Operational Updates
 
+# Gate the production deploy on Code Quality passing on main: this fires only after
+# the "Code Quality" workflow completes for a push to main, so the full test suite
+# runs once (no duplicated steps here) and the deploy depends on its result.
```
aldenks:
```suggestion
# Gate the production deploy on Code Quality passing on main.
```

deploy/Dockerfile
```diff
@@ -132,7 +132,9 @@ COPY . /app/
 USER app
 WORKDIR /app
 
-# Light test that the app can be run
+# Build-time smoke test: fail the build if the app can't start, so a broken image never
+# ships. Mirrors the prod ENTRYPOINT (python src/reformatters/__main__.py) — the exact
+# invocation k8s jobs use, just with --help.
```
aldenks: Revert, this file should have no diff. previous comment was sufficient

PR #686, opened 2026-06-26
Vertical dimension support: group-keyed dims + DataTree templates

aldenks:
Implements the vertical-dimension structure from `docs/plans/vertical_dimension_structure.md` (#679): datasets can mix single-level (root) variables with variables on a dense vertical dimension (`pressure_level`, `model_level`), each vertical group becoming a zarr group named after its dimension.

## Design — Decision A (DataTree is the canonical template)

The dataset template is now an `xarray.DataTree` (`get_template()` returns it; `RegionJob.template_ds` holds it). This was chosen over a flat path-keyed Dataset so the in-memory object matches the on-disk group hierarchy, coordinate scoping is structural via inheritance, and there's no unenforceable "don't `to_zarr` me raw" invariant. **A single-level dataset is a one-node tree and writes byte-identically to before** — all 15 existing templates regenerate with zero diff.

## What's in it

- **Type scheme** (`config_models.py`): `RootGroup`/`ROOT`, `VerticalGroup`, `Group`, `var_path`; `DataVar.group` (default `ROOT`) + `.path`. `Dim` gains `pressure_level`/`model_level`.
- **Template layer** (`template_config.py`): `dims` is `dict[Group, tuple]`; `all_dims`/`groups`; `update_template` builds + writes a DataTree (`write_inherited_coords=True` so groups open standalone); `get_template` returns a DataTree; structure validators (group==added dim, path uniqueness, root-var≠group-name, no orphan group, uniform append-dim chunk, encoding lengths).
- **One write path** (`template_utils.py`): a bare Dataset wraps as a one-node tree.
- **Region jobs**: `template_ds` is a DataTree; `get_jobs` walks vars by `var.path`; `source_groups` → `source_file_var_groups` (a source file may span vertical groups). Virtual write loop keys refs/probe/chunk_key by `var.path`; `sync_dims_to` grows the append dim per group node (`DataTree.to_zarr` has no `append_dim`).
- **Coordination/zarr/dataset**: thread the DataTree through; `copy_zarr_metadata` handles group-nested coords/metadata.
- **Guard**: materialized chunk-writes are not yet group-aware, so `DynamicalDataset` rejects a materialized dataset that declares a non-root var (fails loudly rather than miswriting). Group-aware materialized writes + the real **NOAA HRRR** / **DWD ICON-EU** datasets are the follow-up.

All 16 existing configs migrated to `dims = {ROOT: (...)}`.

## Tests

- New `tests/common/multi_group_template_test.py` (7) — DataTree write, standalone group open with shared coords, reindex, validators (incl. the name-collision case `pressure_level/temperature` + `model_level/temperature`).
- New `tests/common/virtual_multi_group_test.py` (5) — virtual backfill landing refs at `pressure_level/temperature`, per-group `sync_dims_to`, per-file commit atomicity across groups, two-worker + operational.
- All existing suites migrated to DataTree templates. `ty` + `ruff` clean; full suite green.

Two review passes (correctness/simplicity + comment-style conformance) were run and their findings folded in.

### DIFF
--- a/src/reformatters/common/config_models.py
+++ b/src/reformatters/common/config_models.py
@@ -1,5 +1,4 @@
 from collections.abc import Sequence
-from enum import Enum, auto
 from typing import Annotated, Any, Generic, Literal, TypeVar
 
 import numcodecs
@@ -7,20 +6,7 @@ import numcodecs.abc
 import pydantic
 
 from reformatters.common.pydantic import FrozenBaseModel
-from reformatters.common.types import TimedeltaUnits, TimestampUnits
-
-
-class RootGroup(Enum):
-    ROOT = auto()  # pure sentinel; the root zarr group has no name
-
-
-ROOT = RootGroup.ROOT
-
-# A variable on a dense, comparable vertical dimension lives in a zarr group named
-# after that dimension (group name == dimension name). Expand as new types are added.
-type VerticalGroup = Literal["pressure_level", "model_level"]
-# A variable's group: ROOT (single-level, lives at the dataset root) or a vertical group.
-type Group = VerticalGroup | RootGroup
+from reformatters.common.types import ROOT, Group, TimedeltaUnits, TimestampUnits
 
 
 def var_path(group: Group, name: str) -> str:
--- a/src/reformatters/common/config_models.py
+++ b/src/reformatters/common/config_models.py
@@ -217,5 +217,4 @@ class DataVar(FrozenBaseModel, Generic[INTERNAL_ATTRS_co]):
 
     @property
     def path(self) -> str:
-        """The variable's zarr path / identity key (`name` at root, else `group/name`)."""
         return var_path(self.group, self.name)
--- a/src/reformatters/common/config_models.py
+++ b/src/reformatters/common/config_models.py
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from enum import Enum, auto
 from typing import Annotated, Any, Generic, Literal, TypeVar
 
 import numcodecs
@@ -7,6 +8,27 @@ import pydantic
 
 from reformatters.common.pydantic import FrozenBaseModel
 from reformatters.common.types import TimedeltaUnits, TimestampUnits
+
+
+class RootGroup(Enum):
+    ROOT = auto()  # pure sentinel; the root zarr group has no name
+
+
+ROOT = RootGroup.ROOT
+
+# A variable on a dense, comparable vertical dimension lives in a zarr group named
+# after that dimension (group name == dimension name). Expand as new types are added.
+type VerticalGroup = Literal["pressure_level", "model_level"]
+# A variable's group: ROOT (single-level, lives at the dataset root) or a vertical group.
+type Group = VerticalGroup | RootGroup
+
+
+def var_path(group: Group, name: str) -> str:
+    """The zarr path / identity key of a variable: `name` at root, else `group/name`."""
+    if group is ROOT:
+        return name
+    return f"{group}/{name}"
+
 
 type AttributeStr = Annotated[str, pydantic.Field(pattern=r"^[A-Z0-9].*[^.]$")]
 type Sentence = Annotated[str, pydantic.Field(pattern=r"^[A-Z].*\.$")]
@@ -188,6 +210,12 @@ INTERNAL_ATTRS_co = TypeVar(
 
 class DataVar(FrozenBaseModel, Generic[INTERNAL_ATTRS_co]):
     name: str
+    group: Group = ROOT
     encoding: Encoding
     attrs: DataVarAttrs
     internal_attrs: INTERNAL_ATTRS_co
+
+    @property
+    def path(self) -> str:
+        """The variable's zarr path / identity key (`name` at root, else `group/name`)."""
+        return var_path(self.group, self.name)
--- a/src/reformatters/common/region_job.py
+++ b/src/reformatters/common/region_job.py
@@ -1,4 +1,4 @@
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Iterator, Mapping, Sequence
 from enum import Enum, auto
 from itertools import chain, pairwise
 from pathlib import Path
@@ -28,7 +28,7 @@ from zarr.abc.store import Store
 from reformatters.common import storage
 from reformatters.common.config_models import DataVar
 from reformatters.common.iterating import (
-    dimension_slices,
+    chunk_slices,
     split_groups,
     spread_evenly,
 )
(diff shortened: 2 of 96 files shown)


src/reformatters/common/region_job.py
```diff
@@ -118,7 +127,8 @@ def region_slice(s: slice) -> slice:
 
 class RegionJob(pydantic.BaseModel, Generic[DATA_VAR, SOURCE_FILE_COORD]):
     tmp_store: Path
-    template_ds: xr.Dataset
+    # Whole-dataset template tree: root + one node per vertical group (single-level = one node).
```
aldenks: remove comment

src/reformatters/common/region_job.py
```diff
@@ -327,35 +335,45 @@ def get_jobs(
...
-        sharded = [
-            v.encoding.get("shards") is not None for v in template_ds.data_vars.values()
-        ]
-        assert len(set(sharded)) == 1, (
-            "all data vars must agree on whether they are sharded"
+        # Materialized arrays partition by shard; virtual arrays use chunks. All data
+        # vars must agree: a mix would partition by whichever var is first, and a
+        # chunk-sized region over a sharded var would let multiple workers write the
+        # same shard. The tree-walking twin of iterating.dimension_slices (whose flat
+        # Dataset can't hold same-named vars from different groups).
```
aldenks: This tree walking twin comment isn't clear what in this file it is referring to, but it also seems unnecessary. Remove or clarify

Should we just get rid of dimension slices function or change it to take a data tree instead? Why do we need to keep Dataset specific behavior

src/reformatters/common/template_config.py
```diff
@@ -26,11 +28,32 @@
...
+                ordered[dim] = None
+        return tuple(ordered)
+
+    @property
+    def groups(self) -> tuple[Group, ...]:
+        """Groups to write: ROOT (always, holds shared coords) then each vertical group
+        that has at least one data var (empty vertical groups are omitted)."""
+        used = {var.group for var in self.data_vars}
+        vertical = tuple(g for g in self.dims if g is not ROOT and g in used)
+        return (ROOT, *vertical)
```
aldenks: It should be a run time error AND be caught by a general cross all template configs test if there's a dim that is not used by any vars.

src/reformatters/common/template_config.py
```diff
@@ -26,11 +28,32 @@
 class TemplateConfig(FrozenBaseModel, Generic[DATA_VAR]):
     """Define a subclass of this class to configure the structure of a dataset."""
 
-    dims: tuple[Dim, ...]
+    # Dimensions per zarr group, keyed by group. dims[ROOT] is the root/single-level
+    # dims; dims["pressure_level"] etc. are a vertical group's full dims (the shared
+    # dims plus the vertical dim, whose name equals the group name). A single-level
+    # dataset declares only {ROOT: (...)}. See docs/plans/vertical_dimension_structure.md.
+    dims: dict[Group, tuple[Dim, ...]]
     append_dim: AppendDim
```
aldenks: Remove comment, this is covered in agents

src/reformatters/common/template_config.py
```diff
@@ -26,11 +28,32 @@
...
     append_dim_frequency: Timedelta
 
+    @property
+    def all_dims(self) -> tuple[Dim, ...]:
+        """De-duplicated union of every group's dims, root dims first."""
+        ordered: dict[Dim, None] = {}
+        for group_dims in self.dims.values():
+            for dim in group_dims:
+                ordered[dim] = None
+        return tuple(ordered)
```
aldenks: Append to a list if dim not in list for more clear intent, then return tuple of final list

src/reformatters/common/template_config.py
```diff
@@ -91,88 +116,175 @@ def append_dim_coordinates(self, end: DatetimeLike) -> pd.DatetimeIndex:
...
 
-    def get_template(self, end_time: DatetimeLike) -> xr.Dataset:
+    def get_template(self, end_time: DatetimeLike) -> xr.DataTree:
         """
-        Returns a template dataset expanded to the given end time.
+        Returns the template, as a DataTree, expanded to the given end time.
+
+        A single-level dataset is a one-node tree (just the root); a dataset with
+        vertical groups has one child node per group. Each node's append dimension
+        is reindexed and its coordinates re-derived.
```
aldenks:
```suggestion
        Returns the template as a DataTree, with append dim in all Zarr groups expanded to the given end time.
```

src/reformatters/common/config_models.py
```diff
@@ -8,6 +9,27 @@
...
+    ROOT = auto()  # pure sentinel; the root zarr group has no name
+
+
+ROOT = RootGroup.ROOT
+
+# A variable on a dense, comparable vertical dimension lives in a zarr group named
+# after that dimension (group name == dimension name). Expand as new types are added.
+type VerticalGroup = Literal["pressure_level", "model_level"]
+# A variable's group: ROOT (single-level, lives at the dataset root) or a vertical group.
+type Group = VerticalGroup | RootGroup
```
aldenks: This belongs in types.py yeah?

src/reformatters/common/zarr.py
```diff
@@ -97,8 +98,17 @@ def _copy_data_var_chunks(
         sync_to_store(store, key, file.read_bytes())
 
 
+def _coord_chunk_globs(template_ds: xr.DataTree) -> list[str]:
+    """Glob patterns for every coordinate's chunk files, group-prefixed."""
+    globs = []
+    for node in template_ds.subtree:
+        prefix = f"{group}/" if (group := node_group_name(node)) else ""
+        globs.extend(f"{prefix}{coord}/c/**/*" for coord in node.to_dataset().coords)
```
aldenks: Does this catch non nested coords?

src/reformatters/common/region_job.py
aldenks: Removed the confusing comment. On whether to unify: `dimension_slices` operates on an opened-store `Dataset` for arbitrary dims (validation reads the published store as a Dataset), whereas `get_jobs` partitions the multi-group template `DataTree` — where two groups can hold same-named vars (e.g. `pressure_level/temperature` + `model_level/temperature`), so it can't be flattened to one Dataset. Different inputs, so I kept `dimension_slices` Dataset-based and dropped the misleading cross-reference rather than forcing one to wrap the other.

src/reformatters/common/zarr.py
aldenks: Yes. For the root node `node_group_name(node)` is None so the prefix is empty, giving `<coord>/c/**/*` — the non-nested (root) coords. Child nodes additionally yield `<group>/<coord>/c/**/*`, which catches each group's own duplicated copy of the shared coords (written by `write_inherited_coords=True`). Both on-disk paths exist and need copying.

PR #693, opened 2026-06-29
Offline/backfill validation tooling for virtual datasets (PR 11)

aldenks:
Implements **PR 11** of #513: make `src/scripts/validation/` feasible on a virtual, multi-vertical-group store, and add the thorough whole-archive correctness pass that's run after a backfill (distinct from the cheap operational `-validate` checks in PR 8).

## What changed

**Open virtual + grouped stores** — `load_zarr_dataset` authorizes a virtual store's persisted chunk containers (anonymous S3) and opens via `open_flattened_dataset`, so vertical-group vars (`pressure_level/temperature`, `model_level/…`) are visible, not just root. `var_slug` keeps group-pathed names filesystem/anchor-safe in plot filenames and rendered HTML.

**Vertical levels (orthogonal to virtual, new with HRRR-spatial's two level groups)** — the plots sample one representative level per group variable (middle by default, `--level` to override), recorded in the summary and titles. Generic across `pressure_level`/`model_level`; `compare_spatial`/`compare_timeseries` select the level and align the reference; `value_timeseries` pins it. `manifest_scan` is the exception — it covers every level.

**Virtual read economics** — a virtual chunk decodes a source GRIB message on read, so a spatial snapshot is cheap and an append-dim point column is not. `report_nulls` skips value-based scanning on virtual (availability is a manifest question → `manifest_scan`); `value_timeseries` pins lead/level/member and samples the append dim with a logged decode budget.

**Whole-archive scans** (dataset-aware, reuse the operational validators):
- `manifest_scan.py <dataset-id>` — strict completeness gate via ref existence (no decode; per source **file**, extended to all of a file's refs by the one-file-one-commit invariant), availability plot + backfill `--filter-contains` retry filters, non-zero exit on gaps.
- `decode_scan.py <dataset-id>` — sampled decode health across positions/leads/members/levels; explicitly **not** exhaustive.
- `CheckVirtualDecodeHealth` gains `sampled_levels` (bounds group-var decodes — also improves the operational check on grouped datasets) and `max_positions`.

**Docs** — `docs/validation.md` gains a "Virtual and multi-level datasets" section: the inverted cost model, the completeness-exhaustive / decode-sampled guarantee split, the two scans, and level sampling.

## Notes
- The guarantee split: **completeness is exhaustive at file granularity** (every expected source file ingested), **decode health is sampled** — a level missing *within* an ingested file is only caught by decode sampling. "Validated" means every reference is present and a representative sample decodes cleanly — not that every value was decoded. Per the design discussion, the K8s every-chunk decode sweep is intentionally out of scope.
- Single-level-per-var (with `--level`) mirrors how the tool already samples one ensemble member / one snapshot, rather than exploding ~14 pressure vars × 39 levels + model vars × 50 into 550+ report rows.

`uv run ruff format && ruff check && ty check` clean; validation/virtual/spatial test suites pass (new unit tests included).


aldenks:
Self-review pass (in an isolated worktree, merged against current `main`) — findings and the fixes pushed in 2b81564b:

**Fixed**
- **Decode-scan sampling could skip whole variable groups.** Jobs are ordered region-major with var groups innermost, so evenly-spaced sampling over the raw job list strides `(region × var group)` — on a dataset with `max_vars_per_job` set, the stride can alias to the same group every pick. Now samples evenly over *regions* and keeps every var-group job at each sampled region. (Both current virtual datasets have one var group, so this is a future-proofing correctness fix, not a live bug.)
- **Manifest scan was serial over jobs.** A whole HRRR archive is ~11.6k single-init region jobs; each probe batch is concurrent internally but jobs ran one at a time (est. 30–60 min). Now threaded (16 workers) with progress logging — same probes, ~10× faster.
- **Overclaim: "completeness covers all levels".** `filter_already_present` probes **one representative ref per source file**; the whole-file guarantee comes from the one-file-one-commit invariant. A level missing *within* an ingested file (e.g. a source GRIB index that omitted one level's message while the representative message existed) is **not** caught by the manifest scan — only decode_scan's level sampling / NaN checks can surface it. Docs, PR description, and scan summary wording corrected to "exhaustive at file granularity".
- **`validation_summary.md` said "None detected" for nulls on virtual runs** where the value scan was skipped — now says it wasn't value-scanned and points to `manifest_scan.py`.
- **Merged current `main`** and fixed the semantic drift from #701 (`process_virtual` signature) in the new multi-group test.

**Verified**
- Materialized validation unbroken: `tests/common/` + `tests/dwd/` + `tests/noaa/gfs/` (821 passed) plus the scripts suite, on the merged state. The only materialized-path behavior change is `load_zarr_dataset` opening via `open_flattened_dataset` (single-group stores flatten to exactly their root dataset) with explicit `consolidated=True` for zarr stores — matches what `template_utils.write_metadata` always writes.
- Decode budget documented: virtual `run-all` ≈ 2×200 messages/var for value-ts + per-point series for temporal — order 10⁵ full-field decodes for a ~150-var dataset. Per-var counts are logged pre-read; bound with `-v` / `--start-date`.

**Known limitations (pre-existing, unchanged)**
- `is_virtual_store` / `load_zarr_dataset` only recognize `s3://…icechunk` stores; a local-path icechunk store was never openable by these scripts before this PR either.
- `--min-fraction` applies uniformly per position — no newest-first tiering like the operational validator (use `--end` to exclude the still-publishing tail).

aldenks:
Field test: ran this PR's tooling for real against the live `noaa-hrrr-forecast-48-hour-spatial` store (**177 variables**, 3 groups), and used it to exercise the new batched agent-review process for datasets whose plots can't fit one agent context. Three commits pushed from this exercise.

## Tooling vs prod store — all passing
- **Open path**: virtual store detected from persisted containers, all groups flattened, 177 vars visible.
- **`manifest_scan`** (3-day window): 12/12 positions 100% complete in **9s**.
- **`decode_scan`** (1 region ≈ 1,200 real GRIB decodes incl. level sampling): **passed, 2m28s**.
- **`run-all`** (12 representative vars incl. both vertical groups, scoped window): **2m36s (~13s/var)** — sane physics in stats, group-var slugs and virtual routing all correct. Extrapolates to ~40 min for all 177 on a scoped window.
- Whole-archive manifest scan (11.6k positions, ~1.7M file probes) in flight; its runtime points at the next optimization lever (memoize the per-job `zarr.open_group` in `filter_already_present` callers).

## Found and fixed by the field test
1. **Scans required Kubernetes credentials to *read* a public bucket** — `store_factory.primary_store()` loads write creds from k8s secrets; a laptop scan died with a k8s 401. Scans now open the primary URL read-only + anonymously (`open_icechunk_readonly`, shared with the plots path).
2. **`-v` typos silently shrank runs** — `select_variables_for_plotting` dropped unknown names quietly; a wrong `-v precipitation_surface` (this dataset calls it `total_precipitation_surface`) cost a variable its review and nothing said so. Now raises listing the missing names. *(Caught by a batch reviewer, not by me.)*
3. **Random spatial lead could outrun the reference** — lead=28h put HRRR's valid time 22h past GEFS analysis's newest snapshot. Random lead choice now prefers leads the reference covers.
4. **`std 0.0` reported where spread is undefined** (analysis / virtual pinned-lead series) — now `n/a`.

## Batched review for many-variable datasets (new §3f in docs/validation.md)
At 177 vars × 4 PNGs a single review context is ~1M tokens — the review must fan out. §3f documents the contract, validated end-to-end on the real run output:
- **Lead agent** reads only the summary header + scan summaries, partitions **theme-coherent batches of 10–15 vars**, spawns parallel batch subagents, aggregates structured verdicts, dispatches verification subagents (§3d), writes `## Summary`. The lead never opens a plot image.
- **Measured**: batch reviewers spend **~6k tokens/var**, ~2.5 min per 6-var batch; both returned precise structured findings (e.g. flagged 2m dewpoint spatial min of −81 °C with an exact byte-range verification plan).
- **Verification loop**: a subagent fetched the source `.idx`, byte-ranged the single DPT message, decoded with gribberish — **confirmed −81.025 °C is in NOAA's GRIB** (source-faithful, not a decode bug) in 38s / ~18k tokens.
- The reviewers also caught fix #2 above and the `std 0.0` artifact — the process finds tool bugs, not just data bugs.

aldenks:
Whole-archive manifest scan completed (follow-up to the field-test comment):

**Result: the archive is effectively fully backfilled — and the scan surfaced a template-modeling gap.**

- 11,648 positions, 1.71M expected source files probed: **92.60% present, 3,495 positions flagged incomplete**, runtime 154 min.
- The availability plot is bimodal and structural, not random: flat at exactly **0.755 from 2018-07 → 2020-12**, then a step to **1.0 from the HRRR v4 cutover (2020-12-02) → present**. 111/147 files present per pre-v4 position = missing leads 37–48 × 3 file types: **pre-v4 HRRR only forecast to 36h — those files don't exist at NOAA**. The v4 era is 100% complete.
- Remaining real gaps are small and isolated: deeper dips in the first weeks of the archive (2018-07/08, down to 0.25) and single-position dips (~2019-03, 2019-11, 2020-03, late-2020, ~2022-02, ~2026-01) — `missing_source_files.txt` has per-position retry filters for triage against upstream.

**Recommendation (dataset, not scan):** `noaa-hrrr-forecast-48-hour-spatial`'s `expected_forecast_length` is a constant 48h for all inits (`template_config.py` `np.full(..., EXPECTED_FORECAST_LENGTH)`). Encoding the v3→v4 history (36h before 2020-12-02T00, 48h after) would make reader expectations truthful and let this scan (and any expected-coverage check) treat pre-v4 leads 37–48 as structurally absent instead of missing. The scan reporting "the archive doesn't contain what the template claims to expect" is arguably correct behavior — the template's claim is what needs fixing.

**Known follow-up (scan performance):** 154 min for the whole archive is dominated by the per-job `zarr.open_group` inside `filter_already_present` (11.6k single-init jobs × ~180 arrays of metadata). Letting callers pass a pre-opened group (or memoizing per store) is the next lever; the bounded operational validator doesn't suffer from this.

Issue #700, opened 2026-07-01
Virtual dataset performance optimizations


aldenks:
Performance optimizations for virtual (Icechunk) datasets, found while backfilling `noaa-hrrr-forecast-48-hour-spatial`. Sub-issue of #513.

**Not all of these are required to ship** the first virtual dataset. See "Validated commit-cost model" below — the two reformatters-side items (contiguous assignment, per-group splits) are the load-bearing fixes; the icechunk fork items turned out to be nice-to-haves at sane manifest counts.

### Validated commit-cost model (2026-07-02)

Measured on the v0.4.0 run (per-commit `rebase_attempts` from snapshot metadata + per-worker phase timings):

`commit_latency ≈ (1 + rebase_attempts) × per_flush_cost`

- **rebase_attempts ≈ O(parallelism), stationary** — 97% of commits needed ≥1 rebase at p10, median ~4–5; 6.3× write amplification (1,528 flushes / 242 commits).
- **per_flush_cost = read-modify-write of every immutable manifest window the batch touches** — this is what grows with archive fill. `spread_evenly` scattered each worker's 30 inits across the whole append dim → ~3,400 window RMWs and GB-scale rewrite per flush (flush 38s empty → 190s at 62% fill; discover/build/emit flat throughout). The v0.4.0 store accumulated ~6.6M orphaned manifest objects from this churn.
- Manifest-count-driven costs (snapshot serialization, the O(M²) scan) are negligible once M is in the low thousands.

### Implementation items

- [x] **Emit optimization** — `_emit_refs`/`chunk_key` recomputed per-variable geometry per ref (~21s/init); vectorized. #699 (merged).
- [x] **Commit batching** — one commit per worker instead of per job. #701 (merged).
- [x] **Grow manifest split 8 → 120** — #702 (merged). Fixed the O(M²)-scan regime (M≈257k → ~17k).
- [x] **Differentiate manifest split by array group** — #706 (merged).
- [x] **Per-group split sizes from measured bytes/ref (~16.4 B/ref) + v0.5.0** — single-level 2400 (~1.8 MiB), pressure 180 (~5.4 MiB), model 160 (~6.1 MiB); M ≈ 3,085. #707 (merged).
[...]

PR #706, opened 2026-07-02
Differentiate manifest split by array group for HRRR-spatial

aldenks:
## What

Give `noaa-hrrr-forecast-48-hour-spatial` a **per-array-group manifest split**: single-level (ROOT) arrays get a coarse split of 480, dense vertical groups (`pressure_level`, `model_level`) stay at 120.

## Why

Per-commit cost grows with the total manifest split count **M = array_count × ceil(appends / split_size)** (each commit re-serializes the snapshot's manifest list and touches every split of every array).

HRRR-spatial's 177 arrays at a uniform split of 120 gave **M ≈ 17,169**, and **81% of that (14k splits)** came from the 143 single-level arrays — which are ~50× sparser per init than the vertical groups, so their manifests were only ~50 KiB each. We were paying full M-driven commit cost to split arrays that never needed splitting.

This is the axis that made HRRR-spatial's commits climb as the archive filled, where the 19-array GEFS-spatial backfill (M ≈ 2k, same code path) never did — the difference is array count driving M, not manifest bytes.

Coarsening only the single-level arrays cuts **M ≈ 2.5× (17.2k → 6.9k)** with **no manifest larger than today** (~2.5 MiB for the vertical groups; ROOT goes 0.05 → 0.20 MiB, still tiny), so reader cost is unchanged.

Manifest size targets and the count-vs-size tradeoff are anchored on empirical data from the in-progress backfill: ~9 bytes/ref, max full manifest 2.56 MiB.

## How

- Extend `manifest_append_dim_split` to accept either a single `int` (uniform — existing callers unchanged) or a `{path_regex: size, None: catch_all}` mapping. The `None` catch-all is rendered as an `AnyArray` rule **last**, since icechunk matches conditions first-to-last per axis (`session.rs` `get_split_sizes`).
- HRRR-spatial config: `{r"^/(model_level|pressure_level)/": 120, None: 480}`.
- New doc section in `docs/virtual_datasets.md` on manifest splitting (the size-vs-count tradeoff and per-group sizing).

`split_size` is fixed at store creation, so this takes effect only on a fresh dataset version.

Part of #700.


docs/virtual_datasets.md
```diff
@@ -64,6 +64,15 @@ Two operational rules when backfilling a live virtual dataset:
...
 - **Choose `append_dim_end` as the last *fully published* position, not "now".** Finalize resets `main` to the pre-sized branch, so positions past the published data would appear as NaN-filled slots to readers.
 
+## Manifest splitting
+
+Icechunk stores each array's chunk references in one or more manifest objects, split along a dimension via `IcechunkVirtualConfig.manifest_split` (built with `manifest_append_dim_split`). Two independent costs pull in opposite directions:
+
+- **Manifest size (bytes)** — a reader downloads a whole manifest to resolve *any* chunk in it, and a commit rewrites the active split's whole manifest each time it grows. Keep a full manifest within a reader-friendly size (roughly ≤ 5 MiB; a dense HRRR-spatial vertical manifest is ~2.5 MiB). Larger split → bigger manifests.
+- **Total split count `M`** — `M = array_count × ceil(appends / split_size)`. Every commit's cost grows with `M` (it re-serializes the snapshot's manifest list and touches every split of every array). Smaller split → more splits → larger `M`. For a many-variable dataset `M` is dominated by `array_count`, so over-splitting the *sparse* arrays is the biggest, least useful contributor.
```
aldenks: **Total manifest count `M`**

in this whole section "sparse" isn't quite right, its small or lower ref count

docs/virtual_datasets.md
```diff
@@ -64,6 +64,15 @@ Two operational rules when backfilling a live virtual dataset:
...
+
+These two are set per array, so size them per array group: give sparse single-level arrays a **coarse** split (few splits, still-small manifests) and dense vertical-group arrays a **finer** split (more splits, but each manifest stays within the reader budget). `manifest_append_dim_split` takes either one `split_size` for all arrays or a `{path_regex: size, None: catch_all_size}` mapping for this per-group policy. `split_size` is fixed at store creation — changing it needs a new dataset version.
```
aldenks: split size can be resized in icechunk, remove that sentence

docs/virtual_datasets.md
aldenks: Fixed in 4677ef45 — replaced 'sparse' with 'low ref count' throughout the section (and the code comments/docstring), and clarified that a manifest's size is `split_size × refs_per_index × bytes_per_ref`.

docs/virtual_datasets.md
aldenks: Good catch — removed that sentence in 4677ef45. (Also means we may be able to resize the split on the existing v0.3.0 store rather than needing a fresh version — noted.)

docs/virtual_datasets.md
aldenks: Also renamed 'split count' → 'manifest count' (M) in 6d3adc82 per the suggested heading.

PR #708, opened 2026-07-02
Assign virtual backfill workers contiguous append-dim blocks

aldenks:
## Problem

Virtual icechunk backfill flushes slow down dramatically as the archive fills (measured: 38s per flush on an empty archive -> 190s at 62% full, amplified ~6x by rebase-on-conflict). References #700.

## Mechanism

Each worker's `jobs_per_pod` region jobs are scattered across the whole append dim: `RegionJob.get_jobs` reorders regions with `spread_evenly` (bit-reversal) and `get_worker_jobs` assigns round-robin by stride. Icechunk manifests are split into immutable windows along the append dim, and a commit read-modify-writes every window its refs touch — 30 scattered inits touch ~26 of 97 windows per array x 177 arrays ≈ ~3,400 manifest rewrites per flush, with rewrite bytes growing as the archive fills. A contiguous block of inits touches 1-2 windows per array (~200 rewrites) and rewrite bytes stay bounded by one window regardless of archive fill.

Spreading exists to avoid S3 SlowDown hot-spotting on source-data prefixes when many workers download GRIBs from one date range (docs/parallel_processing.md). Virtual datasets never download the GRIBs — only tiny .idx files and prefix listings — so that rationale doesn't apply to them. It still applies to materialized datasets, which keep spreading.

## Change

- `RegionJob.worker_assignment: ClassVar[Literal["spread", "contiguous"]] = "spread"`; `VirtualRegionJob` overrides to `"contiguous"`.
- `RegionJob.get_jobs` only applies `spread_evenly` when the policy is `"spread"`; contiguous classes keep regions in append-dim order.
- New `iterating.get_contiguous_worker_jobs`: worker w gets block `jobs[w*k : (w+1)*k]` with `k = ceil(n / workers_total)`; asserts the block is non-empty (guaranteed since callers derive `workers_total = ceil(n_jobs / jobs_per_pod)`).
- `DynamicalDataset._process_region_jobs` dispatches stride vs contiguous on `self.region_job_class.worker_assignment`.
- Materialized behavior is byte-for-byte unchanged; virtual operational updates (single one-job worker) are unaffected.
- Docs: parallel_processing.md "Append dim region spreading and worker assignment", virtual_datasets.md Backfill section, AGENTS.md parallelization model.

## Tests

- `get_contiguous_worker_jobs`: exact blocks, disjoint/complete/ordered partition across awkward n/jobs_per_pod combos (11641/30, 7/3, ...), invalid-index and empty-block asserts.
- Virtual `get_jobs` returns regions in append-dim order (not spread); a virtual worker's assigned jobs are contiguous along the append dim.
- Materialized `get_jobs` still returns regions in `spread_evenly` order.
- `uv run pytest tests/common/ -k "iterating or region_job or virtual" -q` (172 passed) and `uv run pytest tests/noaa/hrrr/forecast_48_hour_spatial/ -q` (24 passed); ruff format/check and ty pass.

https://claude.ai/code/session_014e3bGfVzk8raawnczFMZ4D


aldenks:
Addressed review feedback (and rebased onto main, picking up #707/#709):

1. **Single `get_worker_jobs` entry point.** Removed the separate public `get_contiguous_worker_jobs`. `iterating.get_worker_jobs` now takes a required keyword-only `worker_assignment: Literal["spread", "contiguous"]`, holds the shared index-bounds assertions, and dispatches to private `_round_robin_worker_jobs` / `_contiguous_worker_jobs` helpers. `DynamicalDataset._process_region_jobs` makes one call passing `worker_assignment=self.region_job_class.worker_assignment` — no if/else. Tests, AGENTS.md, and docs/parallel_processing.md all route through the single entry point; test coverage is unchanged (round-robin behavior, contiguous disjoint/complete/ordered/non-empty partition properties, invalid-index asserts).

2. **Corrected the hot-spotting overclaim** in docs/parallel_processing.md: the virtual paragraph now says contiguous assignment concentrates .idx reads and prefix listings on adjacent source prefixes, which can still hot-spot; aligning workers with the manifest split structure matters more since virtual workers fetch only small index files rather than the data files.

https://claude.ai/code/session_014e3bGfVzk8raawnczFMZ4D

aldenks:
Second round of review feedback addressed: the assignment policy now lives entirely in `iterating.get_worker_jobs` — each mode is a self-contained permutation + selection over the canonical append-dim-ordered job list.

- `RegionJob.get_jobs` no longer conditionally applies `spread_evenly`; it always returns jobs in append-dim order (regions in order, region-major over var groups).
- `get_worker_jobs` `"spread"` mode now does `spread_evenly(jobs)` then round-robin stride (`_spread_worker_jobs`); `"contiguous"` is unchanged (identity order, block slice). Docstrings on `get_worker_jobs` and `spread_evenly` updated to say each mode owns its ordering.
- `docs/parallel_processing.md` materialized paragraph now says the spread permutation is applied during worker assignment, not before the job list is built; AGENTS.md bullet matched.
- Tests: `get_jobs` order tests assert canonical append-dim order; the spread property moved to `get_worker_jobs(..., worker_assignment="spread")` output (round-robin over `spread_evenly` order, plus a concurrent-window-covers-the-range property test). Disjoint/complete partition properties are tested for both modes through the single entry point; contiguous tests unchanged. Per reviewer-approved nuance, no test pins the old regions-before-product ordering.

Note: this branch picked up a commit from another session ("Document backfill parallelism ceiling and rebase starvation tail"); it's preserved as its own commit on top of the feature commit.

Checks: ruff format/check, ty clean; `tests/common -k "iterating or region_job or virtual"` 170 passed; `tests/noaa/hrrr/forecast_48_hour_spatial/` 24 passed.

https://claude.ai/code/session_014e3bGfVzk8raawnczFMZ4D

PR #717, opened 2026-07-07
Add --metadata-only to backfill for existing virtual stores

aldenks:
## Motivation

When a template change touches only **metadata and coordinates** — new `GribberishCodec` options (e.g. `north_up`), reordered coordinates, changed attrs — but not the chunk data, there was no supported way to deploy it to an already-published virtual dataset:

- An operational `-update` **can't** do it: updates never rewrite existing coordinates or encoding, and a reordered dimension coordinate makes the update's drift check (`assert_no_structural_drift_from_existing_store`) **fail**.
- A full re-backfill is **wasteful**: the virtual chunk refs point at GRIB byte ranges and decode flags apply at read time, so the (millions of) refs are unchanged and don't need re-emitting.

This came up deploying #716 (gribberish `north_up`) to the live HRRR/GEFS spatial stores.

## Approach

The insight: the icechunk backfill **already** does exactly the needed metadata overwrite in `parallel_setup`/`finalize` — temp branch from `main` → `copy_zarr_metadata` (rewrites coordinate arrays + every array's `zarr.json`, keeping existing chunks) → reset `main`. So a metadata-only refresh is just **a backfill that processes zero jobs**.

`backfill-kubernetes --metadata-only` runs a single pod that generates no jobs; setup + finalize rewrite coordinates and encoding and publish to `main`, leaving the chunk manifest untouched. No new write logic — only a flag that skips job generation and runs one pod.

```
uv run main <dataset> backfill-kubernetes <append_dim_end> <jobs_per_pod> <max_parallelism> --metadata-only
```

## Safety: append-dim truncation guard

Because `append_dim_end` sizes the written metadata, a too-small value would shrink the array and orphan published chunks. Added `_assert_append_extent_not_truncated`, run on worker 0 for every backfill into an existing store:

- `--metadata-only`: template extent must **match** the store exactly (a pure refresh).
- any other backfill (incl. `--overwrite-existing`): may grow but must **not shrink** the store.
- fresh backfill (no store yet): nothing to protect, skipped.

## Tests

- `test_metadata_only_backfill_rewrites_coords_keeping_refs`: backfill emits refs; a metadata-only run with reordered latitude + a changed data-var attr propagates the new coordinate/metadata while the chunk values remain readable and unchanged.
- `test_metadata_only_backfill_rejects_extent_mismatch` and `test_backfill_rejects_append_extent_truncation`: the guard fires for exact-mismatch and shrink.

Full fast suite passes (1514) and the virtual + coordination suites (82) pass. Docs: new "Metadata-only backfills" subsection in `docs/virtual_datasets.md`.


aldenks:
too much complexity and needs alignment with --overwrite-existing option. i also want to work in automatic append_dim_end selection for metadata only and chunk overwrite only options

PR #738, opened 2026-07-12
Retry only failed keys in _exists_many virtual chunk probe

aldenks:
## Problem
A whole-archive virtual manifest scan (`availability` on a virtual store, and the completeness gate in `run-all`) issues enough `store.exists` probes — `manifest_scan` flushes batches of up to `_EXISTS_BATCH_SIZE` (20,000) keys — that a transient object-store error during the run is near-certain. `_exists_many` gathered the whole batch with `asyncio.gather` **without** `return_exceptions`, so a single failed probe raised and the surrounding `retry()` re-issued the **entire** batch. Re-firing thousands of concurrent probes tends to hit another transient failure, so the retries can keep failing and eventually exhaust — killing a multi-hour run. (Observed: a full `noaa-hrrr-forecast-48-hour-virtual` run died ~2h in on a single `HTTP connect timeout` while probing `pressure_level/relative_humidity`.)

## Change
- `_exists_many` now gathers with `return_exceptions=True` and retries **only the keys that failed**, with backoff, up to `max_attempts`. A transient blip on one key no longer discards the successful probes or re-issues the whole batch.
- `manifest_scan._flush_var_probes` calls `_exists_many(store, keys, max_attempts=_SCAN_MAX_RETRIES)` directly instead of wrapping it in an outer whole-batch `retry(...)`.
- The write path (`filter_already_present` → `_exists_many`) gets the same per-key resilience for free.

## Testing
- New unit tests: `_exists_many` re-probes only the failed key (not the whole batch) and returns correct results; and raises after exhausting attempts when a key keeps failing.
- `uv run ruff check && uv run ty check` clean; `pytest -k "manifest or exists or completeness or decode_health"` → 22 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y29NPqgcTDZd5dQKeuUnp2)_

### DIFF
--- a/src/reformatters/common/virtual_region_job.py
+++ b/src/reformatters/common/virtual_region_job.py
@@ -626,12 +626,43 @@ class _NoRegion:
 _NO_REGION: Final = _NoRegion()
 
 
-def _exists_many(store: IcechunkStore, keys: Sequence[str]) -> dict[str, bool]:
-    """Probe many chunk keys concurrently (store.exists is async)."""
+def _exists_many(
+    store: IcechunkStore, keys: Sequence[str], *, max_attempts: int = 8
+) -> dict[str, bool]:
+    """Probe many chunk keys concurrently (store.exists is async).
+
+    Transient object-store errors are retried per key: each attempt re-probes only the
+    keys that still failed, so one flaky probe never discards the successful probes or
+    forces the whole batch to be re-issued. A whole-archive scan issues enough probes
+    that a transient failure is near-certain, so retrying the failed subset (rather than
+    the entire batch) is what keeps a single blip from failing the run."""
     if not keys:
         return {}
 
-    async def _check() -> list[bool]:
-        return list(await asyncio.gather(*(store.exists(key) for key in keys)))
+    async def _check(probe_keys: Sequence[str]) -> list[bool | BaseException]:
+        return list(
+            await asyncio.gather(
+                *(store.exists(key) for key in probe_keys), return_exceptions=True
+            )
+        )
 
-    return dict(zip(keys, asyncio.run(_check()), strict=True))
+    result: dict[str, bool] = {}
+    pending: Sequence[str] = keys
+    last_exception: BaseException | None = None
+    for attempt in range(max_attempts):
+        failed: list[str] = []
+        for key, outcome in zip(pending, asyncio.run(_check(pending)), strict=True):
+            if isinstance(outcome, BaseException):
+                last_exception = outcome
+                failed.append(key)
+            else:
+                result[key] = outcome
+        if not failed:
+            return result
+        pending = failed
+        if attempt < max_attempts - 1:
+            rng = np.random.default_rng()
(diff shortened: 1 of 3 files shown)


src/reformatters/common/virtual_region_job.py
```diff
@@ -626,12 +626,43 @@ def __getattr__(self, name: str) -> Any:  # noqa: ANN401
...
+def _exists_many(
+    store: IcechunkStore, keys: Sequence[str], *, max_attempts: int = 8
+) -> dict[str, bool]:
+    """Probe many chunk keys concurrently (store.exists is async).
+
+    Transient object-store errors are retried per key: each attempt re-probes only the
+    keys that still failed, so one flaky probe never discards the successful probes or
+    forces the whole batch to be re-issued. A whole-archive scan issues enough probes
+    that a transient failure is near-certain, so retrying the failed subset (rather than
+    the entire batch) is what keeps a single blip from failing the run."""
```
```suggestion
    """Probe many chunk keys concurrently."""
```

src/reformatters/common/virtual_region_job.py
```diff
@@ -626,12 +626,43 @@ def __getattr__(self, name: str) -> Any:  # noqa: ANN401
...
+                last_exception = outcome
+                failed.append(key)
+            else:
+                result[key] = outcome
+        if not failed:
+            return result
+        pending = failed
+        if attempt < max_attempts - 1:
+            rng = np.random.default_rng()
+            time.sleep(attempt * rng.uniform(0.8, 1.2) + 0.1)
```
```suggestion
            time.sleep((attempt - 1) * rng.uniform(0.8, 1.2) + 0.1)
```
Would this work? I don't want prod update retries to wait long other first retry attempt

src/scripts/validation/manifest_scan.py
```diff
@@ -308,7 +308,9 @@ def _flush_var_probes(
     out: dict[str, dict[pd.Timestamp, bool]],
 ) -> None:
     keys = [key for _, _, key in probes]
-    present = retry(lambda: _exists_many(store, keys), max_attempts=_SCAN_MAX_RETRIES)
+    # _exists_many retries the failed subset internally, so a transient blip on one of
+    # these (up to _EXISTS_BATCH_SIZE) keys no longer re-probes the whole batch.
```
aldenks: Remove unnecessary comment

src/reformatters/common/virtual_region_job.py
aldenks: The literal `(attempt - 1)` sleeps a negative duration on the first retry — `attempt` is 0-indexed, so before the first retry it's `(0 - 1) * ~1.0 + 0.1 ≈ -0.9`, which `time.sleep` rejects with a ValueError. I pushed `max(0, attempt - 1) * rng.uniform(0.8, 1.2) + 0.1`, which does what you want: the first retry waits ~0.1s and back-off only kicks in from the second retry onward (0.1s, 0.1s, ~1.1s, ~2.2s, …).

---

PR #746, opened 2026-07-15
Add VBDSF (direct/beam solar radiation) to NOAA HRRR

mrshll:
## What

Adds `visible_beam_downward_solar_flux_surface` (GRIB element **VBDSF** — Visible Beam Downward Solar Flux, the direct/beam shortwave irradiance) to the shared materialized HRRR data-variable config (`NoaaHrrrCommonTemplateConfig.get_data_vars`). Because that method is shared, both materialized HRRR datasets gain the variable:

- `noaa-hrrr-forecast-48-hour`
- `noaa-hrrr-analysis`

The new `DataVar` is placed right after the existing DSWRF entry (`downward_short_wave_radiation_flux_surface`) and mirrors its internal pattern (`units="W m-2"`, `step_type="instant"`, `hrrr_file_type="sfc"`, `keep_mantissa_bits=7`). The f00 (analysis-time) VBDSF record is not properly diagnosed in the source (non-physical spikes >9000 W m-2), so it carries `hour_0_values_override=False` — the same treatment as the other unreliable-at-f00 instant surface fields.

## Why

Customer solar use case: DSWRF already provides **global** (total) surface downward shortwave irradiance; VBDSF adds the **direct/beam** component, which downstream solar models need alongside the global value.

## Consistency

Name and externally-visible attrs exactly match the existing HRRR **virtual** dataset's VBDSF (`short_name="vbdsf"`, `long_name="Visible Beam Downward Solar Flux"`, `units="W m-2"`, no `standard_name`), so the two HRRR products stay in sync. Per AGENTS.md, VBDSF (a visible-band beam flux) has no exact CF standard name, so `standard_name` is left unset — it is already listed in `ALLOWED_MISSING_STANDARD_NAME` in the CF compliance test. The virtual dataset is not touched (it already ships this variable).

## Verification

Verified on the canonical Python 3.14.6 toolchain: `update-template` for both datasets produces **zero diff** vs. the checked-in metadata; `ruff` / `ty check` / `pytest` (`common_template_config_subclasses_test` + `datasets_cf_compliance_test`, 259 passed) all pass; and a decode of real source GRIB confirmed physically sensible values. See the verification comment below for full detail, including the f00 fix.

## Scope

Template metadata only. Post-merge work — history backfill, validation (incl. the VBDSF version-boundary note), and catalog docs — is tracked in #750.

https://claude.ai/code/session_015ZY5VagHhMYq9D6WXVXxvq

### DIFF
--- a/src/reformatters/noaa/hrrr/template_config.py
+++ b/src/reformatters/noaa/hrrr/template_config.py
@@ -323,9 +323,6 @@ class NoaaHrrrCommonTemplateConfig(TemplateConfig[NoaaHrrrDataVar]):
                     index_position=128,
                     keep_mantissa_bits=default_keep_mantissa_bits,
                     hrrr_file_type="sfc",
-                    # HRRR f00 VBDSF (the analysis-time record) is not properly diagnosed
-                    # and contains non-physical spikes (>9000 W m-2); use the f01 value.
-                    hour_0_values_override=False,
                 ),
             ),
             NoaaHrrrDataVar(
--- a/src/reformatters/noaa/hrrr/template_config.py
+++ b/src/reformatters/noaa/hrrr/template_config.py
@@ -308,6 +308,24 @@ class NoaaHrrrCommonTemplateConfig(TemplateConfig[NoaaHrrrDataVar]):
                 ),
             ),
             NoaaHrrrDataVar(
+                name="visible_beam_downward_solar_flux_surface",
+                encoding=encoding,
+                attrs=DataVarAttrs(
+                    short_name="vbdsf",
+                    long_name="Visible Beam Downward Solar Flux",
+                    units="W m-2",
+                    step_type="instant",
+                ),
+                internal_attrs=NoaaHrrrInternalAttrs(
+                    grib_element="VBDSF",
+                    grib_description='0[-] SFC="Ground or water surface"',
+                    grib_index_level="surface",
+                    index_position=128,
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                    hrrr_file_type="sfc",
+                ),
+            ),
+            NoaaHrrrDataVar(
                 name="downward_long_wave_radiation_flux_surface",
                 encoding=encoding,
                 attrs=DataVarAttrs(
--- a/src/reformatters/noaa/hrrr/template_config.py
+++ b/src/reformatters/noaa/hrrr/template_config.py
@@ -323,6 +323,9 @@ class NoaaHrrrCommonTemplateConfig(TemplateConfig[NoaaHrrrDataVar]):
                     index_position=128,
                     keep_mantissa_bits=default_keep_mantissa_bits,
                     hrrr_file_type="sfc",
+                    # HRRR f00 VBDSF (the analysis-time record) is not properly diagnosed
+                    # and contains non-physical spikes (>9000 W m-2); use the f01 value.
+                    hour_0_values_override=False,
                 ),
             ),
             NoaaHrrrDataVar(
--- a/src/reformatters/noaa/hrrr/analysis/templates/latest.zarr/visible_beam_downward_solar_flux_surface/zarr.json
+++ b/src/reformatters/noaa/hrrr/analysis/templates/latest.zarr/visible_beam_downward_solar_flux_surface/zarr.json
@@ -1,0 +1,83 @@
+{
+  "shape": [
+    1,
+    1059,
+    1799
+  ],
+  "data_type": "float32",
+  "chunk_grid": {
+    "name": "regular",
+    "configuration": {
+      "chunk_shape": [
+        2160,
+        540,
+        450
+      ]
+    }
+  },
+  "chunk_key_encoding": {
+    "name": "default",
+    "configuration": {
+      "separator": "/"
+    }
+  },
+  "fill_value": "NaN",
+  "codecs": [
+    {
+      "name": "sharding_indexed",
+      "configuration": {
+        "chunk_shape": [
+          2160,
+          45,
+          45
+        ],
+        "codecs": [
+          {
+            "name": "bytes",
+            "configuration": {
+              "endian": "little"
+            }
+          },
+          {
+            "name": "blosc",
+            "configuration": {
+              "typesize": 4,
+              "cname": "zstd",
+              "clevel": 3,
+              "shuffle": "shuffle",
+              "blocksize": 0
+            }
+          }
+        ],
+        "index_codecs": [
+          {
+            "name": "bytes",
+            "configuration": {
+              "endian": "little"
+            }
+          },
+          {
+            "name": "crc32c"
+          }
+        ],
+        "index_location": "end"
+      }
+    }
+  ],
+  "attributes": {
+    "long_name": "Visible Beam Downward Solar Flux",
+    "short_name": "vbdsf",
+    "units": "W m-2",
+    "step_type": "instant",
+    "coordinates": "latitude longitude spatial_ref",
+    "_FillValue": "AAAAAAAA+H8="
+  },
+  "dimension_names": [
+    "time",
+    "y",
+    "x"
+  ],
+  "zarr_format": 3,
+  "node_type": "array",
+  "storage_transformers": []
+}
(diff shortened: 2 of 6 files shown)


mrshll:
## Verification on canonical Python 3.14 toolchain

Re-verified on real 3.14.6 (`uv sync`, template regen, ruff/ty/pytest, and decode of real source GRIB).

- **Template diff**: regenerated `noaa-hrrr-forecast-48-hour` + `noaa-hrrr-analysis` → **zero diff** vs. checked-in metadata.
- **ruff format / ruff check / ty check**: all clean.
- **`common_template_config_subclasses_test` + `datasets_cf_compliance_test`**: 259 passed.
- **Decode (init 2018-07-13T12z, HRRR sfc)**: VBDSF `units=W m-2`, non-negative, no NaN — but see the correction below.

### Correction pushed: skip the garbage f00 record
Decode-verify surfaced that HRRR's **f00 (analysis-time) VBDSF record is not properly diagnosed** — it contains non-physical spikes up to **~9944 W m-2** (~7x the solar constant) in ~1300 CONUS cells. f01+ decodes cleanly (max ~930 W m-2), and DSWRF at f00 is fine, so this is VBDSF-specific to the analysis record, not a config/index error.

Because `step_type="instant"` makes `has_hour_0_values` true, the original config wrote f00 into the forecast **and** made the analysis dataset pull VBDSF from f00 — baking the spikes into every analysis timestep. I set **`hour_0_values_override=False`** (the same treatment the other unreliable-at-f00 instant surface fields already use). After the fix: forecast f00 → NaN, remaining leads clean (max ~930 W m-2); template stays zero-diff (internal attr only).

### For the maintainer — analysis backfill + validation note
VBDSF only exists in HRRR from **~2016-08-23 (HRRRv2)** onward (verified against the S3 `.idx`: absent 2016-08-23 and earlier, present 2016-09-01+). The HRRR **analysis** archive starts 2014-10-01, so a full historical backfill will leave VBDSF **NaN for ~2014-10-01 → ~2016-08-22** and log per-file "found 0 matching bands" read failures over that window (non-fatal → NaN fill). The **forecast** dataset starts 2018-07-13 (HRRRv3) and is unaffected. This version-boundary window should be recorded in the dataset's `validation_summary.md` `### Review notes` during the step-3 validation pass.

_Automated verification; PR left as draft for human review._

src/reformatters/noaa/hrrr/template_config.py
```diff
@@ -307,6 +307,27 @@ def get_data_vars(self, encoding: Encoding) -> Sequence[NoaaHrrrDataVar]:
...
+                    step_type="instant",
+                ),
+                internal_attrs=NoaaHrrrInternalAttrs(
+                    grib_element="VBDSF",
+                    grib_description='0[-] SFC="Ground or water surface"',
+                    grib_index_level="surface",
+                    index_position=128,
+                    keep_mantissa_bits=default_keep_mantissa_bits,
+                    hrrr_file_type="sfc",
+                    # HRRR f00 VBDSF (the analysis-time record) is not properly diagnosed
```
mrshll: @aldenks this is the main part I want your eyes on

aldenks: double check
- sample a handful of inits per era of hrrr (v2, v3, v4, etc) to see how pervasive it is
- this is a known thing mentioned on the web -- may still be real even if it isn't but a helpful cross check

if it is mostly garbage we want to keep this as it is (hour_0_values_override=False will leave hour 0 nan for all forecasts for this var.) If it's scattered or only showed up in older data, we will want to keep the data in (remove hour_0_values_override), and document it in the validation report.

mrshll: Ran the double-check with real byte-range GRIB requests against the `noaa-hrrr-bdp-pds` S3 archive, decoding the actual VBDSF:surface record at f00 vs f01 with rasterio across eras and init hours.

**The f00 spikes are real, systematic, and present in every era — including current v4. Not scattered, not historical-only.**

f00 at 12z inits (max W/m², cells >1500, cells >9000) vs the always-clean f01:

| Era | Init | f00 max | f00 >1500 | f00 >9000 | f01 max |
|-----|------|--------:|----------:|----------:|--------:|
| v2 | 2017-07-15 12z | 7299 | 662 | 0 | 851 |
| v3 | 2018-07-13 12z | 9944 | 1226 | 15 | 891 |
| v3 | 2018-07-20 12z | 9946 | 1323 | 24 | 900 |
| v4 | 2020-12-02 12z | 9851 | 995 | 17 | 835 |
| v4 | 2021-01-01 12z | 9977 | 1038 | 18 | 829 |
| v4 | 2023-07-15 12z | 9968 | 1192 | 22 | 883 |
| v4 | 2024-07-15 12z | 9941 | 1105 | 18 | 877 |
| v4 | 2025-06-01 12z | 9823 | 1064 | 15 | 858 |

Peaks hit ~10,000 W/m² (≈7× the solar constant) in hundreds-to-1000+ CONUS cells in every year sampled 2017→2025. f01 is clean everywhere (≤~920 W/m²).

**It tracks solar elevation, not the archive version.** Concentrated at low-sun init hours over CONUS: 06z inits are all-zero (night, no spikes); 10z–13z affected (worst near 12z); 14z and 18z f00 clean. Since the override is per-variable all-or-nothing, this doesn't change the call. (Also: VBDSF is entirely absent at f00 in HRRRv1 / 2015.)

**Verdict:** it's mostly garbage → keep `hour_0_values_override=False` as-is. With it, the analysis dataset pulls VBDSF from f01 (always clean) and the forecast drops lead 0. No "scattered/historical-only" validation note needed. The separate pre-2016-08-23 absence note (VBDSF didn't exist yet, so the analysis backfill NaNs that window) is unrelated and still stands.

---

mrshll: Correction on my "mostly garbage" wording above — that overstates it. Decoding a representative case in full (v4, 2024-07-15 12z, the 1.9M-cell CONUS grid):

- The f00 field is **99.92% physically sensible.** Everything up through the 99.9th percentile is ≤949 W/m² — a normal morning solar field.
- The defect is the top **0.076% — ~1,442 cells** — ramping smoothly from 1,200 up to **9,941 W/m², i.e. 7.3× the solar constant** (1,361 W/m² is the top-of-atmosphere max; clear-sky surface tops out ~1,000–1,100). Impossible, but sparse.
- Spatially it's a scatter, not a wipeout: 102 clusters, 33 of them isolated single pixels, largest patch 793 cells.
- Those same cells at f01 read 0–660 W/m² (mean 474) — fully physical. So it's an analysis-instant diagnostic artifact that clears after 1h of integration.

So it's **not "mostly garbage" — it's a mostly-good field with a small set of physically-impossible outliers**, present at every low-sun init (~10z–13z) across all eras (not historical, not one-off; per-init bad-cell counts run ~500–1,300).

That makes the call a real tradeoff rather than the clean "garbage → drop it" binary:

- **Keep `hour_0_values_override=False` (current PR):** lead-0 never carries an impossible value — analysis substitutes the clean f01, forecast drops lead-0. Cost: you also throw away the 99.9%-good f00 field.
- **Remove it:** keep the 99.9%-good f00, but bake in ~500–1,300 impossible point values per low-sun init, caught only by a validation-report note.

For the solar use case that motivated VBDSF, where point queries matter, I still lean toward keeping the override — a handful of 7×-solar-constant values reaching a downstream solar model is worse than a missing analysis-time step — but it's a judgment call, not a foregone conclusion.

---

mrshll: Lol

aldenks: ha. so keep hour 0 in and remove this override and this comment. This will show up in the validation report

aldenks (approved):
Approved after removing `hour_0_values_override=False`

src/reformatters/noaa/hrrr/template_config.py
mrshll: Done — removed both `hour_0_values_override=False` and the comment, so VBDSF now keeps its f00 analysis-time record (forecast keeps lead-0, analysis pulls VBDSF from f00). The f00 diagnostic artifact will be documented in the validation report during the step-3 validation pass.

`update-template` for both HRRR datasets is zero-diff (the override is an internal attr, not serialized); ruff/ty/pytest all pass.

The fix is in #793, which targets this branch (`claude/hrrr-add-vbdsf`) so it's a single-commit diff — merging it here updates this PR.

---

PR #755, opened 2026-07-19
Safer, simpler backfills: explicit overwrite API, auto append-dim end, manual GitHub action

aldenks:
backfill-kubernetes now defaults --append-dim-end to "now" for a new store
and to the store's current end for an existing one, and replaces
--overwrite-existing with explicit --overwrite-chunks / --overwrite-metadata
flags. Guards (assert_safe_overwrite, checked in the driver and again on
worker 0) reject structural drift, never allow trimming an existing store,
and only allow new arrays or extension when explicitly requested.
--overwrite-metadata alone refreshes store metadata from the template in
place via the shared refresh_store_metadata helper (extracted from
VirtualRegionJob.refresh_metadata), creating newly added variables without
launching workers.

Overwrite backfills preserve job-written coordinate values (template-null
coords like ingested_forecast_length are excluded from metadata copies),
publish zarr3 metadata at finalize so a new variable appears only once its
data is written, and suspend the dataset's update/validate cronjobs so an
operational update can't move icechunk main and strand the backfill's
finalize branch reset — which now raises loudly (with correct crash-retry
detection) instead of silently skipping publication.

New workflow_dispatch GitHub actions, generated alongside the existing
manual workflows: "Manual: Backfill" exposing only the safe operations
(create new store, overwrite chunks/metadata) using the deployed image, and
"Manual: Suspend/Resume Dataset CronJobs" to resume updates after a
backfill.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01SAB2gwVGczEvJGx4d9DnMm


src/reformatters/common/dynamical_dataset.py
```diff
@@ -194,21 +196,71 @@ def update(
...
+        explicit --append-dim-end past the store's current end, to extend the store.
+        Trimming an existing store is never supported.
+
+        --append-dim-end defaults to now for a new store and to the store's current
+        end for an existing store (leaving its extent unchanged).
+
+        An operational update that publishes to an icechunk main while an overwrite
+        backfill runs makes the backfill's finalize fail loudly and the backfill must
+        be re-run, so run overwrite backfills in between update runs (see
+        "Concurrent jobs writing to the same dataset" in docs/parallel_processing.md).
```
aldenks: remove and refer to Backfills section in AGENTS.md

AGENTS.md
```diff
@@ -4,7 +4,7 @@ This project contains code to reformat weather data into the Zarr v3 / Icechunk
...
-2. A zarr backfill is run. The backfill uses kubernetes indexed jobs to run work in parallel. When the user runs a `uv run main <dataset-id> backfill-kubernetes ...` command the metadata for the zarr is first written by the local process to the final zarr store, then a kubernetes index job is kicked off with each job index responsible for writing a portion of the zarr chunk data into the zarr archive.
+2. A zarr backfill is run. The backfill uses kubernetes indexed jobs to run work in parallel. A `uv run main <dataset-id> backfill-kubernetes ...` command (or the "Manual: Backfill" GitHub action) validates the request against the existing store, then kicks off a kubernetes indexed job with each job index responsible for writing a portion of the zarr chunk data into the zarr archive. With no flags it creates a new store (ending at "now" by default); `--overwrite-chunks` / `--overwrite-metadata` write into an existing store with guards that never trim it and only extend it on an explicit `--append-dim-end`.
```
aldenks: Remove "With no flags..." sentence. Make 

```suggestion
2. A zarr backfill is run. A backfill validates the request against the existing store, then kicks off a kubernetes indexed job with each job index responsible for writing a portion of the zarr chunk data into the zarr archive.
```

PR #787, opened 2026-07-23
Better Stack heartbeats mirroring cron monitoring (Sentry migration, 3/3)

mrshll:
Third of three sequenced PRs migrating observability from Sentry to Better Stack. **Stacked on #786** (base retargets as the stack merges).

## What

Better Stack uptime **heartbeats** for update/validate crons, pinged *alongside* the existing Sentry cron check-ins — so cron monitoring is double-covered during the migration and revert is just reverting this PR.

- **`common/betterstack.py`** — `monitor_heartbeat` pings a **start** heartbeat when a run begins and a **complete** heartbeat when it finishes (`/fail` on error). Best-effort (`httpx` errors are swallowed + logged) so a Better Stack outage never breaks a run.
- **`common/deploy.py`** — deploy-time provisioning that **mirrors the Sentry monitor config**: period from the cron schedule (via `croniter`), start grace = the 10-min Sentry `checkin_margin`, complete grace = margin + `pod_active_deadline`. `reconcile_heartbeats` idempotently creates/updates monitors via the Better Stack Uptime API; `write_heartbeat_secret` merges the url map into a k8s secret (so a single-dataset deploy doesn't drop others).
- **`common/kubernetes.py`** — `BETTERSTACK_HEARTBEATS_SECRET_NAME`; update/validate crons mount the url-map secret via a new `HeartbeatCronJob` base's `mounted_secret_names()`.
- **`dynamical_dataset._monitor`** — enters `monitor_heartbeat` alongside `monitoring.monitor_cron`.
- Deploy workflow passes `BETTERSTACK_API_KEY_RW` (required to provision).

## ⚠️ Deploy ordering

Deploy **this PR before #786 (errors DSN)**. Once #786 points the DSN at Better Stack, Sentry cron check-ins stop reaching Sentry, so heartbeats must already be live. Order: #785 (logging) → this (heartbeats) → #786 (errors).

## Scope notes

- **Staging heartbeats deferred** to a follow-up: `deploy_staging` passes `provision_heartbeats=False`, so staging deploys don't need the API key.
- Provisioning + runtime ping are covered by unit tests (schedule→period, spec grace mirroring, reconcile create/update via mocked API, secret merge, start/complete/fail pings). The Uptime API request/response shape is inherited from the prior heartbeats work (#685); the first prod deploy with `BETTERSTACK_API_KEY_RW` is the live check.

## Prerequisite

Add the `BETTERSTACK_API_KEY_RW` GitHub Actions secret (read/write Uptime API token).

## Revert runbook

Revert this PR — Sentry crons remain live throughout, so cron monitoring is never lost.

### DIFF
--- a/src/reformatters/common/deploy.py
+++ b/src/reformatters/common/deploy.py
@@ -5,7 +5,7 @@ from typing import Any
 
 import typer
 
-from reformatters.common import docker, kubernetes, staging
+from reformatters.common import betterstack, docker, kubernetes, staging
 from reformatters.common.dynamical_dataset import DynamicalDataset
 from reformatters.common.logging import get_logger
 
@@ -42,6 +42,11 @@ def deploy_operational_resources(
 
     assert len(reformat_jobs) > 0, "No cronjobs to deploy" + (
         f" for dataset_id_filter={dataset_id_filter!r}" if dataset_id_filter else ""
+    )
+
+    # Provision heartbeats (and write their secret) before the cron pods that mount it.
+    betterstack.provision_heartbeats(
+        cj for cj in reformat_jobs if isinstance(cj, kubernetes.CronJob)
     )
 
     k8s_resource_list = {
@@ -103,14 +108,17 @@ def register_commands(
         version: str,
         force: bool = False,
     ) -> None:
-        """Clean up staging resources: kubernetes cronjobs and git branch."""
+        """Clean up staging resources: Better Stack heartbeats, kubernetes cronjobs and git branch."""
         staging.find_dataset(datasets, dataset_id)  # validate dataset_id
         if not force:
             cronjob_names = staging.staging_cronjob_names(dataset_id, version)
             branch = staging.staging_branch_name(dataset_id, version)
             log.info(
-                f"Will delete cronjobs {cronjob_names} and branch {branch}. "
+                f"Will delete heartbeats, cronjobs {cronjob_names} and branch {branch}. "
                 "Run with --force to execute."
             )
             return
+        # Clean up cronjobs + branch first; delete heartbeats last so a missing
+        # BETTERSTACK_API_KEY_RW can't leave the more important resources orphaned.
         staging.cleanup_staging_resources(dataset_id, version)
+        betterstack.delete_staging_heartbeats(dataset_id, version)
--- a/src/reformatters/common/dynamical_dataset.py
+++ b/src/reformatters/common/dynamical_dataset.py
@@ -1,11 +1,11 @@
 import json
 import subprocess
 from collections.abc import Iterator, Sequence
-from contextlib import contextmanager
+from contextlib import AbstractContextManager, ExitStack, contextmanager
 from datetime import datetime
 from functools import partial
 from pathlib import Path
-from typing import Annotated, Any, Generic, Literal, Self, TypeVar
+from typing import Annotated, Any, Generic, Literal, Protocol, Self, TypeVar
 
 import icechunk
 import numpy as np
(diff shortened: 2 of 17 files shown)


src/reformatters/common/deploy.py
```diff
@@ -95,6 +301,9 @@ def transform(cronjob: kubernetes.CronJob) -> kubernetes.CronJob:
             docker_image=docker_image,
             dataset_id_filter=dataset_id,
             cronjob_transform=transform,
+            # Staging crons are not heartbeat-provisioned, so a staging deploy needs
+            # no BETTERSTACK_API_KEY_RW.
+            provision_heartbeats=False,
```
aldenks: I would like to see heartbeats from staged datasets if that's not a pain

src/reformatters/common/dynamical_dataset.py
```diff
@@ -56,6 +55,34 @@
...
+    ) -> AbstractContextManager[None]: ...
+
+
+_RUN_MONITORS: list[RunMonitor] = []
+
+
+def register_run_monitor(monitor: RunMonitor) -> None:
+    """Register a monitor to wrap every operational cron run. With none registered,
+    monitoring is a no-op."""
+    _RUN_MONITORS.append(monitor)
```
aldenks: move run monitoring helpers to bottom of file so readers open the file and see `class DynamicalDataset`

src/reformatters/common/deploy.py
```diff
@@ -1,22 +1,34 @@
+import base64
```
aldenks: can we move all the betterstack provisioning in this file into betterstack.py and have this call it so betterstack stuff is collected in one place?

src/reformatters/common/kubernetes.py
```diff
@@ -274,7 +282,15 @@ def as_kubernetes_object(self) -> dict[str, Any]:
         }
 
 
-class ReformatCronJob(CronJob):
+class HeartbeatCronJob(CronJob):
+    """A cron whose runs report to Better Stack heartbeats; mounts the heartbeat
+    url-map secret alongside any dataset-specific secrets."""
+
+    def mounted_secret_names(self) -> Sequence[str]:
+        return [*self.secret_names, BETTERSTACK_HEARTBEATS_SECRET_NAME]
```
aldenks: do we need this class, is there any instance of a CronJob that doesn't send heartbeats? even if there is im ok including it on all CronJobs

src/reformatters/common/deploy.py
mrshll: Done — staging now provisions its own heartbeats (removed the `provision_heartbeats=False`), and `cleanup-staging` deletes them via `betterstack.delete_staging_heartbeats`. The staging deploy workflow now passes `BETTERSTACK_API_KEY_RW`.

src/reformatters/common/dynamical_dataset.py
mrshll: Done — `RunMonitor` / `register_run_monitor` / `_RUN_MONITORS` now live at the bottom of the file, so `class DynamicalDataset` is the first thing you see.

src/reformatters/common/deploy.py
mrshll: Done — all provisioning (`provision_heartbeats`, `reconcile_heartbeats`, `schedule_period`, secret read/write, `delete_staging_heartbeats`) moved into `betterstack.py`; `deploy.py` just calls `betterstack.provision_heartbeats(...)`. Provisioning tests moved to `betterstack_test.py` too.

src/reformatters/common/kubernetes.py
mrshll: Done — dropped `HeartbeatCronJob`; `CronJob.mounted_secret_names()` now mounts the heartbeat secret on every cron (archive included — its runtime monitor just finds no matching entry and skips).

PR #798, opened 2026-07-24
Fix GEFS NOMADS redirect and IMERG publish-latency error noise

mrshll:
## What

Fixes the two highest-volume error patterns in Better Stack for the `reformatters` app.

**1. GEFS 35-day: NOMADS 302 redirects logged as exceptions (~1.1K occurrences)**

`noaa-gefs-forecast-35-day-update` requests f840 (35-day) output for all 31 ensemble members at the 00Z cycle, per the dataset's own template. NOAA doesn't always produce that full extended-range tail for every perturbed member on every cycle, and NOMADS signals "not found" for these files via a 302 redirect rather than a 404. `materialized_region_job.py`'s not-found classifier only recognized `FileNotFoundError` and HTTP 403/404, so every one of these expected gaps was logged via `log.exception` (shipped to error tracking) instead of `log.info`.

Fix: treat an `httpx.HTTPStatusError` whose response `is_redirect` the same as 403/404 in the classifier — a redirect surviving `raise_for_status()` despite `follow_redirects=True` already means httpx couldn't resolve it to a real resource, so on NOMADS it's functionally "not found."

**2. IMERG Late: precipitationQualityIndex RasterioIOError (584 occurrences, top volume)**

`nasa-imerg-analysis-late-update` requests granules up to `pd.Timestamp.now()` with no publish-latency buffer. The Late run's `precipitationQualityIndex` grid isn't populated in the trailing granule until ~14.7h (p95) after nominal time (`analysis_late/dynamical_dataset.py`'s own comment documents this), so the update job kept trying to read a field that wasn't written yet, every run, on the same trailing granule — which also cascaded into `nasa-imerg-analysis-late-validate` failing with "No data found within 18:00:00 of now" since the store never advanced.

Fix: add a `publish_latency: ClassVar[pd.Timedelta]` to `NasaImergAnalysisMaterializedRegionJob` (mirroring `VirtualRegionJob.operational_update_window`), set per run (Early: 4.6h, Late: 14.7h — the p95 figures already measured by `src/scripts/imerg_latency_probe.py`), and subtract it from `append_dim_end` in `operational_update_jobs`.

## Other errors triaged (no code change)

- SMAP 404 for very recent dates and IMERG jsimpson `ConnectionResetError` — both transient/self-healing (source hasn't published yet / server-side network flakiness), already retried, and within each dataset's tolerated staleness window. Resolved as noise in Better Stack rather than code changes.

## Test plan

- [x] `uv run ruff format && uv run ruff check --fix && uv run ty check` clean
- [x] Added `test_httpx_302_recent_logs_info` to `TestDownloadErrorLogging` (mirrors existing 403/404 cases)
- [x] Added `test_variant_region_jobs_carry_publish_latency`; updated the existing `test_backfill_local_and_operational_update` integration test to account for the new latency buffer
- [x] Full test suite: `uv run pytest` — 1893 passed, 10 skipped

https://claude.ai/code/session_013g8BeQ81opdTTmS5QYHpjF

---
_Generated by [Claude Code](https://claude.ai/code/session_013g8BeQ81opdTTmS5QYHpjF)_

### DIFF
--- a/src/reformatters/common/materialized_region_job.py
+++ b/src/reformatters/common/materialized_region_job.py
@@ -312,8 +312,13 @@ class MaterializedRegionJob(
                 two_days_ago = pd.Timestamp.now() - pd.Timedelta(hours=48)
                 is_not_found = isinstance(e, FileNotFoundError) or (
                     isinstance(e, httpx.HTTPStatusError)
-                    and e.response.status_code
-                    in (HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND)
+                    and (
+                        e.response.status_code
+                        in (HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND)
+                        # Some sources (e.g. NOMADS) redirect instead of 404ing when
+                        # a file isn't published yet.
+                        or e.response.is_redirect
+                    )
                 )
                 if (
                     is_not_found
--- a/src/reformatters/nasa/imerg/analysis_early/region_job.py
+++ b/src/reformatters/nasa/imerg/analysis_early/region_job.py
@@ -1,6 +1,9 @@
+import pandas as pd
+
 from reformatters.nasa.imerg.imerg_config_models import ImergRun
 from reformatters.nasa.imerg.region_job import NasaImergAnalysisMaterializedRegionJob
 
 
 class NasaImergAnalysisEarlyRegionJob(NasaImergAnalysisMaterializedRegionJob):
     run: ImergRun = "early"
+    publish_latency = pd.Timedelta(hours=4.6)
--- a/src/reformatters/nasa/imerg/analysis_late/region_job.py
+++ b/src/reformatters/nasa/imerg/analysis_late/region_job.py
@@ -1,6 +1,9 @@
+import pandas as pd
+
 from reformatters.nasa.imerg.imerg_config_models import ImergRun
 from reformatters.nasa.imerg.region_job import NasaImergAnalysisMaterializedRegionJob
 
 
 class NasaImergAnalysisLateRegionJob(NasaImergAnalysisMaterializedRegionJob):
     run: ImergRun = "late"
+    publish_latency = pd.Timedelta(hours=14.7)
--- a/src/reformatters/nasa/imerg/region_job.py
+++ b/src/reformatters/nasa/imerg/region_job.py
@@ -1,6 +1,6 @@
 from collections.abc import Callable, Mapping, Sequence
 from pathlib import Path
-from typing import Literal, assert_never
+from typing import ClassVar, Literal, assert_never
 from urllib.parse import urlparse
 
 import numpy as np
@@ -121,6 +121,12 @@ class NasaImergAnalysisMaterializedRegionJob(
 ):
     run: ImergRun
 
+    # How long after its nominal time a granule typically takes to fully publish
+    # (all grids populated, not just the file existing). Operational updates stay
+    # behind this so they don't repeatedly request a granule whose fields (e.g.
+    # precipitationQualityIndex) haven't landed yet. See src/scripts/imerg_latency_probe.py.
+    publish_latency: ClassVar[pd.Timedelta]
+
     def generate_source_file_coords(
         self,
         processing_region_ds: xr.Dataset,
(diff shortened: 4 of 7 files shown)


mrshll:
@aldenks this was from the morning exception review. close if not useful

mrshll:
Surfacing this during a routine Sentry triage pass: the GEFS NOMADS-302 fix here (part 1) is the exact root cause of `REFORMATTERS-BJ`, which is by far our highest-volume open error — 44,370+ occurrences since 2026-02-25, still firing today, still ongoing on `noaa-gefs-forecast-35-day-update`. I've marked that Sentry issue resolved pointing at this PR rather than duplicating the fix.

One heads up: this PR's `mergeable_state` is currently `dirty` (merge conflict with `main`), presumably from the ~5 days it's sat as a draft. Flagging so whoever picks it up knows a rebase/merge is needed before it can land.

---

aldenks:
Closing — one half of this is obsolete and the other has a replacement.

**IMERG `publish_latency` — superseded by #811.** That PR (merged 2026-07-25, a day after this one opened) landed the same diagnosis with a better fix: GES DISC serves an HTML/JSON body with HTTP 200 for a not-yet-published granule, which is what failed at read time as `RasterioIOError ... //Grid/precipitationQualityIndex: No such file or directory`. #811 validates the HDF5 magic bytes and falls through to the next candidate, and #804 did the same for jsimpson connection resets. Those errors stop in error tracking at 2026-07-25 12:49 and there are zero in the last 7 days; the Late update is advancing and its validation cron is passing. The fall-through is also strictly better than a latency buffer here — it ingests whatever is ready instead of putting a permanent 14.7h floor under the dataset's freshness.

**GEFS NOMADS redirects — still live, replaced by #862.** ~2800 events/week as of today, exactly as described here. But the fix belongs in `gefs/utils.py`, not in the shared classifier: `noaa_utils.NOMADS_RETRY_STATUS_CODES` already treats a NOMADS 302 as transient Akamai bot mitigation, and GFS and HRRR both pass it (plus `nomads_rate_limiter`) to `httpx_download_to_disk`. The GEFS fallback passed neither, which is why it is the only NOMADS consumer generating this noise. #862 makes it consistent rather than teaching `is_not_found()` that every redirect everywhere means "file absent".

aldenks:
Superseded by #811 (IMERG) and #862 (GEFS NOMADS redirects).

aldenks:
Follow-up on the GEFS half, since #798 and #862 carry contradictory diagnoses of the same 302 and it's worth pinning down which is right for the record.

**#798's premise here was incorrect.** This PR read the NOMADS 302 as "not found" — that NOAA doesn't always produce the extended-range tail for every perturbed member, and NOMADS signals the gap with a redirect instead of a 404. #862 reads the same 302 as Akamai bot mitigation (transient, retryable).

Probing NOMADS directly settles it:

| status | file |
|---|---|
| **200** | `gefs.20260805/00/atmos/pgrb2bp5/gep25.t00z.pgrb2b.0p50.f450.idx` — the exact file from the error trace |
| 200 | same path at `f006` (control) |
| **404** | a deliberately non-existent filename in that directory |
| 403 | same file under `gefs.19990101/` (aged out of retention) |

Two conclusions:

1. NOMADS signals a genuinely absent file with **404**, not 302. A 302 is never "file is missing."
2. The specific file this error fires on **exists and serves 200**. The download was failing because we were being throttled, not because there was nothing to fetch — GEFS was the only NOMADS consumer passing neither `nomads_rate_limiter` nor `NOMADS_RETRY_STATUS_CODES`, exactly as #862 describes.

So #862's fix addresses the real cause, and the "Alternative considered" rejection in that PR was right on the facts as well as on principle.

Worth recording why this matters beyond bookkeeping: had #798 merged, a bot-mitigation 302 would have been classified by `download.is_not_found()` into the quiet recent-gap path, writing NaN for data that exists on NOMADS. That is silent data loss, and because the change was to the shared classifier rather than the GEFS path, it would have applied to every dataset that falls back to NOMADS. It would also have presented as success: error volume drops to zero while gaps appear in the archive.

Closing this out as superseded is the right call. `REFORMATTERS-BJ` should clear when #862 lands.

PR #810, opened 2026-07-25
DWD ICON-EU: tolerate rare single-cell deaccumulation noise for aswdir_s

mrshll:
## What

Fixes [REFORMATTERS-D2](https://dynamical.sentry.io/issues/REFORMATTERS-D2) (513 occurrences since March, still ongoing).

`dwd-icon-eu-forecast-5-day`'s `downward_direct_short_wave_radiation_flux_surface` (`aswdir_s`) deaccumulation intermittently fails with `ValueError: Found 1 values (0.0%) below threshold, expected at most 0.0%`. GRIB precision jitter occasionally pushes a single grid cell just under the deaccumulation threshold — 1 cell out of the ~84M in a region (93 lead times × 657 latitudes × 1377 longitudes) is `1.2e-8` of the array, but `expected_invalid_fraction` defaults to `0.0`, so any nonzero count trips it.

This variable already has an analogous override for a different kind of GRIB noise (`deaccumulation_expected_clamp_fraction=0.25`, "GRIB precision jitter in the running mean drives nighttime clamping to ~20%"). This PR adds the missing `deaccumulation_expected_invalid_fraction` counterpart (mirroring the existing `expected_clamp_fraction` plumbing end to end) and sets it to `1e-6` for `aswdir_s` — generous enough to absorb this single-cell noise while still catching a real systemic problem.

Note: while investigating this issue, I found that Sentry's default exception fingerprinting groups **every** dataset's validation failure into one bucket (`REFORMATTERS-D2`'s "latest event" was this radiation noise, but the same issue also aggregated an unrelated one-time MRMS backfill event with 35-94% invalid precipitation values from 2026-07-21, which appears to already be resolved/complete — no recurrence since). I've resolved the umbrella Sentry issue and left a comment; worth considering a fingerprint rule so future unrelated validation failures don't keep landing in the same bucket (same problem exists for `REFORMATTERS-AY`, also resolved with a comment).

## Test plan

- [x] `uv run ruff format && uv run ruff check --fix && uv run ty check` clean on the touched files
- [x] Extended the existing `test_region_job_apply_data_transformations_deaccumulation_optional_kwargs` test to cover the new `expected_invalid_fraction` kwarg
- [x] `uv run pytest tests/dwd/icon_eu/` — 33 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01M1FfmFACxhdckvWdYTN8y7)_

### DIFF
--- a/src/reformatters/dwd/icon_eu/forecast_5_day/region_job.py
+++ b/src/reformatters/dwd/icon_eu/forecast_5_day/region_job.py
@@ -182,6 +182,7 @@ class DwdIconEuForecast5DayRegionJob(
             optional_kwargs: dict[str, Any] = {
                 "invalid_below_threshold_rate": attrs.deaccumulation_invalid_below_threshold_rate,
                 "expected_clamp_fraction": attrs.deaccumulation_expected_clamp_fraction,
+                "expected_invalid_fraction": attrs.deaccumulation_expected_invalid_fraction,
             }
             optional_kwargs = {
                 k: v for k, v in optional_kwargs.items() if v is not None
--- a/src/reformatters/dwd/icon_eu/forecast_5_day/template_config.py
+++ b/src/reformatters/dwd/icon_eu/forecast_5_day/template_config.py
@@ -46,6 +46,8 @@ class DwdIconEuInternalAttrs(BaseInternalAttrs):
             to tolerate the larger negative noise produced by lossy-compressed radiation fields.
         deaccumulation_expected_clamp_fraction (float | None): Override for the fraction of values
             expected to be clamped to 0.
+        deaccumulation_expected_invalid_fraction (float | None): Override for the fraction of
+            values expected to fall below `deaccumulation_invalid_below_threshold_rate`.
         deaccumulation_type (AccumulationType): Whether the source values are cumulative totals
             ("accumulated", default) or running-mean rates whose averaging window grows from
             forecast start ("running_mean", used for ICON-EU averaged radiation fields).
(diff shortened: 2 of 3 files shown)


aldenks:
Closing in favour of #865, which fixes the same Sentry issue by widening the clamp threshold instead of allowing a nonzero `expected_invalid_fraction` — so consumers get `0` rather than occasional NaNs.

The deciding evidence: running the production deaccumulation over the real GRIBs for 18 init times spanning the whole archive (~1.5e9 values), the most negative rate anywhere is **−77.2 W/m²**, and nothing reaches −100. Meanwhile a corrupt or mismatched source step deaccumulates to thousands of W/m². That leaves a clean gap, so `aswdir_s` can take a −1000 W/m² invalid threshold and keep `expected_invalid_fraction` at 0: the artifact clamps to 0, and only genuinely broken data becomes NaN and pages us.

Two details worth carrying over from that investigation:

- The values tripping the threshold aren't GRIB precision jitter. The packing quantum is 1/64 W/m², which with the running-mean amplification factor can only produce negatives down to about −1.2 W/m². Everything past ~−2 is real non-monotonicity in DWD's running mean, concentrated at the domain edges, high latitudes and the Caucasus — and the current −50 threshold has been silently clamping hundreds of those per init all along.
- The `expected_clamp_fraction` of 0.25 is also mis-tuned and will start firing this winter: negatives reach 27.7% of an init's values in February, and February is the darkest data the archive holds. #865 raises it to 0.4 for both radiation variables.

---

Issue #761, opened 2026-07-20
NOAA HRRR 18-hour forecast dataset


mrshll:
Add a NOAA HRRR 18-hour forecast dataset to the catalog.

Part of catalog fill-out (2026-07-20 planning).

## Dataset development stages

- [x] Explore real NOAA GRIB2 and `.idx` source files.
- [x] Checkpoint A: confirm dataset scope.
- [x] Implement the dataset, template, registry entry, and tests.
- [x] Review the implementation independently and resolve findings.
- [x] Merge PR #814 to `main`.
- [x] Backfill the production Icechunk store.
- [x] Validate all variables and upload a draft report.
- [x] Checkpoint B: obtain report approval and publish it.
- [x] Publish the dataset through `dynamical-stac`.

mrshll:
🔗 Linked via toad — https://github.com/dynamical-org/reformatters/pull/740

toad-link:github:dynamical-org/reformatters#740

mrshll:
🔗 Linked via toad — https://github.com/dynamical-org/reformatters/pull/743

toad-link:github:dynamical-org/reformatters#743

mrshll:
validation running

mrshll:
## Validation results — draft report uploaded

Draft: https://dataset-validation-reports.dynamical.org/noaa-hrrr-forecast-18-hour-virtual/drafts/v0.1.0_2026-07-29T15-50/validation_report.html

Full `run-all` over the production store (`v0.1.0`, 70,492 hourly inits 2018-07-13T12:00 → 2026-07-28T15:00, 176 variables), reviewed per `docs/validation.md` §3f: 16 theme-coherent batches covering every variable's `value_timeseries_`/`spatial_`/`temporal_` plots, then 13 verification passes for each SUSPECT finding. Spatial snapshot pinned to init 2026-07-24T04:00 lead 14h (valid 18:00 UTC, midday CONUS) and points pinned to the 48-hour virtual report's points so the two are directly comparable.

### The store itself is sound

[...]

PR #820, opened 2026-07-30
Fix HRRR virtual validation findings

mrshll:
Relates to #761.

## Plan

- [x] Reproduce the two availability-reporting defects.
- [x] Match carrier variables by full Zarr path instead of leaf name.
- [x] Count unprobed init times as incomplete rather than dropping them from the denominator.
- [x] Correct the HRRR shear units, ground-heat-flux standard name, and time-maximum wind metadata.
- [x] Regenerate the 18-hour and 48-hour virtual templates.
- [x] Add regression coverage and update validation documentation.
- [x] Launch and verify the targeted production backfill for the seven recoverable init times.

## Operational work

The `overwrite-chunks` backfill for the seven init times completed successfully (`1/1`, 87s): https://github.com/dynamical-org/reformatters/actions/runs/30566193787

The existing 18-hour draft report was repaired locally without repeating the manifest scan: all 176 variables now use a 70,492-init denominator, 173 are listed as incomplete, and the intrinsic-variable-facts section was removed. Uploading the replacement report files remains separate from this code change.

## Validation

- `uv run ruff format src tests`
- `uv run ruff check --fix src tests`
- `uv run ty check --output-format concise --exclude tests/common/operational_run_guard_integration_test.py src tests`
- Focused HRRR/availability regression set: 100 passed
- Full tracked suite: 1,918 passed, 10 skipped, 5 unrelated failures (three platform-specific HRRR GeoTransform last-digit comparisons; two pre-existing SMAP template round-trip mismatches)

## Retro

The availability findings were real report-tooling bugs, not archive corruption. Correcting the completed scan output directly avoided another whole-archive manifest pass; regression tests now preserve both denominator and group-path behavior.


aldenks (approved):
Good catches

PR #821, opened 2026-07-31
HRRR virtual-fast: NOMADS-fed cache + reduced variable set

mrshll:
## Summary

Adds `noaa-hrrr-forecast-48-hour-virtual-fast` and the machinery to feed it from **NOMADS** rather than NOAA's AWS archive (NODD). Two independent latency wins that compose: fewer files to wait for (sfc only), and a fresher source.

**Not yet shippable as a dataset** — the private cache bucket and secret are provisioned and the mirror is running its observation trial, but ingest remains suspended until NOMADS → bucket → expiry is verified over scheduled cycles.

## Why NOMADS

Measured from wxopticon's 90-day/360-init arrival stats (`external-noaa-hrrr-ftp` vs `external-noaa-hrrr-aws`), seconds after init:

| lead group | NOMADS p50 | AWS p50 | Δp50 | NOMADS p99 | AWS p99 | Δp99 |
|---|---|---|---|---|---|---|
| f000 | 50m00s | 51m52s | +112s | 54m00s | 74m01s | **+1201s** |
| f006 | 58m00s | 62m22s | +262s | 61m00s | 84m55s | **+1435s** |
| f018 | 73m00s | 77m15s | +255s | 76m00s | 89m58s | +838s |
| f048 | 106m00s | 109m41s | +221s | 109m00s | 123m16s | +856s |

NOMADS is also far more predictable (f000 p50→p99 spread 240s vs 1329s), and the gain is largest at the earliest leads. A per-file check over one day agreed: NODD lagged NOMADS by a median 97s and was never earlier (0/803 files).

Measured in-cluster throughput is 18.4 MB/s, so a ~152 MiB wrfsfc file copies in ~8.3s and end-to-end overhead lands around 19s — comfortably inside the smallest margin.

## The commits

- **Extension points** on the shared HRRR virtual base (`source_file_coord_class`, `_download_index`); v3/v4 forecast-length constants hoisted to `noaa/hrrr/template_config.py`; icechunk's deprecated `None` credential replaced with `LocalFileSystemAccess`.
- **The dataset** — 28 root variables, all wrfsfc, *derived* from the materialized catalog rather than hand-listed, substituting only the variables the materialized product deaccumulates (deaccumulation spans lead times, so no read-time codec can produce those rates):
  - `precipitation_surface` → `total_precipitation_surface` + `precipitation_rate_surface`
  - `snowfall_surface` → `total_snowfall_run_total_surface`

  Asserts pin the substitute map to exactly the deaccumulated set *and* fail if any target name lacks a virtual-catalog equivalent, so drift in either catalog breaks the build.
- **The mirror** — a `-mirror` CronJob running `mirror-nomads-gribs`, shaped like the DWD/ECCC archivers (same `common/rclone.py`, same CronJob+CLI+`_monitor` pattern) with differences forced by sitting on the latency path: it polls through the publication window instead of sweeping once; it passes an explicit `--files-from` list and never re-passes a copied file; it serializes NOMADS source transfers/checks and avoids range-stream fan-out; and it omits `--min-age`, which would eat most of the median gain.
- **Suspended ingest** so the mirror can be deployed and observed alone.
- **Cache-first discovery** with archive fallback, so a broken or expired cache degrades to today's latency rather than leaving holes, and backfills run unchanged over inits older than the cache's retention.

## Trial results

Provisioned `dynamical-noaa-hrrr-nomads-cache` as a private, ownership-enforced, SSE-S3 bucket in `us-west-2`, with 1-day object expiry and 1-day incomplete multipart cleanup. A least-privilege IAM principal backs `noaa-hrrr-nomads-cache-storage-options-key`; an in-cluster write/read check passed.

The first manual 00Z catch-up exposed the request-budget failure: rclone multiplied four transfers by four range streams and Akamai returned 302s. Commit `42a9134e` removes those unsafe knobs and forces one transfer/checker with no multi-thread range requests.

The Sentry-disabled retry mirrored all 49 GRIBs and indexes without a 302 at 18.4–21.7 MiB/s. Karpenter evicted the first pod for underutilization; the disruption retry resumed from already cached objects and completed successfully. The fixed scheduled mirror is active, update/validate remain suspended, and Sentry is temporarily disabled only for mirror trial cycles.

## Verification

1830 fast tests, ruff and ty clean before deployment. After `42a9134e`, the focused mirror/CronJob suite passes 15/15 and the remote Code Quality workflow passes. The registry-driven template-subclass and CF-compliance suites pick up the new dataset with no allowlist edits.

Load-bearing assumptions were spiked before building on them: repointing a chunk ref to a different container works and detaches cleanly (the cache object can be deleted afterward with the read intact); `all_virtual_chunk_locations()` returns container-qualified URLs; NOMADS→us-west-2 throughput measured in-cluster.

## Notes for review

- The fast coord is **mutable**, unlike every other `SourceFileCoord`. The write loop drops coords by object identity, so discovery must record the source on the coord it was handed. Note `Field(frozen=False)` does *not* achieve this — `FrozenBaseModel` sets `frozen` at the model level and that wins — so the subclass overrides `model_config`. The existing `frozen=False` markers on `status`/`downloaded_path` (`common/region_job.py:72-73`) have no effect; those are replaced via `replace()`, not mutated. Worth deleting separately.
- The private cache needed credential plumbing in two shared utilities: `skip_signature` on `s3_download_to_disk` (`s3_store` already had it) and `container_credentials` on `IcechunkVirtualConfig` so readers can resolve leading-edge refs while the archive stays anonymous.
- `dynamical-noaa-hrrr-nomads-cache` is assumed as the bucket name in `nomads_cache_rclone_root` and `CACHE_LOCATION_PREFIX`.

## Remaining

Observe scheduled 0/6/12/18Z cycles, compare cache/NODD `LastModified`, and verify 1-day expiry; then raise retention to 3 days. Add docs (`virtual_datasets.md` two-source section, `ops_card.md` runbook) and a slow end-to-end test against a `file://` cache; backfill the store; restore mirror Sentry check-ins; then unsuspend ingest. The rebase-to-archive step is deliberately deferred — spiked and working, but rolling-window phase 1 doesn't need it.

https://claude.ai/code/session_01JQVTXKVrRpzZP7HtQEHpXJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01JQVTXKVrRpzZP7HtQEHpXJ)_


aldenks:
@mrshll gonna close this for now to help clean up but reopen whenever

PR #843, opened 2026-08-04
Reduce hrrr-18 alert noise: bound update polling with a fire-anchored deadline

### LINKED ISSUE #839: review hrrr-18 update/validate schedule to reduce alert noise

mrshll:
Closes #839.

`noaa-hrrr-forecast-18-hour-virtual` produced 7 of the ~14 alerts in `#ops-reformatters` in its first ~3 days operational (~2.3/day), all of them false: the data was complete and correct every time. Two changes remove the causes.

## What the logs showed

The update job's runtime *is* the source's last-file arrival time (each tick logs `Ingesting N files, M still pending`, and the run ends within ~2 min of the final file). It fires at `:50` with a 55 min `pod_active_deadline`, so it is killed at **init+105m**. Measured last-file arrival on `noaa-hrrr-bdp-pds` (288 inits / 12 days): p50 **+87m**, p95 **+99m**.

Three things exhausted that headroom:

1. **The 23Z cycle, nightly and drifting.** The NODD mirror publishes 42–48 of the 57 files as one late bulk drop. NCEP is on time — 08-01 23Z `wrfprsf18` hit NOMADS at **+88m**, S3 at **+113m** (25 min mirror lag; every other init's is ~1 min). The lag creeps **+1.8 min/day** with periodic resets (+93m on 07-21 → +116m on 08-02), crossing +105m on 07-28, before this dataset went operational. 3 alerts.
2. **Single-file mirror hiccups.** 08-01 03Z: `wrfsfcf18`/`wrfnatf18` at +89m, `wrfprsf18` alone at +106m. 1 alert.
3. **Pod evictions that can never finish.** 5 of 48 fires had their pod replaced mid-run (spot capacity). The replacement recomputes `append_dim_end = now`, so it adopts the *next* init (pending jumped 27 → 83 at 08-02 22:12) and cannot finish before `concurrencyPolicy: Replace` kills it at the next fire — even though 21Z was fully published at +88m. 1 alert, plus the nonsense Sentry durations (276s "ok", 894s/1326s "timeout") that measure killed processes.

Plus one unrelated alert, left unaddressed here: the 08-02 06:48 **validate** run passed in 36 s (`Zarr validation passed all checks`) and was still reported as a timeout — Sentry got `in_progress` and never got the `ok`. If that recurs, an explicit check-in flush is the fix.

**Latency is already optimal**, which is why this is purely an alerting fix: the 23Z files hit S3 at 00:53–00:54 and the 00:50 fire ingested them at 00:52–00:55, ~2 min after publication. Waiting in-run and handing off to the next fire cost the same.

## Changes

**Bound virtual update polling with a fire-anchored deadline.**

- `CronJob.previous_fire_time(now)` (`common/kubernetes.py`) — hand-rolled parse of the `<minute> <hours> * * *` shape every operational schedule uses (`*`, `*/N`, `A-B/N`, comma list); asserts on anything else, with `tests/deploy_test.py` checking every registered dataset's schedules parse so an unsupported one fails in CI, not at 3am.
- `VirtualRegionJob.poll_deadline` — the poll loop returns when it passes, leaving unpublished files pending. This is the shape the backfill branch already had; nothing downstream requires a fully-ingested window.
- `DynamicalDataset._virtual_poll_deadline()` — `previous_fire_time(now) + pod_active_deadline − 30 s`, pulled off the running `ReformatCronJob` so the schedule and deadline have one source of truth. The grace only has to cover the exit and final check-in (the last tick's commit has already landed), and it is also the window in which a file that would have been ingested in-run is deferred to the next fire, so it is kept short.

Anchoring on the **scheduled fire** rather than process start is what fixes failure mode 3: every pod of a fire, including a replacement, stops at the same wall-clock time, so an evicted run hands off instead of polling into the next fire and being killed. A positive grace always lands the poll deadline before the pod's own deadline, which runs from pod start (never earlier than the fire).

**Give hrrr-18 the full hour between fires.** `pod_active_deadline` 55 → 59 min, so polling stops at :48:30, 90 s before the next fire, instead of :45. No fire-time change: firing later would delay f00–f05 (published at +51m) for every init. The poll deadline is now init+108.5m, past p95 (+99m).

**Tighten hrrr-18's validation.** The validate cron moves :48 → :49 — the update's fire plus its `pod_active_deadline`, so the run being validated has always stopped writing — and `min_present_fraction` goes `(0.0, 0.0, 1.0)` → `(0.0, 0.05, 1.0)`. The leading `0.0` is structural: the validator's window ends at `now`, so its newest position is the init the *next* fire will chase, with nothing ingested for it yet. The second position is the init the just-ended run chased, which may legitimately have deferred late files (the 23Z bulk drop defers 42–48 of 57), so it can't be held to 1.0 — but f00 lands an hour before the poll deadline, so a floor of 5% (3 of 57 files, one lead's worth) still catches a cycle the source never published, an hour earlier than the `1.0` position would.

| | before | after |
|---|---|---|
| 23Z bulk drop (+113m) | killed at 00:45 → alert | exits ok at 00:48:30, 00:50 fire ingests at 00:52–00:55 |
| `wrfprsf18` straggler (+106m) | killed at 04:45 → alert | ingested in-run at 04:46 |
| eviction at 22:12 | killed by Replace at 22:50 → alert | same 22:48:30 deadline → exits ok |

## Deliberately not changing

- **No fire-time changes anywhere, and no new margin for the 23Z cycle.** Its mirror lag drifts +1.8 min/day, so any deadline chosen to cover it is eaten within days; those files roll to the next fire and land ~2 min after publication either way.
- **`failure_issue_threshold` stays at 1.** With handoff being a success, a timeout again means a genuinely stuck run.
- **Real outages still alert.** A source that stops publishing is caught by the validate cron (`check_forecast_current_data(2h)`, `CheckVirtualManifestCompleteness`), not by the update job — this change only stops treating "not published yet" as a failure.
- **No schedule changes for `noaa-hrrr-forecast-48-hour-virtual` or `noaa-gefs-forecast-10-day-spatial-dev`.** They inherit the same failure semantics — a late source ends their run `ok` with the stragglers deferred to the next fire — using their existing `pod_active_deadline`. Both keep generous headroom (poll stops ~35 min and ~11–20 min past observed completion), and a deferral there is not silent: both validate with the default `min_present_fraction=(1.0,)`, so an incomplete init surfaces at the next validation.
- `pod_active_deadline` remains as the backstop it should be, rather than the normal exit path.

## Follow-up worth considering

Mode 3 is spot-eviction-driven; moving this cronjob to on-demand capacity would cut its frequency independently. Not needed for correctness — the anchored deadline makes evictions harmless — so it's a cost/noise tradeoff, not a fix.

## Testing

`uv run pytest`: 1941 passed, 10 skipped. New coverage: `previous_fire_time` across all four hour forms and boundary cases (including walking back into the previous day) plus rejection of unsupported schedules; the poll loop returning at a past deadline with files still pending; the deadline reaching the write loop; every registered dataset's schedules parsing. `ruff format`, `ruff check`, `ty check` clean.

### 2026-08-03 — investigation and implementation

- Root-caused the 5 update timeouts and 2 validate alerts from Sentry check-ins, `#ops-reformatters`, reformatter logs, and direct `Last-Modified` measurements against `noaa-hrrr-bdp-pds` and NOMADS.
- wxopticon could not have surfaced this: `external-noaa-hrrr-aws` is `cadence_hours=6`/49 leads, i.e. only the 00/06/12/18Z cycles, which are the *fastest* ones (+73–77m). The 23Z cycle is unmonitored. Worth adding an off-cycle HRRR product there.


src/reformatters/common/dynamical_dataset.py
```diff
@@ -46,14 +46,19 @@
...
 DATA_VAR = TypeVar("DATA_VAR", bound=DataVar[Any])
 SOURCE_FILE_COORD = TypeVar("SOURCE_FILE_COORD", bound=SourceFileCoord)
 
 log = get_logger(__name__)
 
+# Room for a virtual update's final commit and check-in after it stops polling. Any
+# positive value lands the poll deadline before the pod's own deadline, which runs from
+# pod start (never earlier than the scheduled fire).
+_POLL_DEADLINE_GRACE = pd.Timedelta("3min")
+
```
```suggestion
_POLL_DEADLINE_GRACE = pd.Timedelta(seconds=30)  # stop polling before pod active deadline
assert _POLL_DEADLINE_GRACE.total_seconds() > 0

```

src/reformatters/common/kubernetes.py
```diff
@@ -13,10 +14,44 @@
...
+    if "," in spec:
+        return frozenset(int(hour) for hour in spec.split(","))
+    if spec == "*":
+        start, end = 0, 23
+    elif "-" in spec:
+        start, end = (int(bound) for bound in spec.split("-"))
+    else:
+        return frozenset({int(spec)})
+    return frozenset(range(start, end + 1, int(step) if step else 1))
+
```
aldenks: Move _SCHEDULE_PATTERN _parse_schedule and _parse_hours to the bottom of the file

src/reformatters/common/monitoring.py
```diff
@@ -10,6 +10,8 @@
 from reformatters.common.iterating import digest
 from reformatters.common.kubernetes import CronJob
 
+_CHECKIN_FLUSH_TIMEOUT_SECONDS = 10.0
+
```
aldenks: remove

src/reformatters/common/monitoring.py
```diff
@@ -43,6 +45,9 @@ def capture_checkin(status: Literal["ok", "in_progress", "error"]) -> None:
                 "recovery_threshold": 1,
             },
         )
+        # Deliver now: relying on the at-exit flush drops a check-in when the process
+        # is killed or exits right after, which Sentry reports as a failed run.
+        sentry_sdk.flush(timeout=_CHECKIN_FLUSH_TIMEOUT_SECONDS)
```
aldenks: remove this. now we are explicitly exiting with a 30s grace period we should be fine without

src/reformatters/common/virtual_region_job.py
```diff
@@ -70,6 +70,10 @@ class VirtualRegionJob(
     # Updates wait for source files as the provider publishes them, backfills check once
     processing_mode: Literal["backfill", "update"] = "backfill"
 
+    # Wall clock time an update stops waiting on files the source has not published yet,
+    # leaving them to the next fire. None waits until every file is ingested.
+    poll_deadline: Timestamp | None = None
```
aldenks: can we remove having None be an option here? 

also single line comment:   `# stop polling after this time`

src/reformatters/example_virtual/dynamical_dataset.py
```diff
@@ -56,14 +56,16 @@ def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
...
         """
         # suspend = True  # Defaults to False, remove after backfilling to run operational updates and validation
         # operational_update_cron_job = ReformatCronJob(
         #     name=f"{self.dataset_id}-update",
         #     schedule="0 6 * * *",
-        #     # Bound how long the single writer polls for files that never publish;
-        #     # keep it well under the gap between fires so runs never overlap.
+        #     # Sets how long a fire chases its own source files (the poll deadline is
+        #     # the fire plus this, less a small grace); keep it well under the gap
+        #     # between fires so runs never overlap.
```
aldenks:
```suggestion
        #     # the fire plus this, less a small grace); keep it under the gap
        #     # between fires so runs never overlap.
```

src/reformatters/common/dynamical_dataset.py
aldenks: Applied, with the comment on its own line — ruff format wraps the trailing-comment version to three lines at 88 chars.

src/reformatters/common/kubernetes.py
aldenks: Moved to the bottom of the file.

src/reformatters/common/monitoring.py
aldenks: Removed.

src/reformatters/common/monitoring.py
aldenks: Removed (both the flush and the constant), and dropped it from the PR title/description.

One caveat: the 08-02 06:48 **validate** alert was this failure mode, not the poll one — that run finished in 36 s and never delivered its `ok`. Nothing in this PR changes the validate path, so if it recurs the flush is the fix. Noted in the description rather than kept in code.

src/reformatters/common/virtual_region_job.py
aldenks: Yes — `poll_deadline: Timestamp = pd.Timestamp.max`, comment is now `# Stop polling after this time, leaving unpublished files to the next fire.`

`None` only existed because backfill jobs come from `RegionJob.get_jobs()`, which knows nothing about polling; a far-future sentinel covers that just as well and the write loop check drops to a plain `pd.Timestamp.now() >= self.poll_deadline`.

src/reformatters/example_virtual/dynamical_dataset.py
aldenks: Applied.

aldenks (approved):

PR #844, opened 2026-08-04
Add probe-position validation command

mrshll:
Adds a `probe-position` command to the validation CLI, plus the upstream-verification guidance that motivated it.

## Why

`unavailable_timestamps.txt` reports only present/expected counts per position. That is not enough to satisfy the [§3e verification gate](../blob/main/docs/validation.md) — classifying a gap as an upstream absence versus an ingestion miss requires knowing *which* source file is missing, and a count like `110/111 present` doesn't say.

This came up re-validating the HRRR virtual stores: one flagged position had all its expected files present upstream at full size with valid indexes, which reads as an ingestion miss. Probing the store named the single missing file, and that file turned out to be a short object upstream whose `.idx` describes a different run entirely — so the reformatter had correctly declined it.

## Changes

- `probe-position <DATASET_URL> <POSITION>` — logs the count and prints the URL of every expected source file the store has no reference for at that position. Registered alongside `availability` / `decode-scan` in `plots.py`, following the existing module-per-command pattern.
- Promotes the `manifest_scan` helpers it reuses to public: `probe_jobs`, `expected_lead_limits`, `coord_is_expected`, `coord_position`.
- `docs/validation.md`: documents the command, and strengthens the §3d upstream-verification bullet. A key existing is not proof the data does — upstream archives carry zero-byte placeholder objects, sidecars orphaned from a 404 object, and objects holding another run's data, all of which a bare existence check reads as present.

## Testing

`ruff format`, `ruff check`, `ty check` clean; `pytest tests/scripts/validation` 50 passed.

https://claude.ai/code/session_013ZjPuhRk3AyPHwVWt3vqGK

---
_Generated by [Claude Code](https://claude.ai/code/session_013ZjPuhRk3AyPHwVWt3vqGK)_


aldenks:
Good stuff

PR #833, opened 2026-07-31
Add ecmwf-aifs-single-forecast-virtual dataset

mrshll:
Closes #831. Part of #763.

Virtual (map-optimized) variant of `ecmwf-aifs-single-forecast`, following the `noaa-hrrr-forecast-48-hour-virtual` pattern. Full exploration report and Checkpoint A scope decisions are in #831.

## What's in this PR

- New dataset `ecmwf-aifs-single-forecast-virtual` under `src/reformatters/ecmwf/aifs_single/forecast_virtual/`: all source variables/levels — 23 single-level + 4 soil (level-in-name) + lead-0-only statics at root, `pressure_level` group of 6 params × 14 levels.
- Era handling via `date_available`: 2025-02-26 (`aifs/` → `aifs-single/` path change; tp/cp unit change m → kg m-2; vars added) and 2026-05-13 (fscov, 10 hPa level).
- Refs into `s3://ecmwf-forecasts` (eu-central-1, anonymous) via the ECMWF JSON-lines `.index`; `GribberishCodec` decodes the CCSDS/AEC packing (verified against raw source bytes in the integration test's snapshot values).
- Filters: K→C ScaleOffset on 2t/2d/skt/sot and pressure-level t; geopotential → height (÷9.80665). Accumulations served raw with `_run_total_` naming per HRRR-virtual precedent.
- Manifest splits: root 600 init_times, pressure_level 200. Crons ship suspended.
- Operational timing derived from measured S3 publish latency (all 61 leads land in a ~2-min burst at ~init+5h25m–6h03m, p99 ~+6h10m): update cron `20 5,11,17,23` (init+5h20m, 1h deadline), validate `20 0,6,12,18` (update+1h).
- Shared changes: `ecmwf_grib_index._parse_index_file` → public `parse_index_file`; `"sol"` added to the levtype Literal; CF-compliance allowlist entries.

## Plan

- [x] Explore source (#831) + Checkpoint A scope decisions
- [x] Scaffold from virtual teaching template
- [x] TemplateConfig / RegionJob / DynamicalDataset + registration
- [x] update-template + tests green (incl. integration test with snapshot values computed independently from raw source messages)
- [x] Code review pass (1 confirmed finding — publish-latency assumption — fixed in a43128b7)
- [x] ruff / ruff check / ty / pytest green, CI green, undrafted

## After merge (tracked in #831)

[...]

PR #861, opened 2026-08-06
Retry icechunk reads through a throttling source bucket

aldenks:
## Why

CI hit a transient failure on #833 ([run](https://github.com/dynamical-org/reformatters/actions/runs/31045423176/job/92442273545)) — not in our download code, but in icechunk fetching a virtual reference:

```
icechunk.StorageError: error fetching virtual reference: service error:
unhandled error (SlowDown): Please reduce your request rate.
```

Auditing test load against `s3://ecmwf-forecasts` showed the failing suite is the *lightest* of the four ECMWF suites, so volume wasn't the problem — retry coverage was. Measured request counts per suite:

| suite | requests | peak concurrency |
|---|---|---|
| `ifs_ens` | 277 | 16 |
| `aifs_ens` | 98 | 16 |
| `aifs_single/forecast` (materialized) | 72 | 16 |
| `aifs_single/forecast_virtual` | ~37 | 6 |

Two paths with very different hardening:

- **Our downloads** (`common/download.py:s3_store`) — `max_retries=16`, 1s→16s backoff. Absorbs `SlowDown` silently.
- **icechunk reads** — `RepositoryConfig.default()`, never setting `.storage`, so icechunk's default 10 tries applied.

The timing confirms retries ran out early rather than never running: the test failed 13.7s after starting (vs ~11s for a clean run), where a full default ladder would burn ~100s.

## What

**1. Set `RepositoryConfig.storage` retries**, matching `s3_store`'s policy, for every icechunk dataset — virtual and materialized.

icechunk builds `storage_settings` from `config.storage` and hands the same object to `VirtualChunkResolver::new`, so this covers both the repo's own reads and virtual source-file fetches. It **merges** with backend defaults rather than replacing them, so concurrency and timeout defaults are untouched.

**2. Reject duplicate `url_prefix`es in `IcechunkVirtualConfig`.** `set_virtual_chunk_container` is a map insert keyed by prefix, so two containers sharing one prefix silently kept only the last container's store config. Verified against the installed build: three distinct prefixes all register; a repeated prefix overwrites. Every current dataset has exactly one container, so this is a guard, not a fix.

## Notes

Unrelated to this PR but found while measuring, if someone wants to pick it up: `VirtualRegionJob.download_concurrency = 64` is unoverridden, so one backfill worker opens ~61 concurrent index GETs (1 init × 61 leads) and virtual backfills run up to 10 workers — against a bucket `docs/backfill.md` says supports at most 8.

### DIFF
--- a/src/reformatters/common/storage.py
+++ b/src/reformatters/common/storage.py
@@ -4,7 +4,7 @@ from collections.abc import Mapping, Sequence
 from enum import StrEnum
 from functools import cache
 from pathlib import Path
-from typing import Any, Literal, assert_never
+from typing import Any, Literal, Self, assert_never
 from urllib.parse import urlparse
 from uuid import uuid4
 
@@ -15,7 +15,7 @@ import zarr
 import zarr.abc.store
 import zarr.storage
 from icechunk.store import IcechunkStore
-from pydantic import Field, InstanceOf, computed_field
+from pydantic import Field, InstanceOf, computed_field, model_validator
 from zarr.abc.store import Store
 
 from reformatters.common import kubernetes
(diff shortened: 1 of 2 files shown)


aldenks:
Decided we don't want this: the dupe url prefix would be obvious if configured wrong and the default 10 retries is already quite a bit, we'd like to fail rather than stall

PR #870, opened 2026-08-07
Add google-weathernext2-forecast-virtual dataset

mrshll:
# 870 — `google-weathernext2-forecast-virtual`

Virtual Icechunk dataset over Google DeepMind **WeatherNext 2**, ensemble-mean product, in
`gs://weathernext/weathernext_2_0_0_mean/`.

Exploration report: `plans/weathernext2_exploration.md`.

## Scope (settled at Checkpoint A, 2026-08-07)

| Decision | Value |
|---|---|
| Product | ensemble **mean** only (`weathernext_2_0_0_mean`); 64-member ensemble is a possible follow-on (~771M refs, needs manifest engineering) |
| `dataset_id` | `google-weathernext2-forecast-virtual` |
| `name` | `Google WeatherNext 2 forecast, virtual` |
| Path | `src/reformatters/google/weathernext2/forecast_virtual/` |
| Root (surface) vars | full archive, `init_time` 2022-01-01 → present |
| `pressure_level` group | **2025-01-01 onward only** |
| Units | asserted, and now empirically verified (below) |

### Why `pressure_level` starts in 2025

Pre-2025 stores pack all 13 levels into a **single blosc buffer** (`chunks [1,1,13,721,1440]`,
~30 MB/object for the mean). A virtual ref addresses whole objects; one level's bytes inside a
compressed buffer are not addressable. 2025+ uses one chunk per level, so only that era can
back a `pressure_level` group. Surface variables use an **identical** grid in all four eras
(`[1,1,721,1440]`, lz4/clevel5/shuffle1, `<f4` — verified in each era), so they span the whole
archive. Result: one store whose `pressure_level` group is NaN before 2025-01-01.

## Source facts (verified, not inferred)

- **Two store layouts.** `2022_to_2023`/`2023_to_2024`/`2024_to_2025` are **one
  `predictions.zarr` per year** with a leading init dim (1460/1460/1464 inits).
  `2025_to_present` is **one store per init**:
  `2025_to_present/{YYYYMMDD}_{HH}hr_01_preds/predictions.zarr` (~2335 stores), with `init_time`
  scalar. `RegionJob` must handle both.
- **6719 inits**, 00/06/12/18Z, no gaps. Operational: newest init is same-day 12Z, `success`
  marker lands at **init + ~6h20m–6h50m**; a whole store uploads in a 3–4 min burst.
- **Lead 6–360 h, step 6 h, 60 steps. No lead 0.**
- Grid 721×1440 @0.25°, **ascending latitude** −90→+90 (poles included), **longitude 0–360**.
  The source's `stored_direction: decreasing` attr is wrong — ignore it.
- 13 pressure levels, source order **ascending** `[50,100,150,200,250,300,400,500,600,700,850,925,1000]`.
- Zarr v2, consolidated `.zmetadata`, `<f4`, C order, `fill_value` NaN, **no sharding**
  anywhere. One chunk = one whole object → refs are `(object, offset=0, size)` and sizes come
  free from a listing.
- **Dim-naming trap**: in 2025+ stores the *lead* dim is named `time`; `datetime` is valid time.

### Codec: proven, no custom codec needed

zarr v3 `BloscCodec(typesize=4, cname="lz4", clevel=5, shuffle="shuffle", blocksize=0)` decodes
these zarr-v2 blosc chunks **byte-for-byte identically** to `numcodecs.Blosc` — verified
end-to-end by swapping a real source chunk into a zarr v3 array and reading it back. This is
our first virtual dataset over a **zarr** source rather than GRIB, and it is *simpler*: the GRIB
ones need `GribberishCodec` as serializer, this needs only the standard `BytesCodec` + `BloscCodec`.

### Verified units and value ranges

Decoded from a real 2026-08-07 12Z chunk (lead index 4; level index 6 = 400 hPa):

| Source var | Range observed | Unit — confirmed |
|---|---|---|
| `2m_temperature` | 205.9 – 319.7 | K |
| `mean_sea_level_pressure` | 94508 – 104132 | Pa |
| `total_precipitation_6hr` | max 0.041, mean 0.0007 | **m** (ERA5 convention) |
| `10m/100m_{u,v}_component_of_wind` | ±27 | m s-1 |
| `sea_surface_temperature` | 271.4 – 310.8, 34% NaN | K, NaN over land |
| `geopotential` | 60227 – 75406 @400 hPa | **m2 s-2** (/9.80665 = 7216 m ✓) |
| `specific_humidity` | 0 – 0.0055 | kg kg-1 |
| `vertical_velocity` | ±4.3 | Pa s-1 |
| `temperature` (pl) | 213.9 – 268.2 @400 hPa | K |

**`10m_wind_speed` is the mean of member wind *speeds*, not the speed of the mean vector** —
`speed - hypot(mean_u, mean_v)` is systematically positive (mean +0.16, max 7.23 m s-1), the
expected Jensen gap. So it is **not** derivable from the archived `u`/`v` and is a genuinely
distinct field. This is an intrinsic property of the variable → belongs in its `comment`.

## Two things a virtual dataset can and cannot fix

A dim with **chunk size 1** is freely remappable (each chunk is independently addressable); a
dim **inside** the chunk is not, short of decoding.

- **Can** present `pressure_level` **descending** `[1000…50]` to match `ecmwf/aifs_single`
  and HRRR, by mapping target level index → source level index per ref. Same for `lead_time`.
- **Cannot** flip latitude or roll longitude — both live inside a single chunk. So this store
  serves **ascending latitude and 0–360 longitude**, deviating from our global convention
  (`ecmwf/aifs_single/template_config.py:45` uses descending / −180–180). Accepted for the
  demo; the later materialized variant can normalize it. Note it in the validation report.

## Variable mapping

Target names follow existing equivalents exactly (`ecmwf/aifs_single/forecast_virtual`).
`ScaleOffset` filters chain on read (decode is `value / scale + offset`), so unit conversions
cost nothing — this is how AIFS serves Celsius and geopotential height.

### Root group — full archive, 2022-01-01 →

| Source | Target | Units | Filter |
|---|---|---|---|
| `2m_temperature` | `temperature_2m` | degree_Celsius | K→C |
| `10m_u_component_of_wind` | `wind_u_10m` | m s-1 | — |
| `10m_v_component_of_wind` | `wind_v_10m` | m s-1 | — |
| `100m_u_component_of_wind` | `wind_u_100m` | m s-1 | — |
| `100m_v_component_of_wind` | `wind_v_100m` | m s-1 | — |
| `10m_wind_speed` | `wind_speed_10m` | m s-1 | — |
| `100m_wind_speed` | `wind_speed_100m` | m s-1 | — |
| `mean_sea_level_pressure` | `pressure_reduced_to_mean_sea_level` | Pa | — |
| `sea_surface_temperature` | `sea_surface_temperature` | degree_Celsius | K→C |
| `total_precipitation_6hr` | `total_precipitation_surface` | kg m-2 | m→kg m-2 (`ScaleOffset(scale=0.001)`) |

### `pressure_level` group — 2025-01-01 →

| Source | Target | Units | Filter |
|---|---|---|---|
| `temperature` | `temperature` | degree_Celsius | K→C |
| `geopotential` | `geopotential_height` | m | ÷9.80665 |
| `specific_humidity` | `specific_humidity` | 1 | — |
| `u_component_of_wind` | `wind_u` | m s-1 | — |
| `v_component_of_wind` | `wind_v` | m s-1 | — |
| `vertical_velocity` | `vertical_velocity` | Pa s-1 | — |

Confirm each target name/attr set against the equivalent in `aifs_single/forecast_virtual`
before writing; match exactly rather than re-deriving.

**Not available in WeatherNext 2** (checked): any radiation flux, surface pressure, total cloud
cover, 2 m relative/specific humidity, 2 m dew point.

### Reference count

Root: 6719 inits × 60 leads × 10 vars ≈ **4.0M**. Levels: 2335 inits × 60 leads × 13 levels ×
6 vars ≈ **10.9M**. Total ≈ **15M refs**.

## Tasks

- [x] Explore source; report at `plans/weathernext2_exploration.md`
- [x] Checkpoint A — scope settled
- [x] Prove zarr-v2-blosc → zarr-v3 `BloscCodec` round-trip
- [x] Verify all asserted units against real chunk values
- [x] Scaffold `src/reformatters/google/weathernext2/forecast_virtual/`
- [x] `TemplateConfig` — dims/coords/vars, `pressure_level` group, descending levels
- [x] `RegionJob` — `discover_available()` + `file_refs()`, both store layouts
- [x] `DynamicalDataset` — storage config, cronjobs (ship `suspend=True`), validators
- [x] Register in `__main__.py`; run the `generate-manual-workflows` prek step
- [x] `update-template`; run `common_template_config_subclasses_test.py` + `datasets_cf_compliance_test.py`
- [x] Integration test with snapshot values
- [x] Code review pass
- [ ] **Human review + merge PR #870** — the backfill runs from `main`
- [ ] **Blocker: GCS credentials in the cluster.** `storage.py` uses
      `icechunk.gcs_from_env_credentials()` and `region_job.py` uses `gcs_store()`, both reading
      from the environment, but the jobs pass only `store_factory.k8s_secret_names()` (the R2
      secret). A backfill or update pod therefore has no way to read `gs://weathernext/`.
      Needs: a Google service account with read on that bucket, a k8s secret holding it, and
      that secret name added to the backfill job's and both cronjobs' `secret_names`.
- [ ] Backfill
- [ ] Validate

## Open questions for the WeatherNext team (ask at the demo, not blocking)

1. Is `total_precipitation_6hr` accumulated over `[valid_time − 6h, valid_time]`? Assumed yes.
2. Confirm `*_wind_speed` is the mean of member speeds (our measurement says yes).
3. Is it the same model across the 2025 boundary? The yearly stores carry `model_wid` 43/44/45;
   2025+ stores have **empty root attrs**, so provenance is unrecoverable from the data. Since
   our surface variables span that boundary, this needs a **validation-report review note**
   (a time-windowed archive characteristic, so not a variable `comment`).
4. Why did coordinate attrs disappear between the 2025-10-01 and 2025-11-01 inits?

### 2026-08-07 — Explore + Checkpoint A

Explored `gs://weathernext/`; report written. Verified independently of the explore agent: the
2025 chunk-grid change (levels 13→1, ensemble sample 4→1), surface-grid uniformity across all
four eras, the blosc round-trip, all units, and the ascending level order. Scope settled with
marsh: mean product, surface full archive + levels 2025+, `google-weathernext2-forecast-virtual`,
assert-and-confirm units. Branch + draft PR #870 opened with the scaffold.

Note: running repo-wide `uv run ruff format` from the repo root walks into the untracked
`.worktrees/` checkout and reformatted 2 files belonging to another branch's work-in-progress.
Scope lint/type commands to `src/reformatters/google tests/google` for the rest of this work.

### 2026-08-07 — Implement

Implemented `src/reformatters/google/weathernext2/forecast_virtual/` (template config, region
job, dynamical dataset), registered it, generated `templates/latest.zarr`, and wrote four test
files. Class names dropped the scaffold's `Spatial` infix. Everything green:
`ruff format`/`check`, `ty check`, and the full `uv run pytest` suite.

**The source bucket is NOT anonymously readable.** `https://storage.googleapis.com/weathernext/...`
returns 403 and the JSON API 401. This was never checked during explore (all reads went through
authenticated `gcloud storage`). Consequences:

- `download.py` gained a `gcs_store()` helper, and obstore needs `GOOGLE_APPLICATION_CREDENTIALS`
  pointing at a credentials file. `GOOGLE_SERVICE_ACCOUNT_PATH` does *not* work with a gcloud ADC
  authorized-user file (object_store insists on a service-account JSON there).
- `storage._repository_config_and_credentials` explicitly rejected non-S3/non-local virtual chunk
  containers. It now maps a GCS container to `icechunk.gcs_from_env_credentials()`. Verified
  end-to-end against real chunks.
- **Readers of the published store will need Google credentials.** Fine for an internal demo; it
  makes any public publication contingent on the license question (already open) *and* on the
  bucket's access policy.
- Operationally: the update pod will need those credentials in the environment. Cronjobs ship
  `suspend=True`, so this is a pre-enable task, not a blocker.

**Corrections to the plan:**

1. `total_precipitation_6hr` maps to **`total_precipitation_surface`**, not `precipitation_surface`.
   In this repo `precipitation_surface` is a *rate* (`kg m-2 s-1`, `standard_name=precipitation_flux`,
   `step_type=avg`); a per-step accumulation amount is `total_precipitation_surface`
   (`kg m-2`, `precipitation_amount`, `accum`) — see `noaa/hrrr/forecast_virtual_template_config.py`
   and the naming note in `noaa/gefs/forecast_10_day_spatial/template_config.py`.
2. `DynamicalDataset` asserted `compressors == ()` on every virtual data var. That invariant was
   GRIB-shaped; a zarr-backed reference must declare the *source's* compressor. Relaxed to
   `compressors is not None` (an unset value would let zarr stack its own default on the source
   bytes), and `tests/common/common_template_config_subclasses_test.py`'s
   `test_virtual_serializers_flip_to_shared_orientation` now skips vars with no serializer.
   `docs/virtual_datasets.md` and `example_virtual/template_config.py` updated to match.
3. `sea_surface_temperature` in `degree_Celsius` needed a `CF_UNITS_VARIANCES_ALLOWLIST` entry.
4. `license` is a `Literal["CC-BY-4.0"]` in `config_models.py`, so the template asserts CC-BY-4.0.
   The plan's "license deliberately not researched" still stands as an open question — the type
   left no way to record "unknown".
5. Publication lag (init+6h20m..6h50m) **exceeds** the 6h cycle, so the newest template init is
   always unpublished at fire time. Cron fires at init+6h55m (`55 0,6,12,18`) with a 30 min
   `pod_active_deadline`: the target cycle is ingested on the first tick and the fire then polls
   for the next cycle's slot until the deadline and rolls it over. Window 30h (4 cycles).
   Validation at `55 1,7,13,19`, `max_latest_init_time_age=9h`.

**Design decisions worth reviewing:**

- **One `SourceFileCoord` per init time** (a whole source store), not per (init, lead). The source
  publishes a store atomically behind a `success` marker, so this matches the source's own
  granularity: 6719 coords instead of ~403k, one commit per init, one manifest probe per init.
  `out_loc()` names a representative `lead_time` (6h) for the per-file probe, the same trick HRRR
  uses for a representative `pressure_level`.
- **`obstore` matches a listing prefix by whole path component**, so a yearly store's
  `<var>/<init>.` prefix lists nothing. `_list_chunk_sizes` lists the variable's directory with
  `offset=<var>/<init>` and `takewhile`s the matching keys — they are lexicographically contiguous
  — so one request returns exactly that init's 60 chunk sizes instead of the year's 87,600.
- **An exact object path is not a listable prefix either**, so `discover_available` probes the
  success marker with `obstore.head` (deduped: yearly-store coords share one marker) rather than
  reusing `discover_available_by_obstore_listing`.
- `discover_available` returns file size `0` (`_NO_SINGLE_FILE_SIZE`): a coord is a whole store with
  no single length, and each reference's length is its own chunk object's size from the listing.

**Test coverage split** (the source is private, so CI cannot read it):

- `template_config_test.py` — structure/metadata, plus `test_encoding_decodes_bytes_the_source_wrote`,
  which blosc-encodes a synthetic chunk with **numcodecs** (the source's encoder) and decodes it
  through the variable's declared zarr v3 pipeline. This is the plan's blosc round-trip claim,
  pinned in CI with no network.
- `region_job_test.py` — URLs, markers, chunk keys, era-dependent variable availability, reference
  routing incl. the descending→ascending level map, marker gating, the update window.
- `dynamical_dataset_test.py` — suspended cronjobs, validators, manifest splits, container prefix.
- `real_files_integration_test.py` — **gated on `WEATHERNEXT_CREDENTIALED_TESTS=1`** (the pattern
  `tests/nasa/imerg/real_files_integration_test.py` uses). Backfills one init per era against real
  GCS and asserts snapshot values at 0N/0E, lead 6h: 2025-03-01T06 `temperature_2m` 28.20932 degC,
  `total_precipitation_surface` 0.46721694 kg m-2, `pressure_level/temperature` @850 18.29538 degC;
  2022-01-02T12 (yearly store, init index 6) `temperature_2m` 26.838043 degC, `wind_speed_10m`
  4.212481, and `pressure_level/temperature` NaN. Both pass.

**Remaining gaps:** CI never reads the real source (private bucket); the backfill's GCS credential
story for the k8s pod is unresolved; manifest bytes/ref is still the HRRR-derived 16.4 B estimate and
should be measured once data lands.

### 2026-08-07 — Code review fixes

Applied the agreed fix list from the review (commit on `weathernext2-forecast-virtual`).

**License.** `config_models.License` gained `"LicenseRef-Google-WeatherNext"` (an SPDX
custom-reference identifier) and the template now declares it, so the archive no longer
asserts CC-BY-4.0 over data whose terms are unstated. Open question 5 for the WeatherNext
team: what are the actual terms?

**`file_refs` now asserts its chunk count.** Per (init, variable) it requires exactly
`len(lead_time) * len(levels)` matched chunk objects — 60 for a root var, 780 for a
pressure-level var — and raises otherwise. `_list_chunk_sizes`'s `offset` + `takewhile`
scan depends on lexicographic listing order, which GCS guarantees but object_store does
not, and a short scan previously dropped refs while still committing the coord. Pre-2025
inits are unaffected: `generate_source_file_coords` filters pressure vars out by
`date_available`, so they never reach the assert. Verified against the real source — the
credentialed integration test passes with the assert in place, so the source really does
publish every chunk of every variable behind the success marker.

This write-time assert is the dataset's only real completeness gate. `out_loc()` returns a
single `Mapping`, so one-coord-per-init structurally cannot tell `CheckVirtualManifestCompleteness`
or `manifest_scan.py` that it covers 60 leads; both verify one representative ref per init.
Widening `out_loc` is a base-class change affecting every dataset — out of scope, and worth
a **validation-report review note**.

**Virtual encoding invariant.** `DynamicalDataset` now asserts the exact pairing
`(serializer is None) != (compressors == ())` (kept alongside `compressors is not None`,
which the pairing alone does not cover: a zarr-source var with `compressors=None` satisfies
the xor). This catches a GRIB var with a stray compressor and a zarr-source var with
`compressors=()`.

*Review correction:* the pairing rejects a third legitimate shape the review did not
account for — raw uncompressed source bytes read back by zarr's default `BytesCodec`,
which is exactly what all three synthetic virtual test harnesses
(`virtual_region_job_test`, `virtual_multi_group_test`, `dynamical_dataset_test`) use. They
now declare an explicit `BytesCodec(endian="little")` serializer (`RAW_BYTES_SERIALIZER`),
which is behaviourally identical and honest about how the source bytes decode. The
invariant is therefore really "every virtual var declares explicitly what decodes the
source bytes", which is a better rule than the GRIB-vs-zarr dichotomy it was framed as.

**Orientation in the shared validation scripts.** Added
`scripts.validation.utils.align_longitude_convention`, which relabels the reference into
the validation dataset's longitude convention (either direction) and re-sorts. Wired into
`compare_spatial.align_reference_spatially` (its `longitude=slice(0, 359.75)` cropped the
reference to its eastern half) and into `compare_timeseries._load_timeseries_for_var`
(its `method="nearest"` clamped every point east of 180 E to the reference's 179.75).
`compare_spatial`'s descending-latitude slice is correct as-is. New tests in
`utils_test.py`, `compare_spatial_test.py` and `compare_timeseries_test.py` cover both
conventions and both fail if the normalization is removed (checked); the shared grid
fixtures live in `tests/scripts/validation/grids.py`.

**Nice-to-haves.** `download.py`'s three copies of the obstore client/retry block collapsed
into `_CLIENT_OPTIONS` / `_RETRY_CONFIG`. The cron-timing narrative now lives only in
`dynamical_dataset.py`; `region_job.py`'s comment covers just the window sizing. Dropped the
"Descending like ecmwf-aifs-single..." dataset citation from `template_config.py`. Fixed
`region_job_test.py`'s "6th init" to "7th ... at index 6". `test_encoding_decodes_bytes_the_source_wrote`
is parametrized over the root and pressure-level vars, pinning that the trailing size-1
level dim leaves the chunk's memory layout identical to the source's (lat, lon) buffer.

**`offset` pushdown: confirmed server-side, no caching needed.** object_store's GCS
`ListClient::list_request` talks the GCS **XML** API (`list-type=2`) and passes `offset`
as `start-after`, so the ~43,800 per-(init, var) listings each return only that init's
keys — GCS does not page the year from the offset. (It also means the ordering assumption
is the XML ListObjectsV2 contract, i.e. lexicographic by UTF-8 key.) If this had not been
pushed down the fix would have been a per-(store, var) listing cache.

**Checks.** `uv run ruff format`/`check --fix`/`ty check` scoped to
`src/reformatters tests src/scripts` — all clean. `uv run pytest`: 2104 passed, 12 skipped
(2093/12 before; +11 new tests). `common_template_config_subclasses_test.py` +
`datasets_cf_compliance_test.py`: 342 passed. Template regenerated (license attr only).
PR #870 left as a draft.

### 2026-08-07 — Ready for review; Backfill blocked

Review fixes applied and verified independently: `uv run pytest` 2104 passed / 12 skipped,
and `ruff format --check`, `ruff check`, `ty check` clean over
`src/reformatters tests src/scripts`.

Two findings from the review worth carrying forward:

- **Completeness for this dataset is enforced at write time, not by validation.** One
  `SourceFileCoord` per source store means `out_loc()` reports a single position, so both
  `CheckVirtualManifestCompleteness` and the offline `manifest_scan.py` verify one
  representative ref per init — a store missing 40 of its 60 leads still reports 100%.
  `file_refs()` now asserts it found `len(lead_time) * len(levels)` refs per variable
  (60 root / 780 pressure), which is the real gate. Record this in the validation report.
- **Orientation leaked into shared tooling.** Ascending latitude / 0–360 longitude broke
  `compare_timeseries.py` (points east of 180°E clamped to the reference's 179.75) and
  `compare_spatial.py` (reference cropped to its eastern half). Fixed with
  `utils.align_longitude_convention`, tested in both directions. Nothing in
  `src/reformatters/common/` assumed an orientation.

The virtual-encoding invariant is now `(serializer is None) != (compressors == ())` alongside
`compressors is not None`. That rejects a third shape — raw uncompressed bytes on zarr's default
`BytesCodec` — so the synthetic virtual test harnesses declare an explicit
`BytesCodec(endian="little")`. A future uncompressed non-GRIB source must do the same rather than
relying on the default.

Blocked on a human for the two items added to Tasks above.

### 2026-08-08 — Paused; PR closed

Closed PR #870 without merging. Branch `weathernext2-forecast-virtual` is pushed and intact;
reopen the PR to resume.

Code is complete and green (2104 passed / 12 skipped; ruff + ty clean). What remains is
entirely credentials:

1. **A GCS service account that can read `gs://weathernext/`.** Access today is as the user
   principal `marshall@upstream.tech`; a service account is a different principal and inherits
   nothing, and the bucket is Google's, so the grant has to come from the WeatherNext team.
   Marsh is asking them. Without it a backfill or update pod cannot read the source at all.
2. **A separate R2 output bucket**, in place of `s3://upstream-gridded-zarrs` — marsh's
   preference, to keep demo data out of production infrastructure. Needs the bucket, an R2 API
   token scoped to it, and a secret whose contents are keyed for `icechunk.s3_storage(**options)`
   (`access_key_id`, `secret_access_key`, `endpoint_url`, `region`).

Two code changes were identified but **not** written, since both depend on names that don't
exist yet:

- A new `StorageConfig` subclass for the demo bucket, replacing
  `UpstreamGriddedZarrsIcechunkDatasetStorageConfig` on this dataset.
- An extra-env hook on `common/kubernetes.py`'s `Job`, whose env list is hardcoded, to pass
  `GOOGLE_APPLICATION_CREDENTIALS=/secrets/<name>.json` and to add the GCS secret to the job's
  mount list. The mount mechanism itself already fits: secrets mount as JSON files at
  `/secrets/<name>.json`, and a service-account key is a JSON file. The GCS secret must stay
  out of the storage config — `load_secret()` parses contents as a storage-options dict.


mrshll:
Pausing until credentials exist — closing rather than leaving a stale draft open.

Code is complete and green (2104 passed / 12 skipped; ruff + ty clean). Branch `weathernext2-forecast-virtual` is pushed and intact; reopen this PR to resume. Full state, including the two unwritten code changes, is in the PR description.

Blocked on:
1. A GCS service account that can read `gs://weathernext/` — current access is the user principal `marshall@upstream.tech`, and a service account inherits nothing from it, so the grant has to come from the WeatherNext team.
2. A separate R2 output bucket for the demo store, plus its API token.

mrshll:
Superseded by fresh draft PR #931 for the approved full-ensemble scope. The working changes were preserved and moved to the new branch; no backfill was started.

PR #873, opened 2026-08-10
AIFS Single: read the transitional experimental/ source path

mrshll:
@aldenks fixed this gap in virtual -- up to you if you want to resolve it historically in materialized. just a few inits

---

The materialized `ecmwf-aifs-single-forecast` carries the same two-phase source path assumption that #869 fixed in the virtual dataset, so **2025-02-24T06 through 2025-02-25T18 are all-NaN in the published store** — seven consecutive forecast cycles.

Verified against the live store before this change:

```
2025-02-24T00:00  data mean 4.2      2025-02-25T06:00  all-nan
2025-02-24T06:00  all-nan            2025-02-25T12:00  all-nan
2025-02-24T12:00  all-nan            2025-02-25T18:00  all-nan
2025-02-24T18:00  all-nan            2025-02-26T00:00  data mean 4.3
2025-02-25T00:00  all-nan
```

The published validation report currently tells users this data **"cannot be recovered"** from the source. That is false — all seven cycles are in ECMWF's archive under a path we never constructed. Correcting the report is follow-up work, tracked separately from this code fix.

## The change

`AIFS_SINGLE_PATH_CHANGE_DATE = 2025-02-26T00:00` resolved these inits to `aifs/0p25/oper/`, which holds no forecast GRIBs then. The real layout is three-phase:

| path | inits |
|---|---|
| `aifs/0p25/oper/` | through 2025-02-24T00 |
| `aifs-single/0p25/experimental/oper/` | 2025-02-24T06 → 2025-02-25T00 |
| `aifs-single/0p25/oper/` | 2025-02-25T06 onward |

Both variants now resolve the stream through one `aifs_single_stream_path()` helper, and the two boundary constants live in the shared `ecmwf/aifs_single/template_config.py`. The virtual dataset previously defined its own copies; a value this subtle should not be duplicated across two datasets that read the same source.

**The boundary also moves `expanded_vars_date`**, which drives two things in the materialized dataset besides `date_available`: most importantly the `InitTimeScaleFactor(scale_factor=1000, end=expanded_vars_date)` that converts metres-era precipitation. That has to move in lockstep — the recovered inits are already in kg m-2 (verified by decode: `tp` at lead 24h jumps from max 0.239 to 228.7 exactly at 2025-02-24T06), so leaving the scale boundary at 2025-02-26 would multiply them by 1000.

## Verification

Every init across the transition resolves with the corrected URLs, on both mirrors:

```
init                 s3   gcs  path
2025-02-23 18:00    206   404  aifs/0p25/oper
2025-02-24 00:00    206   404  aifs/0p25/oper
2025-02-24 06:00    206   206  aifs-single/0p25/experimental/oper
2025-02-25 00:00    206   206  aifs-single/0p25/experimental/oper
2025-02-25 06:00    206   404  aifs-single/0p25/oper
2025-02-26 00:00    206   404  aifs-single/0p25/oper
```

The GCS 404s are not a regression — the mirror has a genuine hole here (it lacks `aifs/` entirely in this window and only resumes `oper/` at 2025-02-27T06), and `ecmwf_download_with_fallback(("gcs", "s3"), ...)` already falls back to S3 on missing files. The pre-existing `aifs/` era relies on that same fallback today.

- `update-template` produces **no diff** for either dataset — `date_available` and `init_time_scale_factors` are internal attrs, so no metadata change and no re-backfill of existing positions is implied.
- `ruff format` / `ruff check` / `ty check` clean
- `uv run pytest tests/ecmwf/aifs_single/` — 71 passed
- `common_template_config_subclasses_test.py` + `datasets_cf_compliance_test.py` — 325 passed

The old test asserting `2025-02-25T12` uses `aifs/` encoded the bug; it is replaced by a parametrized test covering all three phases on both S3 and GCS.

## After merge

This only changes where we look. Recovering the data needs an `overwrite-chunks` backfill filtered to the seven timestamps, then the published validation report's "cannot be recovered" note should be corrected.

### DIFF
--- a/src/reformatters/ecmwf/aifs_single/forecast/region_job.py
+++ b/src/reformatters/ecmwf/aifs_single/forecast/region_job.py
@@ -28,6 +28,7 @@ from reformatters.common.types import (
     Timedelta,
     Timestamp,
 )
+from reformatters.ecmwf.aifs_single.template_config import aifs_single_stream_path
 from reformatters.ecmwf.ecmwf_config_models import EcmwfDataVar, vars_available
 from reformatters.ecmwf.ecmwf_grib_index import grib_message_byte_ranges_from_index
 from reformatters.ecmwf.ecmwf_utils import (
@@ -36,9 +37,6 @@ from reformatters.ecmwf.ecmwf_utils import (
 )
 
 log = get_logger(__name__)
-
-# Path changed from aifs/ to aifs-single/ on this date
-AIFS_SINGLE_PATH_CHANGE_DATE = pd.Timestamp("2025-02-26T00:00")
 
 # GRIB master table version changes caused metadata differences for precipitation.
 # Early data (table v27): generic product template codes.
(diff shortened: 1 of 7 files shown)


aldenks (approved):
This is great. It'd be ideal to backfill materialized and virtual and update val report

PR #877, opened 2026-08-11
Handle pod eviction in operational updates gracefully

aldenks:
## Summary
Three sources of timeout alerts on `noaa-hrrr-forecast-18-hour-virtual-update`, none of which were data problems: a replacement pod adopting the next fire's work, a finished run whose terminal check-in was never delivered, and a pod stopped by Kubernetes exiting silently.

## Key Changes

- **An update is bounded by its scheduled fire, not by process start.** `DynamicalDataset` passes the update cron's `previous_fire_time` as `job_fire_time` to `operational_update_jobs`, and `VirtualRegionJob` ends its window there. Every pod of one fire then derives the same work, so a pod replacing an evicted one covers its predecessor's steps rather than adopting the next fire's, whose files publish after the deadline it inherits — such a pod previously idled until that deadline and reported nothing.

- **Terminal check-ins flush when sent.** `monitor_cron` flushes after `ok`/`error`, not after `in_progress`, so no latency is added at run start. A run that completed its work and lost its check-in to the at-exit flush currently alerts as a timeout.

- **SIGTERM is logged.** `install_sigterm_logger()` logs, flushes telemetry, and exits, so a pod stopped by Kubernetes (eviction, pod active deadline, job replacement) says so. A Python signal handler runs between bytecodes, so a run that ends *without* that log was killed outright or stuck in a call that never returned — a distinction that is otherwise invisible.

- **Termination grace period 5s → 30s**, enough for an in-flight write to land and the log to flush. It is now a single `TERMINATION_GRACE_PERIOD` constant, also used to bound that flush.

## Implementation Details

- `job_fire_time` is an optional keyword: `DynamicalDataset` inspects the callee's signature and passes it only when accepted, so the 12 materialized `operational_update_jobs` implementations are untouched and can opt in one at a time. The convention is documented on `RegionJob.operational_update_jobs`, where implementers look.
- `os._exit` in the SIGTERM handler skips unwinding, which would join the 64-thread download pool and can outlast the grace period; the Sentry flush is therefore explicit on the line above rather than left to the at-exit hook.
- Tests: `update()` runs end to end with `previous_fire_time` patched and `now` an interval later, asserting the job's template stops at the fire's step (verified to fail without the anchor); the SIGTERM test asserts the log, the flush, and exit code 143; a unit test pins the signature-inspection rule, including the `**kwargs` case that keeps four dataset test wrappers exercising the fire-time path. Two dataset fixtures had frozen `now` at moments when no update cron fires and now use real fire times.
- Docs: `docs/virtual_datasets.md` on the fire-anchored window, `docs/ops_card.md` on reading the SIGTERM log when an update times out.

## Not included

A watchdog for the fourth failure mode — a run that stops making progress with no signal at all — is deliberately left out; the SIGTERM log now separates a stopped run from a stuck one, which is the input that decision needs.

https://claude.ai/code/session_01JRxcPKwiV9xXaJhgr6ZM3h


aldenks:
Code Quality is red on a failure that predates this branch: `tests/noaa/mrms/conus_analysis_hourly/region_job_test.py::test_download_and_read_precipitation[pre-v12-iowa]` (1 failed, 2024 passed). It fetches a 2019 fixture file from `mtarchive.geol.iastate.edu` and gets a 404.

I checked out `origin/main` and ran that test locally — it fails identically there, so it isn't caused by these changes. The archive's root responds 200 but every dated path I probed 404s, including recent ones, so this looks like a source-side outage or path change rather than one pruned file.

Not fixing it here: it belongs to an unrelated dataset and the fix (re-pointing the fixture, or not depending on a live archive in a test) is its own change. I'll re-run CI when the archive recovers.

---

PR #890, opened 2026-08-13
Add materialized vertical-group support

mrshll:
Prerequisite for ECMWF Extended Range (dynamical-org/meta#186).

Materialized datasets could not declare variables in vertical groups — `dynamical_dataset.py` hard-asserted against any non-root variable. This makes the materialized write path group-aware and removes that assert.

## The change

The gap was one inconsistency: the materialized pipeline keyed variables by bare **name**, while everything already group-aware (`DataVar.path`, `walk_data_arrays`, `flatten_groups`, the virtual write loop) keys them by **path**. Rather than threading a `group` parameter through each function, the materialized path now uses the same key.

- `_get_region_datasets` builds from `flatten_groups(template_ds)[var_paths]` instead of subsetting the root node by name
- `process` passes `data_var.path` to `create_data_array_and_template` and `copy_data_var`, and as the `results` key
- `write_shard_to_zarr` splits the path and writes into that zarr group
- `copy_data_var` uses the path as both the `template_ds` key and the store-relative chunk directory
- `config_models.split_var_path` added as the inverse of the existing `var_path`

This also fixes a latent bug: `results` keyed by bare name would have silently merged the source-file results of two same-named variables in different groups.

## Root-only datasets are unchanged

Structurally, not by inspection — there is no root-only branch to get wrong. At root `path == name`, `flatten_groups` of a single-node tree is its root dataset, `split_var_path("name")` is `(None, "name")`, `group=None` is `to_zarr`'s default, and the chunk directory string is identical. No `templates/latest.zarr` is touched.

## Tests

`tests/common/materialized_multi_group_test.py` — written first, failed with `KeyError: 'temperature'`. A materialized template with a root variable plus **two** vertical groups whose variables share the bare name `temperature`, driven through `get_jobs` → `process` over both shards; asserts disjoint per-group values read back from the primary store and that `pressure_level` opens standalone.

`uv run pytest`: 2044 passed, 10 skipped, 0 failed. `ruff format`/`check` and `ty check` clean over `src tests`.

## Not for merge yet

Held as a draft deliberately. ECMWF Extended Range (6 variables × 10 pressure levels, 101 members) is built on this branch and is its stress test — this reaches `main` once a real dataset has exercised it end to end, rather than landing in shared infrastructure ahead of any user.


mrshll:
@codex review

src/reformatters/common/config_models.py
```diff
@@ -24,6 +24,12 @@ def var_path(group: Group, name: str) -> str:
     return f"{group}/{name}"
 
 
+def split_var_path(path: str) -> tuple[str | None, str]:
```
aldenks, line 27: Add unit tests
- root var (group = None)
- pressure_level/temperature 
- nested: foo/bar/baz/temperature, group = foo/bar/baz name = temperature

src/reformatters/common/materialized_region_job.py
```diff
@@ -276,9 +276,10 @@ def process(
         return results
 
     def _get_region_datasets(self) -> tuple[xr.Dataset, xr.Dataset]:
-        # Materialized datasets are single-level (DynamicalDataset enforces it), so the
-        # root node holds every var; subset it to this job's vars by name.
-        ds: xr.Dataset = self.template_ds.to_dataset()[[v.name for v in self.data_vars]]  # ty: ignore[invalid-assignment]
+        # Keyed by var path so a job spanning vertical groups keeps vars with the same
+        # bare name (e.g. pressure_level/temperature, model_level/temperature) distinct.
```
aldenks:
```suggestion
        # Key by var path so the same var in different groups is distinct
```

src/reformatters/common/dynamical_dataset.py
```diff
@@ -902,15 +902,6 @@ def _validate_virtual_storage(self) -> Self:
...
-        else:
-            # The materialized chunk-write path (zarr.copy_data_var, write_shards) is
-            # not yet group-aware; only virtual datasets support vertical groups today.
-            grouped = [
-                v.path for v in self.template_config.data_vars if v.group is not ROOT
-            ]
-            assert not grouped, (
-                f"materialized datasets do not yet support vertical groups: {grouped}"
-            )
         return self
```
aldenks, line 903: Dropping this guard exposes two materialized-only validators that aren't group-aware. Both are appended below only `if not is_virtual`, so the virtual multi-group dataset never hit them and no test covers them — the validation cronjob for the first materialized grouped dataset would crash rather than report.

- `check_for_expected_shards` (validation.py:588) passes each var's own dims into `iterating.dimension_slices`, which does `var.dims.index(dim)` for *every* var in the flattened dataset — on `pressure_level`, `temperature_2m.dims.index("pressure_level")` raises `ValueError: tuple.index(x): x not in tuple`.
- `compare_replica_and_primary` (validation.py:530) builds `non_append_dim_slices` from all of `replica_ds.sizes`, then `isel`s that same dict per variable: `ValueError: Dimensions {'pressure_level'} do not exist`. It fails for every variable — root has neither level dim, each group var lacks the other's. Restricting the dict to `replica_ds[var].dims` covers it.

Both reproduce.

---

aldenks:
`src/reformatters/common/materialized_region_job.py:149`

```diff
-            for var_name, coords in results.items():
-                worker_results.setdefault(var_name, []).extend(
+            for var_path, coords in results.items():
+                worker_results.setdefault(var_path, []).extend(
```

---

aldenks:
`src/reformatters/common/region_job.py:202`

```diff
-            Mapping from variable names to their SourceFileResult with final processing status.
+            Mapping from variable paths to their SourceFileResult with final processing status.
```

---

src/reformatters/common/dynamical_dataset.py
aldenks: Compare replica and primary doesn't have a good reason to be virtual only anymore either: ideally make it auto apply to all datasets that have a replica

(check_for_expected_shards is a materialized only check)

aldenks:
@mrshll  Approved after comments above

src/reformatters/common/config_models.py
mrshll: Added — split_var_path now has three tests: root, pressure_level/temperature, and the nested foo/bar/baz case.

src/reformatters/common/materialized_region_job.py
mrshll: Taken as suggested.

src/reformatters/common/dynamical_dataset.py
mrshll: Both fixed and covered by tests: check_for_expected_shards now reads each variable's own shape and shard encoding instead of scanning every variable, and compare_replica_and_primary slices by the variable's own dims.

mrshll: Done — compare_replica_and_primary now runs for any dataset with a replica, virtual included, and only check_for_expected_shards stays materialized-only.

mrshll:
Also applied both diffs above: var_name is now var_path in process_worker_jobs, and the update_template_with_results docstring says variable paths.

PR #900, opened 2026-08-15
Make the HRRR virtual modules dataset-shape-agnostic, and let virtual validators target variables and positions

aldenks:
Preparation for a virtual HRRR analysis (#901), which reuses the forecast datasets' variable catalog and ref-building but has a different shape and update schedule. No new dataset here, and no behavior change to the existing ones.

## Shared HRRR virtual modules

Split into a base that knows nothing about forecast dims plus forecast subclasses, dropping the now-inaccurate `forecast_` prefix from the file names:

- `forecast_virtual_template_config.py` → `virtual_template_config.py`, with `NoaaHrrrVirtualTemplateConfig` (catalog, vertical groups/coords, grid) under `NoaaHrrrForecastVirtualTemplateConfig` (init_time/lead_time dims, forecast coords).
- `forecast_virtual_region_job.py` → `virtual_region_job.py`, with `NoaaHrrrVirtualSourceFileCoord` / `NoaaHrrrVirtualRegionJob` (NODD S3 discovery, GRIB-index ref building) under the forecast coord and job.

The chunk shape is no longer hard-coded to the forecast's leading dims. Each catalog group takes it as an argument and binds it once, so the variables are built with their real shape — the same way `noaa-gefs-forecast-10-day-spatial` passes `message_chunks` into its encoding builder. The config computes the shape per group from its own dims, and builds only the groups it declares, so a variant omitting a vertical group no longer has to filter the catalog afterward. Most of this file's diff is the mechanical call-site rename that binding produced.

**The regenerated 18-hour and 48-hour forecast templates are byte-identical**, which is the main thing to check when reviewing this half.

## `out_loc` vs. the manifest probe cell

`out_loc()` was carrying two jobs: the output slab a file fills, and the single cell `filter_already_present` probes to decide whether the file is already ingested. Those only coincide when the file pins every multi-chunk dim, so a group-only file (HRRR `wrfprs`/`wrfnat`, AIFS pressure-level) had to smuggle a representative level into `out_loc` and then have each ref override it.

`out_loc` is now purely the slab. The probe cell moves to `VirtualRegionJob.representative_probe_loc(coord, var)`, beside `representative_var` — the two halves of the same "one cell stands for the whole file" idea — and its default derives the cell instead of hardcoding it: `out_loc` plus the first label along each multi-chunk dim `out_loc` leaves unpinned. HRRR's `_REPRESENTATIVE_LEVEL` / `_vertical_probe_loc` and AIFS's `_REPRESENTATIVE_PRESSURE_LEVEL` are deleted; the derived probe keys are identical to the ones on `main`, so neither existing store re-ingests anything.

This also fixes `CheckVirtualDecodeHealth`'s level sampling as a side effect: it selects the chunk to decode with `out_loc`, so the smuggled level collapsed the vertical dim before `_sample_levels` ran, and group variables were only ever decode-checked at 1000 hPa. They now get the intended `sampled_levels` spread.

## `CheckVirtualManifestCompleteness` variable filtering

Takes `include_vars` / `exclude_vars`, matching the signature the xarray validators already use, restricting a check to the source files carrying those variables. A dataset whose variables become available on different schedules can then run one instance per group, each held to a whole `1.0` from the position where its own files are expected, instead of one instance with a fractional leading tier — which buys tolerance for the group that has not published yet by also tolerating a genuine gap in the group that has. The split must fall on source-file boundaries, since presence is probed through each file's representative variable; that precondition is asserted and tested. Pass/fail messages name the selection, and a filter that matches no files fails rather than vacuously passing.

`CheckVirtualDecodeHealth`'s behavior is unchanged.

## Docs and tests

`docs/virtual_datasets.md` picks up the `out_loc` / `representative_probe_loc` split and the variable filters; the surrounding prose this branch touches is tightened to contracts rather than how-to narration (net +4/−2 lines).

New tests cover the probe-cell contract in `tests/common`, mixed-checked/unchecked-file rejection on the multi-group fixture, and the completeness zero-tier and variable-selection cases. The AIFS and HRRR 48-hour probe tests now assert the full probe cell rather than just `out_loc`.

## Testing

`ruff format`, `ruff check`, `ty check` clean. Full suite passes (2052 tests, including the slow HRRR/AIFS virtual tests). Forecast templates regenerate with no diff.


docs/virtual_datasets.md
```diff
@@ -49,10 +49,10 @@ Virtual datasets are *metadata-heavy*, not storage-heavy. One GRIB message — o
 **On the `SourceFileCoord` subclass** (identifies one source file):
 
 - `get_url() -> str` — the canonical data-file URL refs point at (must match the virtual chunk container prefix).
-- `out_loc() -> {dim: label}` — the output cell the file fills. The inherited default works when the coord's fields are all dim names; override otherwise (e.g. a representative ensemble member for a multi-member file). A file holding *only* vertical-group variables (no root var) must include a representative level (e.g. `pressure_level: <a real level>`) so the per-file manifest probe — which resolves `out_loc` to a single chunk — has a concrete level to probe; per-file commit atomicity makes that one level's presence imply the whole file's. The per-ref `out_loc` overrides this representative level with the ref's actual level. (HRRR `wrfprs`/`wrfnat` files are group-only.)
+- `out_loc() -> {dim: label}` — the output slab the file fills: a label per dim the file pins, omitting any dim it spans wholly. The inherited default works when the coord's fields are all dim names; override otherwise.
```
aldenks:
```suggestion
- `out_loc() -> {dim: label}` — the output region the file fills: a label per dim the file pins, omitting any dim it spans wholly. The inherited default works when the coord's fields are all dim names; override otherwise.
```

docs/virtual_datasets.md
```diff
@@ -49,10 +49,10 @@ Virtual datasets are *metadata-heavy*, not storage-heavy. One GRIB message — o
...
 - `get_index_url() -> str` — *optional*, only if the files have a sidecar byte-range index; used by the obstore-listing discovery util and by index-based `file_refs`. Nothing on the base calls it.
 
-**Optional overrides (working defaults):** `filter_already_present` (manifest probe), `representative_var` (first instant var among the coord's own `data_vars` — only override for a packing that rule misses), the `tick_interval` (1s) / `download_concurrency` (64) class attrs, and `process_virtual_refs` itself (only for a fundamentally different batching policy).
+**Optional overrides (working defaults):** `filter_already_present` (manifest probe), `representative_var` / `representative_probe_loc` (the variable and cell standing for the whole file — first instant var among the coord's own `data_vars`, at `out_loc` plus the first label of each dim it leaves unpinned; only override for a packing those rules miss), the `tick_interval` (1s) / `download_concurrency` (64) class attrs, and `process_virtual_refs` itself (only for a fundamentally different batching policy).
```
aldenks:
```suggestion
**Optional overrides (working defaults):** `filter_already_present` (manifest probe), `representative_var` / `representative_probe_loc` (the variable and cell standing for the whole file — first instant var among the coord's own `data_vars`, at `out_loc` plus the first label of each dim it does not specify; only override for a packing those rules miss), the `tick_interval` (1s) / `download_concurrency` (64) class attrs, and `process_virtual_refs` itself (only for a fundamentally different batching policy).
```

docs/virtual_datasets.md
```diff
@@ -131,6 +131,8 @@ The `-validate` CronJob runs after each operational update. A dataset lists ever
...
+
+  `include_vars` / `exclude_vars` narrow a check to the source files carrying those variables, so variables publishing on different schedules take one instance each, held to `(1.0,)` from the position where their own files are expected. One instance covering all of them would need a leading tier loose enough for whichever group has not published, which also excuses a real gap in the group that has. A `0.0` tier excuses a position entirely.
```
aldenks:
```suggestion
  `include_vars` / `exclude_vars` narrow a check to the source files carrying those variables, so variables publishing on different schedules take one instance each, tuned tightly per schedule.
```

PR #901, opened 2026-08-15
Add the noaa-hrrr-analysis-virtual dataset

aldenks:
Rebased onto `main` now that #900 has merged; the diff is this dataset plus one validator fix.

A virtual (spatially-chunked, map-optimized) HRRR analysis: the materialized `noaa-hrrr-analysis` temporal structure — hourly `time` dim, each hour taking the shortest available lead — serving the full HRRR virtual variable catalog, including the `pressure_level` (39 levels) and `model_level` (50 levels) groups the materialized analysis does not carry.

## Sourcing: which lead each variable comes from

Verified against real GRIB indexes across every HRRR era (2014 v1 → 2026 v4), then by decoding messages where the index alone was ambiguous:

- **Nothing is silently NaN.** No variable that `has_hour_0_values()` reports as available at f00 is missing from the f00 index — the failure mode where a variable would read NaN across the entire archive does not occur.
- **Every variable it excludes was confirmed unusable at f00.** 24 of the 35 decode to a uniform field (accumulations and runoff at 0, categoricals at 0, percent-frozen at the −50 sentinel). The other 11 carry real spatially-varying data, but not the quantity the variable represents: the three 10 m wind maxima are literally the instantaneous analysis (`MAXUW`/`MAXVW` equal `UGRD`/`VGRD` at `anl` to within one 0.0625 packing quantum), and the reflectivity/graupel/hail maxima are extrema over a zero-length window, roughly half the true hourly extremum (echo coverage 3.3% vs 7.2%, mean 0.49 vs 1.38 dBZ).

So `has_hour_0_values()` is exactly the right partition and needed no change: 141 variables come from f00 of the hour's own cycle, 35 from f01 of the prior hour's. The same `(file type, hour-0)` key the materialized HRRR jobs group by drives it.

This also means the published 18-hour and 48-hour virtual forecasts need **no backfill** — their hour-0 NaNs on those 35 variables are correct, and filling them from f00 would store a different quantity under a `step_type=max` variable.

## Variables

Every catalog variable is carried. At the analysis's 1 hour lead a run-total accumulation spans exactly one hour, which a decode confirms is **bit-identical** to the hourly variant, so the five run-total variables keep their catalog names (matching the forecast datasets) and record the equivalence in their `comment`.

## Archive start

2014-10-01, matching the materialized analysis. Reaching back past HRRR v3 required:

- The v1-only `PRMSL` spelling of mean sea level pressure as an element alternative. Measured the transition precisely: `PRMSL` through the 2016-08-23 cycles, `MSLMA` from 2016-08-24 (HRRR v2), never both in one file. This also corrects the materialized config's existing note, which put the boundary at v3.
- Confirming v1-era GRIB decodes through `GribberishCodec` — a test pins the archive's first hour end to end.

Early-era coverage is partial, as expected for a virtual dataset (missing files stay NaN): 98/176 variables in 2014–15, 137 in 2017, 176 from 2024. There is also sporadic single-day patchiness where `wrfprs` is present but `wrfsfc`/`wrfnat` are absent — the same pattern the published forecast validation reports document for 2018-07-13 → 09-15. Both are `### Review notes` material for this dataset's own validation report.

## Operations

Hourly update at :50 (f00 publishes ~:51), 12-hour re-sweep window, validation at :20. Manifest splits scaled ~49× from the 48-hour product since the analysis carries one lead per position (single-level ≈ 0.47 MiB, pressure ≈ 2.7 MiB, model ≈ 3.1 MiB per full manifest).

Completeness partitions on `has_hour_0_values()` using the variable filters from #900:

- f01-sourced variables publish ~7 minutes *before* the hour they label → whole at every ingested position, `(1.0,)`.
- f00-sourced variables publish ~51 minutes *past* the hour → `(0.0, 1.0)`, excusing only the newest position.

Decode health runs at its defaults. The validation window ends at the update cron's previous fire time, so its newest position is one the update has already swept — both files are in hand and all 176 variables are decode-checked.

Cron jobs ship suspended until the initial backfill completes.

## Semantic missing values vs. the all-NaN decode failure

The one non-dataset commit here, from the `noaa-hrrr-analysis` NaN-fraction alert investigated in #905. `percent_frozen_precipitation_surface` read 0.999990 NaN — the source's −50 no-precipitation marker over a quadrant where nothing was precipitating, not a data problem. Measured directly against the source, across 402 consecutive August hours the whole-CONUS marker fraction peaks at 0.9962 for percent frozen precipitation and 0.985 for echo top: these variables sit within a fraction of a percent of being *entirely* marker, so an all-marker hour is a legitimate reading.

`CheckVirtualDecodeHealth` fails a variable whose sampled chunks all decode entirely NaN, which for those is a false alarm waiting to happen. It now holds variables with semantic missing values to decode errors only; whether their references exist is already `CheckVirtualManifestCompleteness`'s question, and exhaustive there rather than sampled. The predicate is the one `src/scripts/validation/availability.py` already used for the same distinction, moved into `common/validation.py` so there is one definition rather than two.

This narrows to the 3 of 176 variables declaring a non-NaN `fill_value` — `echo_top`, `percent_frozen_precipitation_surface`, `pressure_level/dew_point_temperature` — the same three in this dataset and in the published 18-hour and 48-hour virtual forecasts, which the fix covers too. AIFS has none.

## Testing

`ruff format`, `ruff check`, `ty check` clean on the rebased branch. Full suite passes: 2102 tests, 10 skipped. The slow tests backfill real S3 data and pin decoded values across root, pressure-level, and model-level groups for both the modern era and the archive's first hour, then run an operational update and re-check.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUwdEjgXNFYHvCbdkUqRgg)_


src/reformatters/noaa/hrrr/analysis_virtual/region_job.py
```diff
@@ -0,0 +1,59 @@
...
+from reformatters.common.types import Dim, Timedelta
+from reformatters.noaa.hrrr.hrrr_config_models import NoaaHrrrDataVar
+from reformatters.noaa.hrrr.virtual_region_job import (
+    NoaaHrrrVirtualRegionJob,
+    NoaaHrrrVirtualSourceFileCoord,
+)
+
+# A variable with hour-0 values reads f00 of its own hour's cycle; the rest read f01
+# of the prior hour's cycle (see the template config docstring for why).
+_LEAD_TIMES = {True: pd.Timedelta("0h"), False: pd.Timedelta("1h")}
```
aldenks: Remove constant and comment. Inline code into generate source file coords function and it will be self explanatory: lead time = 0h if has hour 0 else 1h

src/reformatters/noaa/hrrr/analysis_virtual/region_job.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class NoaaHrrrAnalysisVirtualSourceFileCoord(NoaaHrrrVirtualSourceFileCoord):
+    def out_loc(self) -> Mapping[Dim, CoordinateValue]:
+        return {
+            "time": self.init_time + self.lead_time,
+            **self._vertical_probe_loc(),
```
aldenks: This is either funny naming (what does a "probe "belong here for? We aren't doing any manifest, checking or validation, this is a general source file coordinate method), or a bug. Closely check in this poor request and the base one about whether we are properly writing the right fields into the right vertical levels. I believe that file_refs takes care of that for us, but check that and make sure that we have proper unit tests. 

The right answer of what to do here may not be clear, talk to me before making changes if so

src/reformatters/noaa/hrrr/analysis_virtual/template_config.py
```diff
@@ -0,0 +1,134 @@
...
+from reformatters.common.zarr import BLOSC_8BYTE_ZSTD_LEVEL3_SHUFFLE
+from reformatters.noaa.hrrr.hrrr_config_models import NoaaHrrrDataVar
+from reformatters.noaa.hrrr.virtual_template_config import (
+    NoaaHrrrVirtualTemplateConfig,
+)
+
+_RUN_TOTAL_COMMENT = (
+    "Despite the run-total name, this analysis reads the 1 hour forecast lead, where "
+    "the accumulation window is exactly one hour, so these are hourly accumulations - "
+    "identical to the equivalent hourly accumulation variable where one exists."
```
aldenks: Much too verbose. Do something like: "Identical to the one hour accumulated <hourly var name> in this analysis dataset."

src/reformatters/noaa/hrrr/virtual_template_config.py
```diff
@@ -655,6 +655,9 @@ def _root_data_vars(chunks: tuple[int, ...]) -> list[NoaaHrrrDataVar]:
         root_var(
             "pressure_reduced_to_mean_sea_level",
             element="MSLMA",
+            # PRMSL through the 2016-08-23 cycles, MSLMA from 2016-08-24 (HRRR v2) on;
+            # a file carries only one spelling.
+            element_alternatives=("PRMSL",),
             level="mean sea level",
             short_name="prmsl",
```
aldenks, line 660: Remove just this line

src/reformatters/noaa/hrrr/analysis_virtual/region_job.py
aldenks: Checked both PRs. **Not a bug — writes are correct — but the name is wrong, and I found a related gap I'd like your call on.**

**Why writes are safe.** `file_refs` builds every ref as `out_loc={**out_loc_base, **level_label}`, and `_message_lookup` gives each vertical-group variable a `level_label` parsed from its own `.idx` line (one entry per level in the template's index). So the representative level in `out_loc()` is always overridden per ref — it never reaches a written chunk for a group variable. It only survives for root variables, which have no vertical dim, where `_resolve_chunk_keys` ignores it (it iterates the variable's own dims). So yes, `file_refs` takes care of it, as you thought.

**Tests.** `test_file_refs_pressure_group_routes_each_level` and `test_file_refs_model_group_routes_hybrid_levels` pin distinct byte ranges per level; `tests/common/virtual_multi_group_test.py` checks per-level values end to end. Those cover the shared mechanism this dataset uses. I've added a physical check to the analysis e2e — temperature must fall from 1000 → 700 → 500 hPa — which catches one level's message being repeated across levels. Passing.

**The name.** It's named for the base class's contract (`out_loc` doubles as the manifest probe key for group-only files, per docs/virtual_datasets.md), which is a poor fit for a general coord method. Suggest renaming to `_representative_level`.

**The gap.** `CheckVirtualDecodeHealth._decode_coord` does `da.sel(loc)` with `loc` including the representative level, *then* calls `_sample_levels`. By then the vertical dim is gone, so `_sample_levels` returns the array unchanged — meaning **group variables are only ever decode-checked at one level** (1000 hPa / model level 1), never the `sampled_levels=3` the docstring advertises. This is pre-existing on `main` and affects the published forecast-virtual datasets too; their reports say "3 vertical levels per grouped variable", which looks overstated.

Two changes, both belonging in #900 rather than here:

1. Rename `_vertical_probe_loc` → `_representative_level`.
2. Have `_decode_coord` drop the representative vertical level from its selection so `sampled_levels` actually samples.

(2) changes shared-validator behavior, so I haven't touched it. Want both, just (1), or neither?

---

src/reformatters/noaa/hrrr/virtual_template_config.py
aldenks: The anchor lands on `short_name="prmsl"`, so I took this as the comment above and removed its second line, leaving:

```python
# PRMSL through the 2016-08-23 cycles, MSLMA from 2016-08-24 (HRRR v2) on.
element_alternatives=("PRMSL",),
```

That also matches the one-line version in the materialized `noaa/hrrr/template_config.py`. If you meant the whole comment, say so and I'll drop the remaining line too.

---

src/reformatters/noaa/hrrr/analysis_virtual/region_job.py
aldenks: Resolved per discussion: `out_loc` is now purely the slab the file fills, and the co-ingested-probe concept lives on `VirtualRegionJob.representative_probe_loc`, beside `representative_var`. The default derives the cell generically (out_loc + first label along each of the representative var's multi-chunk dims the file spans wholly), which is exactly the level HRRR and AIFS were hardcoding — so probe chunk keys are unchanged for existing stores, and both datasets' special cases are deleted. Also fixes decode health's level sampling as a side effect. In #900 (2287728); this coord simplifies to `{"time": init_time + lead_time}`.

---

aldenks:
Pushed 8ca606c here, from the `noaa-hrrr-analysis` NaN-fraction alert investigated in #905.

That alert was `percent_frozen_precipitation_surface` reading 0.999990 NaN — the source's −50 no-precipitation marker over a quadrant where nothing was precipitating, not a data problem. Measuring the source directly: across 402 consecutive August hours the whole-CONUS marker fraction peaks at 0.9962 for percent frozen precipitation and 0.985 for echo top, i.e. these variables sit within a fraction of a percent of being *entirely* marker. `CheckVirtualDecodeHealth` fails a variable whose sampled chunks all decode entirely NaN, which for those two is a legitimate hour, so the commit holds variables with semantic missing values to decode errors only. Reference existence for them is already `CheckVirtualManifestCompleteness`'s job, and exhaustive there rather than sampled.

In this dataset that covers `echo_top`, `percent_frozen_precipitation_surface`, and `pressure_level/dew_point_temperature` — the three of 176 variables declaring a non-NaN `fill_value`. The predicate is the one `src/scripts/validation/availability.py` already used for the same distinction, moved into `common/validation.py` so there is one definition rather than two.

---

aldenks:
rebase and update templates now https://github.com/dynamical-org/reformatters/pull/910 merged to get improved var comments

aldenks:
Rebased onto `main` (`c29b204`, #910) and regenerated the template — `7225ff4`.

Template diff is comments only, on the seven variables #910 touched that this catalog carries: `echo_top`, `percent_frozen_precipitation_surface`, `geopotential_height_cloud_{base,ceiling,top}`, `pressure_cloud_{base,top}`. No structural, encoding, or fill-value drift.

Checked one interaction: `_with_run_total_comment` replaces the catalog `comment` on the five run-total variables, so a new upstream comment on any of them would have been silently dropped. #910 added none there — the only paths where this dataset's attrs differ from the forecast catalog are exactly those five.

`ruff format`, `ruff check`, `ty check` clean; full suite 2175 passed, 10 skipped.

---

src/reformatters/noaa/hrrr/virtual_template_config.py
aldenks: Confirmed offline: the remaining one-line version stays as is.

---

src/reformatters/noaa/hrrr/analysis_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,104 @@
...
+
+    icechunk_virtual_config: IcechunkVirtualConfig = Field(
+        default_factory=lambda: IcechunkVirtualConfig(
+            containers=hrrr_virtual_chunk_containers(),
+            # The analysis carries one lead per time position where the 48-hour
+            # product carries 49 per init, so its splits scale up ~49x to keep full
+            # manifest byte sizes comparable: at ~16.4 bytes/ref, single-level
+            # 30000 x 1 ref/time ~= 0.47 MiB, pressure 4500 x 39 ~= 2.7 MiB, model
+            # 4000 x 50 ~= 3.1 MiB; see "Manifest splitting" in
+            # docs/virtual_datasets.md for the cost model.
```
aldenks: remove the vs 48h references and the doc section ref, just state the details

src/reformatters/noaa/hrrr/analysis_virtual/region_job.py
```diff
@@ -0,0 +1,52 @@
...
+    def out_loc(self) -> Mapping[Dim, CoordinateValue]:
+        return {"time": self.init_time + self.lead_time}
+
+
+class NoaaHrrrAnalysisVirtualRegionJob(
+    NoaaHrrrVirtualRegionJob[NoaaHrrrAnalysisVirtualSourceFileCoord]
+):
+    operational_update_window: ClassVar[Timedelta] = pd.Timedelta("12h")
+
+    def generate_source_file_coords(
```
aldenks, line 27: im not sure if this is the right place to do this, but implement this change: instead of leaving empty values at the latest time for all vars that have hour 0 (because those that don't can take the f01 from the prev init which is already out), limit the store to have the maximum `time` be only extended out to the max `time` of a var that has hour 0 values. then the end won't be ragged.

this also implies tighter + simpler manifest completeness validators.

src/reformatters/noaa/hrrr/analysis_virtual/template_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class NoaaHrrrAnalysisVirtualTemplateConfig(NoaaHrrrVirtualTemplateConfig):
+    """Virtual HRRR analysis: each hour's shortest-lead forecast data.
+
+    Hour-0-valued variables read f00 of the same hour's cycle. The rest read f01 of
+    the prior hour's cycle: their f00 entries exist in the GRIB index but carry no
+    usable data - a windowed or accumulated variable's f00 window is zero-length,
+    and the categorical and percent-frozen fields are uniformly zero or the -50
+    sentinel at analysis time. Matches the materialized noaa-hrrr-analysis.
```
aldenks: 1. remove all "matches this other reference dataset" per AGENTS.md
2. move the info about f00 vs f01 into generate_source_file_coords's docstring and make it much shorter: Uses the shortest present lead time for each variable: f00 if available, f01 otherwise.

PR #902, opened 2026-08-15
Redesign operational validation around configured Validator instances

aldenks:
One Validator base class (a frozen pydantic model with a single
check(ValidationContext) method) replaces the XarrayDataValidator |
VirtualDataValidator union, partial() configuration, and per-kind
runtime dispatch. The context carries the opened flattened dataset,
store, append dim + frequency, template data vars, and (for virtual
datasets) the operational-window region job, so checks are
dimension-agnostic and the forecast/analysis function pairs collapse
into one class each:

- CheckCurrentData(max_delay) replaces check_{forecast,analysis}_current_data.
  Each append-dim position is due max_delay after its own timestamp, evaluated
  on the append-dim grid anchored to the dataset's own positions — so a tight
  deadline holds at any wall-clock moment and an off-schedule run never
  false-alarms on the normal wait for the next cycle. Every dataset's max_delay
  is derived from its real update/validation schedule and documented inline.
  Sensitivity change: forecast datasets that previously tolerated whole missed
  cycles under max_age = 1 day now alert at the first validation after a
  missed update.
- CheckRecentNans replaces check_{forecast,analysis}_recent_nans. Newest-first
  max_nan_fraction tiers over a window replace init_time_offset/time_offset +
  num_recent_* (a leading 1.0 tier skips a still-filling newest position),
  matching CheckVirtualManifestCompleteness.min_present_fraction's shape.
  Hour-0 slices are skipped from the template's has_hour_0_values() instead
  of a caller-maintained additional_skip_lead_time_0_vars list. A new
  sampled_positions field checks N randomly chosen positions within the
  window, for updates that rewrite a deep window each run — replacing SWANN's
  custom check_random_time_within_last_year_nans.
- CheckExpectedShards replaces check_for_expected_shards with no exemptions:
  the write path always passes write_empty_chunks=True, so a missing shard
  always means a failed write. Verified against the production HRRR store —
  every categorical variable has full shard parity with temperature_2m
  (11,821 objects each) — so dropping the old hardcoded skip is safe.
- CheckReplicaMatchesPrimary replaces compare_replica_and_primary and gets the
  primary dataset from the context instead of a pre-bound partial (which
  compared the replica's dataset against itself).
- CheckVirtualDecodeHealth.positions is an int (newest N present positions)
  or "all", replacing the "latest" + max_positions overload; #901 can use
  positions=3.
- CheckVirtualManifestCompleteness keeps #900's include_vars/exclude_vars and
  source-file-boundary assertion, restructured onto the shared
  VariableSelection mixin.
- HRRR's check_forecast_completeness is deleted: every init in the 6-hourly
  48-hour dataset is a 48-hour cycle, so a truncated forecast reads as
  excessive NaN and the strict CheckRecentNans(window=4) subsumes it.

Misconfiguration now fails early and loudly: unknown or mistyped fields are
rejected at construction (extra="forbid" + validated types), include/exclude
variable names are validated against the template catalog (a typo raises;
a selected var missing from a partially backfilled store fails rather than
vacuously passing), and each instance carries a name including its variable
selection so multiple instances of one check are distinguishable in logs,
failure messages, and Sentry fingerprints.

Rebased on #900 (whose manifest-completeness variable filters this absorbs,
with its new tests ported to the new API), #905 (whose adjusted HRRR analysis
thresholds carry through), and #890 (whose vertical-group support carries
through: CheckExpectedShards counts shards from each variable's own shape and
encoding, CheckReplicaMatchesPrimary slices each variable's own dims, and
every replica — virtual included — gets the primary compare).

Testing: ruff format, ruff check, ty check clean; 1895 fast tests pass, plus
the slow e2e suites for SWANN, HRRR 48-hour materialized, HRRR 18-hour
virtual, and the common virtual/multi-group files. Net −164 lines.

https://claude.ai/code/session_01Wnge2sYfWjxDQK3qZ2wi9R


src/reformatters/noaa/mrms/conus_analysis_hourly/dynamical_dataset.py
```diff
@@ -50,62 +49,51 @@ def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
...
         return (
-            partial(
-                validation.check_analysis_current_data,
-                max_expected_delay=max_expected_delay,
-            ),
-            partial(
-                validation.check_analysis_recent_nans,
-                time_offset=-2,
-                max_nan_fraction=0.25,
+            validation.CheckCurrentData(max_delay=timedelta(hours=3, minutes=30)),
```
aldenks: I think we run this hourly now and there's almost zero lag, confirm. So max_delay should be just the validate aka 13mins

aldenks: Confirmed and applied in 6726e4e. The update's `append_dim_end = pd.Timestamp.now()`, so the :03 run writes the current hour's position, and I verified in the production store: at 21:14 UTC the newest `time` was 21:00. Set `max_delay=timedelta(minutes=13)` to match the :13 validation fire.

---

src/reformatters/common/validation.py
```diff
@@ -67,39 +68,109 @@ class OperationalValidationError(ValueError):
...
+        )
+        return self.region_job
+
+    def known_var_paths(self) -> Sequence[str]:
+        """The variable paths a check's include/exclude filters are validated against:
+        the template's when available (the complete catalog, even when a partially
+        backfilled store carries a subset), else the opened store's."""
+        if self.data_vars:
+            return [var.path for var in self.data_vars]
+        return [str(name) for name in self.ds.data_vars]
```
aldenks, line 103: double checking: are we every accidentally skipping vars in nested zarr groups because of this case?

aldenks: No — nested groups are covered on both branches. `ds` is always the product of `open_flattened_dataset`, whose `iterating.flatten_groups` recurses over `tree.subtree`, so vars land keyed by full path at any depth (`a/b/temperature`); the `data_vars` branch uses the template's `var.path`, the same convention. Verified empirically with a depth-2 nested store: the fallback returned all three paths, and NaN-ing only `a/b/temperature` made `CheckRecentNans` fail and name it — so nested vars are read, not skipped.

---

PR #904, opened 2026-08-17
Add the ECMWF IFS ENS 46-day sub-seasonal-range dataset

mrshll:
Adds `ecmwf-ifs-ens-forecast-46-day-1-5-degree`: the daily ECMWF IFS ENS sub-seasonal-range forecast with 35 variables on a 1.5° global grid, 101 members, 47 daily lead times, and pressure-level variables in their own group.

## Operational shape

The source archive is an independent operational unit registered as `ecmwf-ifs-ens-forecast-46-day-gribs`. Its unsuspended 06 UTC cron retrieves the shared ECDS variable set into Source Coop using both the ECDS and Source Coop secrets. The dataset reads that archive and owns only its 09 UTC update and 12 UTC validation crons, which remain suspended until the initial backfill is complete.

The archive variable list is defined beside the retrieval logic, and a test pins the dataset source variables to it so readers and the archive cannot drift.

## Data handling

- Icechunk is the primary store with no replicas.
- Root chunks are 25×24 spatially and shards are 125×240; pressure level precedes the spatial dimensions.
- Raw GRIB units are asserted from `GRIB_COMMENT` metadata before configured scaling and offsets are applied.
- Accumulated precipitation, snowfall, runoff, and radiation fields are converted to rates. Signed net-longwave, heat-flux, and stress fields select signed deaccumulation explicitly.
- Physical-range clipping was removed to avoid materialized/virtual divergence.

## Validators

- `CheckCurrentData(max_delay=4 days)` accounts for the roughly 53-hour ECDS publication delay.
- `CheckRecentNans(window=3)` excludes mask-carrying variables from the strict pass.
- A second quarter-sampled pass checks those masked variables at `max_nan_fraction=0.9999`, so an entirely NaN field still fails.

## Deployment checklist

- [ ] Merge and deploy the standalone archive cron.
- [ ] Confirm the archive advances through the current initialization frontier.
- [ ] Create the dataset store with `backfill-kubernetes` from the available archive start.
- [ ] Remove `reformat_suspend`, deploy update and validation, and confirm both fire successfully.
- [ ] Run the historical overwrite-chunks backfill to `append_dim_start` as the archive fills backward.

### 2026-08-21 — PR feedback

- [x] Merged current `main` and adopted the current validator API.
- [x] Renamed shared modules and classes around the 46-day product.
- [x] Applied the reviewed Icechunk storage, chunk layout, dimension order, and worker parallelism.
- [x] Added masked-variable validation, explicit signed deaccumulation, and edge-case coverage.
- [x] Moved archive ownership beside the archive logic and registered it independently.
- [x] Regenerated metadata and workflows; passed ruff, ty, the full local suite, and both CI architectures.

### 2026-08-21 — Retro

Keeping the archive as a store-less operational unit made its ownership and deployment independent of any one derived dataset. The archive/dataset variable-set contract test preserves that separation without allowing the two sides to drift.


src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
```diff
@@ -0,0 +1,127 @@
...
+# about 51.6 hours after its 00 UTC reference time (measured 51.4-52.1 h daily,
+# 2026-06-26 to 2026-08-11). With the archive cron at 06 UTC, the initialization this
+# selects is one published a couple of hours earlier.
+STAGING_DELAY = pd.Timedelta("53h")
+
+SOURCE_COOP_SECRET_NAME: Final = "source-coop-storage-options-key"  # noqa: S105
+ECDS_API_KEY_SECRET_NAME: Final = "ecmwf-ecds-api-key"  # noqa: S105
+
+
+class EcmwfS2sDynamicalDataset(
```
aldenks: naming of this as S2s instead of EcmwfIfsEns46Day15Degree is surprising/confusing. yes that's pretty long, but consistency with our pattern is useful. If that feels long, im cool going EcmwfIfsEns46day and dropping the resolution part.

For the common one (assumed to be setting up as shared for virtual too?), lets follow our pattern and call it EcmwfIfsEns46DayCommon...

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+            name=f"{self.dataset_id}-archive-grib-files",
+            schedule="0 6 * * *",
+            pod_active_deadline=timedelta(hours=6),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="1.5",
+            memory="8G",
+            ephemeral_storage="60G",
+            secret_names=[SOURCE_COOP_SECRET_NAME, ECDS_API_KEY_SECRET_NAME],
+        )
```
aldenks: if the super class is defining the grib archving command/logic, lets move this into that so the job is with the logic. It will need to be registered in `__main__.py` but that's fine

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+            pod_active_deadline=timedelta(hours=3),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="7",
+            memory="30G",
+            shared_memory="12G",
+            ephemeral_storage="30G",
+            secret_names=self.store_factory.k8s_secret_names(),
+            workers_total=workers,
+            parallelism=min(workers, 10),
```
aldenks: we don't need to limit parallelism when reading from source coop, can remove `min`

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+            partial(
+                validation.check_forecast_current_data,
+                max_latest_init_time_age=timedelta(days=4),
+            ),
+            partial(
+                validation.check_forecast_recent_nans,
+                num_recent_init_times=3,
+                exclude_vars=masked_vars,
+            ),
+        )
```
aldenks, line 79: rebase on main and update to use the new Validators. they also support include and exclude vars args

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def validators(self) -> Sequence[validation.DataValidator]:
+        """Return a sequence of DataValidators to run on this dataset."""
+        # NaN by construction, at a fraction that varies with where the sampled points
+        # land. Whole-grid sampling would make it stable but reads every lead time of
+        # all 101 members, so they are gated on completeness by
+        # check_for_expected_shards instead.
+        masked_vars = (
+            *source_fill_value_var_names(self.template_config.data_vars),
+            "pressure_level/specific_humidity",
```
aldenks, line 65: why is pressure_level/specific_humidity excluded from checks? lets add a second recent nans validator that includes all the masked_vars and is tuned with a much looser nan check but still confirms there's at least one non nan value.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
```diff
@@ -0,0 +1,877 @@
...
+            "latitude",
+            "longitude",
+        ),
+        "pressure_level": (
+            "init_time",
+            "lead_time",
+            "ensemble_member",
+            "latitude",
+            "longitude",
+            "pressure_level",
```
aldenks:
```suggestion
            "pressure_level",
            "latitude",
            "longitude",
```
I'd expect this array order gets us somewhat better compression becuase nearby pixels on the same level will be closer value than pixels on a different level

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
```diff
@@ -0,0 +1,877 @@
...
+        ),
+    }
+
+    @property
+    def lead_time_frequency(self) -> pd.Timedelta:
+        return pd.Timedelta("24h")
```
aldenks: i think fine to just make this an attr

aldenks (commented):
I need to pick up at the s2s_* files, but some comments so far

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
mrshll: s2s is ecmwf's official name for it, fwiw

mrshll: i prefer the long name

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
```diff
@@ -0,0 +1,877 @@
...
+            "longitude": 24,  # 10 chunks over 240 pixels
+        }
+        # Roughly ~551MB uncompressed
+        root_shards: dict[Dim, int] = {
+            "init_time": root_chunks["init_time"],
+            "lead_time": root_chunks["lead_time"],
+            "ensemble_member": root_chunks["ensemble_member"],
+            "latitude": root_chunks["latitude"] * 11,
+            "longitude": root_chunks["longitude"] * 10,
+        }
```
aldenks: we can fix these being too small and make them spatially square like this:


```suggestion
        # Roughly ~10.0MB uncompressed
        root_chunks: dict[Dim, int] = {
            "init_time": 1,
            "lead_time": 47,  # All lead times
            "ensemble_member": 101,  # All ensemble members
            "latitude": 25,  # 5 chunks over 121 pixels
            "longitude": 24,  # 10 chunks over 240 pixels
        }
        # Roughly ~570MB uncompressed
        root_shards: dict[Dim, int] = {
            "init_time": root_chunks["init_time"],
            "lead_time": root_chunks["lead_time"],
            "ensemble_member": root_chunks["ensemble_member"],
            "latitude": root_chunks["latitude"] * 5,
            "longitude": root_chunks["longitude"] * 10,
        }
```

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
```diff
@@ -0,0 +1,127 @@
+"""Shared operational surface of the ECMWF IFS ENS extended range (S2S) datasets.
+
+Each dataset retrieves the ECDS variables it needs into dynamical's GRIB archive with its
+own `archive-grib-files` command, then reformats from that archive.
+"""
+
```
aldenks:
```suggestion
```

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
aldenks: Looks like it's actually sub-seasonal-range forecast now  https://www.ecmwf.int/en/forecasts/about-our-forecasts/sub-seasonal-range-forecasts

> Our sub-seasonal-range forecasts (previously known as extended-range and monthly forecasts) provide an overview of potential weather conditions up to 46 days ahead. 

S2S is sub-seasonal _to seasonal_, which they have a research projects around but that's a multi agency thing not a specific model https://www.ecmwf.int/en/research/projects/s2s and might confuse with their "seasonal" SEAS5 setup.

I think `EcmwfIfsEns46day`  (and no 15Degree) or `EcmwfIfsSubSeasonal` on the common classes/files that's currently `S2s`.

src/reformatters/__main__.py
```diff
@@ -205,6 +208,10 @@ class UpstreamGriddedZarrsDatasetStorageConfig(StorageConfig):
         primary_storage_config=SourceCoopZarrDatasetStorageConfig(),
         replica_storage_configs=[EcmwfIfsEnsIcechunkAwsOpenDataDatasetStorageConfig()],
     ),
+    EcmwfIfsEnsForecast46Day15DegreeDataset(
+        primary_storage_config=SourceCoopZarrDatasetStorageConfig(),
+        replica_storage_configs=[EcmwfIfsEnsIcechunkAwsOpenDataDatasetStorageConfig()],
```
aldenks: make EcmwfIfsEnsIcechunkAwsOpenDataDatasetStorageConfig the primary and have no replicas. (#915 made icechunks always officially the primary)

The archiver cron job requests `secret_names=[SOURCE_COOP_SECRET_NAME, ECDS_API_KEY_SECRET_NAME],` so it has auth to write gribs to source

src/reformatters/ecmwf/ifs_ens/s2s_template_config.py
```diff
@@ -0,0 +1,273 @@
...
+MAXIMUM_LEAD_TIME = pd.Timedelta("1104h")
+ENSEMBLE_SIZE = 101
+LATITUDES = np.arange(90, -90.75, -1.5)
+LONGITUDES = np.arange(-180, 180, 1.5)
+
+
+class EcmwfS2sCommonTemplateConfig(TemplateConfig[EcmwfS2sDataVar]):
+    append_dim: AppendDim = "init_time"
+    # The daily, 101 member era of the ECMWF S2S archive. Earlier initializations run
+    # twice a week with 51 members.
```
aldenks:
```suggestion
    # The daily, 101 member era of the ECMWF sub-seasonal-range forecast archive.
```

src/reformatters/ecmwf/ifs_ens/s2s_config_models.py
```diff
@@ -0,0 +1,33 @@
...
+    # GDAL's unit normalization, so this is the raw GRIB unit.
+    grib_unit: str
+
+    # Applied to the raw GRIB values in this order: value * scale_factor + add_offset.
+    scale_factor: float | None = None
+    add_offset: float | None = None
+
+    # Clamp packing artifacts that push values outside the variable's physical range.
+    minimum_value: float | None = None
+    maximum_value: float | None = None
```
aldenks: I would remove probably these and their functionality unless they are masking some really funky values. From the comments ("a little past physical range") this sounds like it's minor and already part of the grib packing lossyness accepted by ECMWF.

Keeping would introduce a new source of divergence between materialized and virtual (we do already round in materialized, but this would add another kind of difference)

If this was added because of some really bad values, I would mask those extreme values specifically and then in the virtual we'd have review note in the validation report because we couldn't mask

src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
```diff
@@ -0,0 +1,378 @@
...
+        with (
+            Env(GRIB_NORMALIZE_UNITS="NO"),
+            rasterio.open(coord.downloaded_path) as reader,
+        ):
+            assert reader.count == len(coord.present_levels or (None,)), (
+                f"Expected {len(coord.present_levels or (None,))} messages, "
+                f"found {reader.count} in {coord.downloaded_path}"
+            )
+            _validate_grib_metadata(reader, data_var)
+            # Byte ranges are requested in output level order, so band order is too.
```
aldenks, line 196: can an assert over metadata read from `reader` confirm this?  I think `reader.tags(band)` has grib message metadata

src/reformatters/ecmwf/ifs_ens/s2s_region_job.py
```diff
@@ -0,0 +1,378 @@
...
+        )
+
+    def read_data(
+        self,
+        coord: EcmwfS2sSourceFileCoord,
+        data_var: EcmwfS2sDataVar,
+    ) -> ArrayFloat32:
+        # GDAL's unit normalization converts some temperature parameters to Celsius and
+        # leaves others in Kelvin; disabling it keeps every read in the raw GRIB units
+        # each variable's scale_factor and add_offset are written against.
```
aldenks:
```suggestion
        # Disable GDAL's unit normalization on read. Instead,
        # apply_data_transformations uses a variable's scale_factor and add_offset.
```

src/reformatters/ecmwf/ifs_ens/s2s_region_job.py
```diff
@@ -0,0 +1,378 @@
...
+        internal_attrs = data_var.internal_attrs
+        if internal_attrs.scale_factor is not None:
+            data_array *= internal_attrs.scale_factor
+        if internal_attrs.add_offset is not None:
+            data_array += internal_attrs.add_offset
+
+        if internal_attrs.deaccumulate_to_rate:
+            assert internal_attrs.window_reset_frequency is not None
+            threshold = internal_attrs.deaccumulation_invalid_below_threshold_rate
+            if threshold is None:
```
aldenks: will you (the AI) share a per var rationale for which are deaccumed and which are not? I'm not sure we should deaccum any of the more exotic vars beyond precipitation and radiations

If we end up wanting to deaccum signed series, use an explicit data var internal attrs attribute to specify which deaccum function to use

src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
```diff
@@ -0,0 +1,378 @@
...
+            template_ds=template_ds,
+            append_dim=append_dim,
+            all_data_vars=all_data_vars,
+            reformat_job_name=reformat_job_name,
+            filter_start=append_dim_start,
+        )
+        return jobs, template_ds
+
+
+def _deaccumulate_signed_inplace(data_array: xr.DataArray) -> None:
```
aldenks, line 272: definitely need more edge case unit tests for this function if we keep it

aldenks (commented):
I think all this and the previous comments are pretty straightforward, except for I'm not sure if we want to deaccumulate so many variables. We can't do that in virtual so we need to think about variable naming in both cases for anything we deaccum (e.g. run_total_xxx and xxx). The per var rationale will help decide.

aldenks:
@mrshll done reviewing! [part 1](https://github.com/dynamical-org/reformatters/pull/904#pullrequestreview-4962866587), [part 2](https://github.com/dynamical-org/reformatters/pull/904#pullrequestreview-4988752011)

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
mrshll: Renamed the shared modules and classes around `EcmwfIfsEns46Day…` and updated the product wording to sub-seasonal-range.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
mrshll: Moved the archive command and cron into the standalone `EcmwfIfsEns46DayGribArchiver` beside the archive logic and registered it in `__main__.py`.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
mrshll: Removed the cap; update parallelism now equals the variable-group worker count.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
mrshll: Merged latest main and updated this to configured `Validator` instances with include/exclude selections.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/dynamical_dataset.py
mrshll: Added a quarter-sampled `CheckRecentNans` over all masked variables at `max_nan_fraction=0.9999`, which still fails an all-NaN field.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
mrshll: Moved `pressure_level` before the spatial dimensions and regenerated the template.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
mrshll: Changed `lead_time_frequency` to a typed attribute.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
mrshll: Applied the 25×24 spatial chunks and 125×240 shards and regenerated the template.

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
mrshll: Removed the module docstring; the dataset-owned archiver module is now gone entirely.

src/reformatters/__main__.py
mrshll: Made Icechunk the primary with no replicas; the standalone archive cron retains both Source Coop and ECDS secrets.

src/reformatters/ecmwf/ifs_ens/s2s_template_config.py
mrshll: Updated the era comment and surrounding metadata to sub-seasonal-range terminology.

src/reformatters/ecmwf/ifs_ens/s2s_config_models.py
mrshll: Removed the min/max clipping fields, transformation code, and clipping tests.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
mrshll: Confirmed `_validate_grib_metadata` asserts that every band `GRIB_COMMENT` unit matches the configured raw `grib_unit`.

src/reformatters/ecmwf/ifs_ens/s2s_region_job.py
mrshll: Applied the suggested GDAL normalization comment.

src/reformatters/ecmwf/ifs_ens/s2s_region_job.py
mrshll: Checked archived 2026-08-10 messages: precipitation, snowfall, runoff, and radiation fields are running totals; net longwave, latent/sensible heat, and both stresses are signed running totals. Added explicit `deaccumulation_type="signed"` on those six signed fields.

src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
mrshll: Added signed-deaccumulation tests for irregular lead spacing with a non-leading time axis, missing-value propagation, and a single-step series.

src/reformatters/ecmwf/ifs_ens/s2s_dynamical_dataset.py
mrshll: Also renamed the standalone archiver module, class, tests, and CLI ID around `EcmwfIfsEns46Day`.

PR #906, opened 2026-08-17
Add the ECDS GRIB stager for ECMWF IFS Extended Range

mrshll:
Retrieves ECMWF's extended range (S2S) ensemble from ECDS into a dynamical controlled GRIB archive. This is the source side only — the dataset that reads the archive is #904, stacked on this branch.

## Why an archive at all

ECDS serves this data from a request queue with no addressable objects and no throughput guarantee. A reformatter reading it directly would put that queue into its write path, on every backfill and every operational update. Instead each initialization is retrieved once into `s3://us-west-2.opendata.source.coop/dynamical/ecmwf-ifs-grib/ecmwf-ifs-ens-forecast-46-day/<date>/`, a sibling prefix to the 15 day dataset's MARS archive, and the reformatter reads only from there.

## How a request is shaped

`request_shards.py` splits an initialization along the axes ECDS itself charges for — forecast type, level type, variable, level, lead time — into requests whose results stay within a worker's disk and within the undocumented window an ECDS result remains downloadable. For the 35 variables the daily dataset needs, that is 12 requests, 406,828 messages and 22.7 GB per initialization.

## Completeness and addressability

`grib_inventory.py` walks a retrieved blob, names each message from its GRIB field key, and checks the full variable x level x member x lead inventory before anything is published. A staged object's presence therefore means it is complete, so the reformatter needs no separate completeness check. Each blob is published beside a byte range index, which is what lets a later read fetch one message without scanning the blob.

## Durability

`ecds_client.py` persists each request's state. An ECDS result expires without a published SLA, so a second download of the same result may be impossible and an already downloaded blob is kept rather than re-fetched. Transient constraint and costing failures retry, the poll backoff keys to consecutive transport failures rather than the poll index, and a request whose status URL goes dead gives up in minutes rather than days.

## Shared rclone helper

`list_files` moves from the DWD archiver into `common/rclone.py`, with the ECCC archiver and this one joining it rather than a third copy being written. It now merges its environment overrides itself, so callers pass only what they add — which is why `copy_files_from_dwd` no longer builds a `full_env`.

## Notes for review

- The stager has no entry point of its own. `archive-grib-files` is a method on the dataset class, so it arrives in #904; this PR is library code and tests.
- `tests/ecmwf/fixtures/s2s_control_2026-08-10T00Z.grib2` is 11 real control messages of the 2026-08-10T00Z initialization, retrieved from ECDS. The inventory tests assert against all of them.


aldenks (approved):
Keep an eye on the file count in ECCC HRDPS and DWD ICON archives before and after this change ships but besides that approved!

src/reformatters/ecmwf/archive_gribs/stage_gribs.py
```diff
@@ -0,0 +1,217 @@
...
+
+log = get_logger(__name__)
+
+STAGING_WORK_DIR: Final[Path] = DOWNLOAD_DIR / "ecmwf-s2s-staging"
+# ECDS serves 8 simultaneous requests without error; 4 keeps most of the queue-time
+# saving while leaving room for other users of the account.
+DEFAULT_CONCURRENT_REQUESTS: Final[int] = 4
+
+
+def stage_initialization(
```
aldenks: the naming of this and the file, local vars, etc as `stage` was confusing to me -- it sounds like we are getting rid of it not too long after. `archive` or `store`?

mrshll: Agreed, and it is worse than just the connotation: `staging` is already taken in this repo for running a second version of a dataset concurrently (`common/staging.py`, `stage/<id>/v<version>` branches, `docs/staging.md`), so two unrelated meanings were sharing the word.

Renamed to `archive` in 0cd63253, which matches the `archive_gribs` package and the `archive-grib-files` command it serves. `stage_gribs.py` → `archive.py`, `stage_initialization` → `archive_initialization`, and the local names, test names and prose that followed. No `stag*` left in the ECMWF paths.

PR #909, opened 2026-08-18
Validate recent NaNs across trailing output shards

aldenks:
Closes a value-coverage hole where a failed source read can leave NaNs inside a shard that CheckExpectedShards still considers present.

- CheckRecentNans now defaults to the newest two output shards, covering updates that cross a shard boundary. The shards option supports deeper re-sweeps.
- Random spatial points are read across the full append-dimension span in one selection; quarter-grid and whole-grid checks retain explicit position windows.
- Structurally all-NaN sample points are excluded, and the check fails if nothing measurable remains.
- Adds missing NDVI raw/QA coverage.

Checks: ruff format, ruff check --fix, ty check, and 2,061 non-slow tests.


AGENTS.md
```diff
@@ -121,7 +121,9 @@ Base class: `src/reformatters/common/dynamical_dataset.py`, commented example su
...
 - Implement `operational_kubernetes_resources()` - define update/validate cron jobs
-- Implement `validators()` - return validation functions for the dataset
+- Implement `validators()` - return the operational validation checks for the dataset
+
+Operational validation conventions. `CheckCurrentData(max_delay=...)` sets when each append-dim position is due, counted from that position's own timestamp (typically the validation cron's offset after the cycle it validates), so a tight deadline holds even on an off-schedule run. `CheckRecentNans` defaults to covering the two newest output shards: the current trailing shard and the previous one in case an update crossed a shard boundary. Increase its `shards` value when an update deliberately re-sweeps further back. `CheckExpectedShards` proves a written shard exists; only the NaN check sees NaNs left *inside* a shard by a transiently failed source file. With the default `spatial_sampling="random_points"`, points are drawn once and read across the whole window in one pass. `"quarter"` and `"all"` read each position separately, so use `window` to set an explicit position count.
```
aldenks: Cut this line

src/reformatters/common/iterating.py
```diff
@@ -55,7 +55,10 @@ def dimension_slices(
 ) -> Sequence[slice]:
     """Returns the integer offset slices which correspond to each shard or chunk along a single `dim` of `ds`."""
     chunk_sizes_set = {
-        var.encoding[kind][var.dims.index(dim)] for var in ds.data_vars.values()
+        # An unsharded array (e.g. a virtual store) has no "shards" encoding; its
+        # shard is its chunk.
+        (var.encoding.get(kind) or var.encoding["chunks"])[var.dims.index(dim)]
+        for var in ds.data_vars.values()
```
aldenks: Undo this change. 
- is CheckRecentNans being called on virtual datasets? For virtual, manifest completeness check is what we want instead
- caller should pass the right kind

src/reformatters/common/validation.py
```diff
@@ -328,51 +328,52 @@ def check(self, context: ValidationContext) -> ValidationResult:
 
 
 class CheckRecentNans(VariableSelection, Validator):
```
aldenks, line 333: Give this whole implementation and the helpers in this file it calls a review for simplification and implement improvements you find.

Then enumerate real edge cases for test and cross check our unit test coverage, filling in gaps

src/reformatters/contrib/noaa/ndvi_cdr/analysis/dynamical_dataset.py
```diff
@@ -56,5 +56,15 @@ def validators(self) -> Sequence[validation.Validator]:
                 max_nan_fraction=0.97,
                 include_vars=["ndvi_usable"],
                 spatial_sampling="all",
+                window=2,
+            ),
+            validation.CheckRecentNans(
+                # ndvi_raw has no measured stable NaN baseline; this threshold still
+                # catches a position written entirely NaN. qa is integer typed, so it
+                # holds no NaN to measure and a NaN check cannot cover it.
```
aldenks:
```suggestion
                # catches a position written entirely NaN.
```

src/reformatters/common/validation.py
```diff
@@ -328,51 +328,52 @@ def check(self, context: ValidationContext) -> ValidationResult:
...
 
     A variable the store holds no lead_time=0 value for (see `stores_hour_0_values`)
     has that slice dropped before computing the NaN fraction.
     """
 
     max_nan_fraction: float | tuple[float, ...] = 0.0
-    window: int = 2
-    sampled_positions: int | None = None
+    window: int | None = None
+    shards: int = 2
```
aldenks:
```suggestion
    append_dim_window: int | None = None
    append_dim_shards: int | None = None
```


Also add a pydantic validator that at most one of the two is set

When resolving window, default to 2 shards if neither value is set, put 2 in a constant 

Don't edit the names unless their meaning really is append dim shards/window

src/reformatters/common/iterating.py
aldenks: Reverted, and `_resolve_window` now passes `"shards"` explicitly.

On the question: no, `CheckRecentNans` is never called on a virtual dataset. I enumerated every registered dataset's `validators()` — all five virtual datasets (`noaa-gefs-forecast-10-day-spatial-dev`, `noaa-hrrr-forecast-48-hour-virtual`, `noaa-hrrr-forecast-18-hour-virtual`, `ecmwf-aifs-single-forecast-virtual`, and the GEFS spatial dev store) configure zero `CheckRecentNans`; they use `CheckVirtualManifestCompleteness` as you say. Every materialized dataset that does configure one has a uniform, non-`None` append-dim shard size, so the strict `encoding["shards"]` lookup is always satisfiable.

The only place that was exercising the unsharded path was `test_nan_check_covers_group_vars` in `virtual_multi_group_test.py`, which is checking group-variable coverage rather than window derivation — it now sets an explicit `append_dim_window=2`.

---

src/reformatters/common/validation.py
aldenks: Done. Simplifications:

**Collapsed the duplicated read branches.** `check()` had a one-pass branch and a per-position branch making the same `_nan_fractions_by_position` call with identical kwargs, differing only in what they select and how they key results. That is now one loop over `_reads()`, which returns the append-dim selections plus the recency each starts at. This also fixed an inconsistency: the per-position branch pre-seeded `fractions` with every selected variable, so a variable nothing could be measured for printed `nan` in the log there but was omitted in the one-pass branch. Both now omit it and report it under `unmeasured`.

**Smaller things:** the selected-variable subset is a named local rather than an inline `cast` inside the `min()`; `max_workers` is hoisted back out of the loop; the worst-fraction log line is a dict comprehension instead of a 130-character nested f-string.

**One behavior change worth a look.** A selection containing a non-float variable now raises. An integer array holds no NaN, so the check passed vacuously — that is exactly what made the `qa` entry in the ndvi_cdr check a silent no-op. `qa` is the only integer data variable across all registered datasets and after your fix nothing selects it, so this changes no current configuration; it just makes the next one loud.

Edge cases enumerated, against existing coverage:

| Case | Status |
|---|---|
| Unknown field / bad strategy / tiers > window / threshold out of range / all-1.0 tiers / `append_dim_window=0` / `append_dim_shards=0` / `sampled_points=0` | already covered |
| Dead-point exclusion, all-dead failure, quarter / all / random_points, x-y dims, lead_time=0 skipping (×3), per-position independence, empty selection, unknown var, fewer positions than tiers | already covered |
| **Both bounds set** | added |
| **`max_nan_fraction=()`** | added |
| **Partially filled trailing shard** (the operational norm — window must span the whole previous shard, not just the positions in the partial one) | added |
| **More shards requested than the store holds** (young store) | added |
| **`append_dim_window` past the store's extent** | added |
| **Shard size read from the selected variables** when the store's variables disagree along the append dim — the "mixed store" half of 87cff35 | added |
| **Non-float selection** | added |
| **Logged worst fraction excludes an excused position** — point sampling reads it anyway in the single pass, so this is the regression 87cff35 fixed | added |

Full suite 2232 passed, 10 skipped; `ruff format`, `ruff check`, `ty check` clean.

---

PR #920, opened 2026-08-19
Skip GEFS source files the cycle has not published yet

### LINKED ISSUE #893: noaa-gefs-forecast-35-day update cron intermittently exceeds its 30-min pod_active_deadline
## Context

Sentry triage turned up three correlated open issues, all tracing back to `noaa-gefs-forecast-35-day`:

- [REFORMATTERS-EQ](https://dynamical.sentry.io/issues/REFORMATTERS-EQ) — Cron failure: `noaa-gefs-forecast-35-day-update`, "timeout check-in detected." First seen 2026-05-31, still recurring (last seen today).
- [REFORMATTERS-ET](https://dynamical.sentry.io/issues/REFORMATTERS-ET) — Cron failure: `noaa-gefs-forecast-35-day-validate`, "error/missed check-in detected." First seen 2026-06-07, still recurring (last seen today). Already assigned to the team.
- [REFORMATTERS-HH](https://dynamical.sentry.io/issues/REFORMATTERS-HH) — `OperationalValidationError: Zarr validation failed: - No data found within 1 day, 0:00:00 of now`, raised by the validate cron itself. First/last seen today.

## What we found

PR #681 ("Tune operational update/validate schedules and pod deadlines," merged 2026-06-26) tightened `noaa-gefs-forecast-35-day-update`'s `pod_active_deadline` from 3.5 hours down to 30 minutes, based on an observed "runs take <23 min," and moved the validate cron to fire right at that new deadline boundary (only ~7 min of slack past the quoted baseline).

Since then, two more PRs added work to this same dataset without revisiting the deadline:

- #749 "Add 80m wind, temperature, and pressure to NOAA GEFS" (merged 2026-07-15) — added `temperature_80m`, `pressure_80m`, `wind_u_80m`, `wind_v_80m`.
- #836 "Add wind gust to NOAA GEFS forecast and analysis" (merged 2026-07-31) — added `wind_gust_surface`.

Five new variables added to a 31-ensemble-member × ~181-lead-time materialized dataset after the deadline was calibrated is a plausible way to have pushed real runtimes from "<23 min" toward/past the 30-min cutoff on some days, producing exactly the observed chain: update times out (EQ) → no new `init_time` lands → the next day's validate fails with "No data found within 1 day" (HH) → the validate cron itself errors/misses its check-in (ET).

Two general fixes landed very recently that target part of this failure family but are unlikely to be the whole story:

- #883 "Bound icechunk object store requests with explicit timeouts" (merged 2026-08-12) — fixes the "operational update sat N minutes inside a stalled commit and got killed by the pod deadline without a check-in" pattern for any icechunk-backed dataset.
- #888 "Keep the Sentry connection alive across quiet stretches" (merged 2026-08-13) — written for virtual datasets' polling loop, doesn't directly touch this materialized dataset's code path.

Both merged only 1-2 days ago and the cron fires once/day, so there hasn't been enough post-fix signal yet — but REFORMATTERS-HH/ET both have occurrences from today (2026-08-14), after both merges, so the timeout is still happening.

## Recommendation

Re-measure `noaa-gefs-forecast-35-day-update`'s actual current runtime (job history/logs) now that it carries the variables added in #749/#836, then either:

- loosen `pod_active_deadline` (e.g. 40–45 min) with matching headroom before the validate cron's schedule, or
- reduce per-run work (e.g. `max_vars_per_job`/worker count tuning) to bring runtime back under the existing deadline.

This needs an operational decision (current real runtime + acceptable risk tradeoff) rather than a blind code change, so filing here instead of opening a PR.

## Sentry-side handled in this triage pass

Left REFORMATTERS-EQ, REFORMATTERS-ET, and REFORMATTERS-HH unresolved (already assigned/tracked, and this is a real ongoing issue, not noise) with a comment linking here.

---
_Filed by Claude Code during scheduled Sentry error triage._

aldenks:
Fixes REFORMATTERS-EQ
Fixes REFORMATTERS-ET
Fixes REFORMATTERS-HR

## What

`noaa-gefs-forecast-35-day-update` went from ~7 min to ~28 min on 2026-08-07, the first run after #862 deployed. It stayed there every run since, and on **2026-08-20 it crossed the 30 min `pod_active_deadline` and broke** — see "This has now fired in prod" below.

The 35-day update generates coords for all 181 lead times of the newest init, but at cron time (init+6h33m) the cycle has only produced through f384. The remaining 76 leads × 31 members × 12 newest-init jobs are **28,272 requests for files that exist nowhere**. They 404 on the S3 mirror and fall through to NOMADS, where after #862 each one takes a token from the 100/min rate limiter: 2,356 ÷ 100 = **23.6 min of sleep per pod**, which accounts for the entire regression. Those same requests, unthrottled, were the burst that tripped Akamai's bot mitigation — every one of 60 sampled events on the 302 issue is a lead between f468 and f558, all within 45 s of job start.

#862 was the right instinct. It throttled the symptom.

## How

Don't generate coords for the extension until the cycle is old enough that we'd expect to have it:

```python
if pd.Timedelta(lead_time) <= GEFS_PRE_EXTENSION_MAX
or now - pd.Timestamp(init_time) >= GEFS_EXTENSION_REQUEST_MIN_AGE
```

The age only ever has to separate the two init times an operational update touches: the newest, **6h33m** old at cron time, and the previous, **30h33m**. Any value between those splits them, so there is no cliff to engineer around. 28h sits near the top of that range so that a mid-cycle catch-up run skips the extension rather than writing a partial one — which is not hypothetical: a manual re-run at init+15h today left the 08-20 init with members 1–16 out to 840h and the rest at 384h.

Measured from `noaa-gefs-pds` `last_modified`, all 31 members across the a and b files:

| | |
|---|---|
| f≤384 set completes | init+6h31m .. init+6h48m |
| first f>384 file appears | init+6h34m .. init+6h36m |
| f>384 set completes (f840) | init+27h06m .. init+28h06m |

Note the third row: the extension can still be arriving at **init+28h06m**, so 28h is *not* a time by which it is complete. What has to be complete by request time is the previous init, and the update asks for that at init+30h33m — ~2.5h of margin. The constant is named `GEFS_EXTENSION_REQUEST_MIN_AGE` for what it decides rather than a completeness it does not guarantee, so that nobody "corrects" it to 31h and silently stops the extension being requested at all.

Lead times **≤ 384h are still all requested**, so the NOMADS fallback keeps doing the one job it exists for: covering S3 mirror lag, measured at a median of 77 s (p99 240 s) over 930 tail files.

The only behavior given up is opportunistically picking up the extension files that land while the update is running — about 10% of them, concentrated in 8 of 31 members. The next day's run reprocesses that init out to 840h regardless, so the newest init lands consistently at f384 for every member instead of ragged.

## Guarding the constants

`tests/noaa/gefs/source_publication_test.py` lists noaa-gefs-pds for a settled cycle and checks the schedule these constants describe. CI runs the full suite including slow tests, so it is exercised on every push. It reads `last_modified` rather than polling a live cycle, so it does not depend on when CI runs, and it derives the cron offset from `operational_kubernetes_resources` rather than hardcoding it:

| assertion | observed | bound |
|---|---|---|
| first wave complete | init+6h31m .. 6h42m | init+8h33m |
| extension complete before the next run asks | init+27h06m .. 28h06m | init+30h33m |
| extension published while the update runs | ~10% | < 25% |

The third is the canary for the one case where this PR would be leaving data behind: if NCEP gets much faster at publishing the extension, it fails and we should start requesting it. A companion test with no network pins the request age between the two init ages, so moving the cron without revisiting the constant fails loudly. The listing costs ~4.5 s.

## This has now fired in prod

The 2026-08-20 00z run is the failure this regression was heading for. From icechunk ancestry:

| Time (UTC) | Event |
|---|---|
| 06:33:46 | `Expand dataset` — job starts |
| 06:36–07:01 | 24 worker commits, the usual ~28 min pattern |
| **07:03:00** | Sentry monitor records a **timeout** (`max_runtime: 30`) → REFORMATTERS-EQ |
| 07:03:36 → 07:17:54 | validate cron runs against an unfinished update, fails 6× with `CheckCurrentData: Missing init_time=2026-08-20T00:00:00 (latest present is 2026-08-19T00:00:00)` → REFORMATTERS-HR, and the failed check-in → REFORMATTERS-ET |
| 07:19:06, 07:21:03, 07:29:52 | workers 16, 4, 23 commit late — worker 23 commits a *second* time, consistent with a killed pod being retried |
| **07:29:57** | `Update at 2026-08-20T07:29:57Z` — **56.2 min** after cron start |

The archive self-healed through pod retries, so there is **no data loss** — no manual backfill was needed. But publication landed 56 minutes after cron start instead of ~7, the update monitor is red, and validation failed for ~15 minutes on a dataset that was genuinely incomplete at the time it was checked.

All three Sentry issues are the same root cause this PR removes, which is why they are referenced above. Thanks @mrshll for the cross-reference.

## Expected effect

Run time back to ~7 min, 28,272 fewer NOMADS requests per day, ~25 vCPU-h/day of idle pod time recovered, and the 302s stay gone because the burst that caused them is gone. No change to the settled archive: every init still reaches 840h on the run after it.

`CheckRecentNans(max_nan_fraction=(0.45, 0.0))` is unchanged and still correct. The newest init now pins to exactly f384, leaving 76 of 181 lead times NaN: **0.4199**, or **0.4222** for a variable whose hour-0 slice is dropped before the fraction is computed. Its comment now derives from `GEFS_PRE_EXTENSION_MAX` instead of citing an observed 0.420789, which was both stale and below the accumulated-variable value.

## Scope

Only `forecast_35_day` filters. `gefs/analysis` needs nothing: it asks only for the shortest available lead, so it has no futile requests, and its cron runs at init+3h51m against files published ~init+3h47m — a 4-minute margin that leans on the NOMADS fallback. `forecast_10_day_spatial` is virtual and already discovers by listing.

## Not in this PR

1. **`pod_active_deadline` is 30 min**, and as of 2026-08-20 runs are crossing it. This PR removes the pressure by returning the job to ~7 min, so the deadline should then be resized against the real runtime rather than against the regression. If this PR sits unmerged, the 08-20 cascade will likely repeat daily, and bumping the deadline is the stopgap.
2. **The cron fires before the source finishes.** Over 28 days the last f≤384 file lands at a median of init+6h39m (max init+6h48m), so on 24 of 28 days at least one needed file arrives after the 06:33 start. That is what actually costs data — on 08-11 and 08-12 the source ran ~15 min late and the tail dropped to 378h. Moving the cron to ~06:45 would publish at ~06:53, still earlier than the 07:01 the store becomes readable today.

## Test plan

- [x] `uv run ruff format && uv run ruff check --fix && uv run ty check` clean, prek hooks pass
- [x] Two filter tests in `tests/noaa/gefs/forecast_35_day/region_job_test.py`: a cycle too young to hold its extension yields no lead time past 384h; one older than the request age yields every lead time
- [x] New `tests/noaa/gefs/source_publication_test.py` — the schedule assertions above, plus the no-network request-age check
- [x] `uv run pytest` (full suite, slow included, as CI runs it) — 2163 passed, 10 skipped

Findings behind this: https://claude.ai/code/artifact/0a6c5b76-774f-4ec3-b233-af3d44692c5c


mrshll:
Cross-referencing from a Sentry triage pass: this PR's root-cause analysis matches three currently-open Sentry issues exactly —

- [REFORMATTERS-EQ](https://dynamical.sentry.io/issues/REFORMATTERS-EQ) — `noaa-gefs-forecast-35-day-update` cron timeout (hitting ~29.5 min against the 30 min deadline)
- [REFORMATTERS-ET](https://dynamical.sentry.io/issues/REFORMATTERS-ET) — `noaa-gefs-forecast-35-day-validate` cron error check-ins, downstream of the update job not finishing
- [REFORMATTERS-HR](https://dynamical.sentry.io/issues/REFORMATTERS-HR) — `CheckCurrentData` validation failures for missing `init_time` on this dataset

Worth including `Fixes REFORMATTERS-EQ`, `Fixes REFORMATTERS-ET`, `Fixes REFORMATTERS-HR` in the merge commit message so Sentry auto-resolves them. Leaving these three unresolved/unignored in Sentry for now rather than duplicating this fix.

---

aldenks:
Added all three `Fixes` lines to the top of the PR description so they carry into the squash commit — worth a glance at merge time to confirm they survive into the final message.

I checked each against the store before adding them, and in doing so found that the failure this PR was heading for **fired this morning**. From icechunk ancestry on `noaa-gefs-forecast-35-day`:

| Time (UTC) | Event |
|---|---|
| 06:33:46 | `Expand dataset` — job starts |
| 06:36–07:01 | 24 worker commits, the usual ~28 min pattern |
| **07:03:00** | monitor timeout at `max_runtime: 30` → EQ |
| 07:03:36 → 07:17:54 | validate runs against an unfinished update, 6× `Missing init_time=2026-08-20T00:00:00` → HR, failed check-in → ET |
| 07:19:06, 07:21:03, 07:29:52 | workers 16, 4, 23 commit late — worker 23 commits a *second* time, consistent with a killed pod being retried |
| **07:29:57** | final commit — 56.2 min after cron start |

So the three issues aren't just correlated with this PR's analysis, they're one cascade: the 28-minute runtime crossed the 30-minute deadline, pods were killed and retried, and the validate cron at 07:03 checked a dataset that genuinely didn't have the 08-20 init yet.

Two practical notes:

- **No data loss and no backfill needed.** The retries completed the write; `init_time` currently ends at 2026-08-20 and the last two inits are intact. The cost was 56 minutes to publication instead of ~7, plus the red monitor.
- **This will likely repeat daily until this merges.** The margin was 30 seconds before today and it's now negative. If review is going to take a while, bumping `pod_active_deadline` is the stopgap — I've left it out of this PR since it's an operational call, and this PR makes it moot by returning the job to ~7 min.

---

src/reformatters/common/source_listing.py
```diff
@@ -1,9 +1,11 @@
...
-a frontier to probe) implement `discover_available` another way and never import this.
+"""Object-store listing of the source files a job could fetch.
+
+`discover_available_by_obstore_listing` is a `VirtualRegionJob.discover_available`
+implementation for sources listable with obstore (S3, GCS, Azure, local filesystem,
+...), generic over the backend — the caller passes the built store. Sources obstore
+can't list (an HTML directory index, a frontier to probe) implement
+`discover_available` another way and never import it. `listed_keys_by_prefix` is the
+underlying listing, for callers that ask a different question of the same keys.
 """
```
aldenks: Cut this docstring entirely

src/reformatters/common/source_listing.py
```diff
@@ -40,7 +42,7 @@ def discover_available_by_obstore_listing(
     ]
 
 
-def _list_objects(
+def listed_keys_by_prefix(
```
aldenks:
```suggestion
def list_keys_by_prefix(
```

src/reformatters/noaa/gefs/utils.py
```diff
@@ -18,9 +28,72 @@
     nomads_rate_limiter,
 )
 
+log = get_logger(__name__)
+
+GEFS_S3_LOCATION_PREFIX: Final = "s3://noaa-gefs-pds/"
+GEFS_S3_BUCKET_REGION: Final = "us-east-1"
```
aldenks:
```suggestion
GEFS_NODD_S3_LOCATION_PREFIX: Final = "s3://noaa-gefs-pds/"
GEFS_NODD_S3_BUCKET_REGION: Final = "us-east-1"
```

aldenks:
You were right, the listing was overthinking it. Rewritten with two constants — **16 files / 319 insertions → 5 files / 81**, and the functional change is now five lines.

What convinced me: the gate only ever has to separate two init times, the newest at **6h33m** old and the previous at **30h33m**. Any delay in between splits them, so the constant has a **24-hour-wide window** to land in. I had assumed the boundary would be tight because the first f390 file appears at init+6h34m — only ~2 min after cron — but that boundary is irrelevant; the gate is on cycle *age*, not on the production frontier.

Constants, from 28 days of `noaa-gefs-pds` `last_modified`:

| | |
|---|---|
| f≤384 set completes | init+6h31m .. init+6h48m |
| first f>384 file appears | init+6h34m .. init+6h36m |
| f>384 set completes (f840) | init+25h04m .. init+27h07m |

`GEFS_PRE_EXTENSION_MAX = 384h`, `GEFS_EXTENSION_COMPLETE_DELAY = 28h`. 28h rather than the midpoint so it also sits above the ~27h extension completion — a catch-up run on a mid-age cycle then gets gated correctly instead of asking for a half-published extension.

Lead times ≤384h are still all requested, so the mirror-lag fallback is untouched. The only thing given up is opportunistically catching the few extension files that land in the minutes after f384 — the occasional `horizon_max=390` — and the next day's run reprocesses that init to 840h anyway.

Two notes on your review comments, since the code they pointed at is gone:

- **`list_keys_by_prefix`** — reverted to the private `_list_objects`. It only needed to be public for the listing helper, so `virtual_source_listing.py` is back to its original name and shape apart from the docstring.
- **`GEFS_NODD_S3_*`** — reverted along with the hoist, so those constants are private in `forecast_10_day_spatial/region_job.py` again as `_S3_LOCATION_PREFIX` / `_S3_BUCKET_REGION`. Your naming point stands on its own; happy to do it as a separate one-liner if you want it, but it seemed like scope creep in a file this PR no longer touches.

I kept the module docstring cut, since that request was about the file on its own merits.

`pytest -m "not slow"` — 1997 passed.

---

Issue #620, opened 2026-05-25
NOAA GFS: stale upstream `.idx` sidecars cause silent data loss in ingest (2022-11-29/30 example)


aldenks:
## Summary

Some NOAA GFS GRIB files from late November 2022 on `s3://noaa-gfs-bdp-pds/` have **stale `.idx` sidecars** — byte offsets in the published `.idx` do not align with GRIB message boundaries in the corresponding GRIB file. The byte-range download path in `src/reformatters/noaa/gfs/region_job.py` trusts those offsets, so when the idx is stale, the partial file written to disk is non-GRIB bytes and `rasterio.open` fails with:

```
RasterioIOError: '<path>' not recognized as being in a supported file format.
```

The original `noaa-gfs-analysis` backfill silently dropped every affected `(init, lead)` pair (~16 h gap on 2022-11-29 → 2022-11-30). A recent retry with the full `retry-filter:` from `unavailable_timestamps.txt` hit the same failure (~176 rasterio exceptions, zero timestamps recovered).

## Confirmed affected init cycles

Discovered during the [noaa-gfs-analysis validation pass](https://dataset-validation-reports.dynamical.org/noaa-gfs-analysis/drafts/v0.1.0_2026-05-24T13-15/validation_report.html):

| Init cycle | Known affected leads (likely others within these inits exist too) |
|---|---|
| `2022-11-29T18Z` | f002, f003 |
| `2022-11-30T00Z` | f002, f005 |
| `2022-11-30T06Z` | f002, f003, f004, f005 |

Adjacent leads at the same cycles (f001, f004 of 18Z; f003, f004 of 00Z; f001 of 06Z) have correctly-aligned idx files and were ingested fine. Independent control samples (e.g. `2024-01-01T12Z f003`) are also fine. The pattern is per-source-file, not per-cycle.

## Evidence

For `s3://noaa-gfs-bdp-pds/gfs.20221130/00/atmos/gfs.t00z.pgrb2.0p25.f005`:

- **Full GRIB is intact.** 542 109 834 bytes, starts with `GRIB` magic, ends with `7777`, opens in rasterio (`driver=GRIB, count=743 bands, shape=(721, 1440)`), `TMP:2 m above ground` reads successfully (min/max −45.51/40.67 °C).
- **Published `.idx` is intact-looking.** 40 492 bytes, 743 lines.
[...]

PR #921, opened 2026-08-20
Detect stale GRIB index byte ranges in GFS/GEFS downloads

mrshll:
## What

[REFORMATTERS-EH](https://dynamical.sentry.io/issues/REFORMATTERS-EH): `RasterioIOError: not recognized as being in a supported file format` when reading downloaded GFS byte ranges, 1183 occurrences since 2026-05-24, recurring for `noaa-gfs-forecast`/`noaa-gfs-analysis` backfills of the 2022-11-29/30 GFS cycles.

## Root cause

Confirmed against the live NOAA S3 archive rather than guessed from code:

- The full `gfs.20221129/12/atmos/gfs.t12z.pgrb2.0p25.f207` object starts with a valid `GRIB` magic number, and its `.idx` sidecar cleanly parses to a single, unambiguous byte range for the `APCP:surface:204-207 hour acc fcst` message we compute via `grib_message_byte_ranges_from_index`.
- Fetching exactly that byte range from the real object returns bytes that are **not** a GRIB message (no `GRIB` header, no `7777` trailer).
- The `.idx` file's `Last-Modified` (2022-11-29 16:25 UTC) is a full day **before** the `.grb2` file's (2022-11-30 16:09 UTC) — NOAA reprocessed the archived GRIB2 file without regenerating its index sidecar, so the sidecar's byte offsets no longer line up with the reprocessed file's message boundaries.

This is an upstream NOAA archive inconsistency for these specific historical cycles, not a bug in our byte-range math — the same `_lead_time_str`/index-parsing logic correctly identifies a single matching entry; the entry's byte offsets are just wrong for the file as it exists today.

## Fix

A byte range built from a stale index can't be repaired without re-scanning the whole (500+MB) file, so this treats it like a missing source file instead: `assert_downloaded_grib_message` (new, in `noaa_grib_index.py`) checks the downloaded range starts with `GRIB` and raises `FileNotFoundError` if not. That routes it through the not-found handling GFS and GEFS already have for genuinely-missing files — soft `log.info` for old backfills, fallback to NOMADS for recent init times — instead of surfacing an opaque `RasterioIOError` deep in the read path. Wired into both `noaa/gfs/region_job.py` and `noaa/gefs/utils.py`, which share the same idx-byte-range-download pattern.

Scoped to GFS/GEFS, where I verified the failure against real source data. HRRR uses the same `grib_message_byte_ranges_from_index` utility but has its own duplicate-element handling I didn't want to touch without separately verifying it hits the same failure mode — a natural follow-up if it ever does.

## Test plan

- [x] New tests in `tests/noaa/noaa_grib_index_test.py::TestAssertDownloadedGribMessage` covering both the valid and invalid-magic-bytes cases
- [x] Updated existing GFS/GEFS `download_file` mocks to return real GRIB-prefixed temp files instead of bare `Mock()`s, since they now flow through the new check
- [x] `uv run ruff format && uv run ruff check --fix && uv run ty check` clean
- [x] `uv run pytest -m "not slow"` — 1997 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01QefBbWWqP75BFjajYF4xX4)_

### DIFF
--- a/src/reformatters/noaa/gefs/utils.py
+++ b/src/reformatters/noaa/gefs/utils.py
@@ -12,7 +12,10 @@ from reformatters.noaa.gefs.gefs_config_models import (
     GefsSourceFileCoord,
     get_grib_element,
 )
-from reformatters.noaa.noaa_grib_index import grib_message_byte_ranges_from_index
+from reformatters.noaa.noaa_grib_index import (
+    assert_downloaded_grib_message,
+    grib_message_byte_ranges_from_index,
+)
 from reformatters.noaa.noaa_utils import (
     NOMADS_RETRY_STATUS_CODES,
     nomads_rate_limiter,
@@ -53,12 +56,14 @@ def _download_file_from_gefs_source(
         idx_local_path, _index_data_vars(coord), coord.init_time, coord.lead_time
     )
     vars_suffix = digest(f"{s}-{e}" for s, e in zip(starts, ends, strict=True))
-    return download(
+    local_path = download(
         source_url,
         dataset_id,
         byte_ranges=(starts, ends),
         local_path_suffix=f"-{vars_suffix}",
     )
+    assert_downloaded_grib_message(local_path)
+    return local_path
 
 
 def gefs_download_file(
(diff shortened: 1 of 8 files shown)


aldenks:
this is one spot in the history, only comes up on backfills, is tracked in #620 and not worth +70 lines of code

PR #931, opened 2026-08-24
Add full-ensemble WeatherNext 2 virtual forecast

mrshll:
# PR 931 — WeatherNext 2 zero-copy native chunks

Tracking issue: dynamical-org/meta#188

## Product topology and assumptions

The annual and operational source stores have incompatible native chunk grids, so they
will be published as two logical datasets rather than represented as one rechunked array:

- `google-weathernext2-forecast-historical-virtual`: the fixed annual 2022–2024
  archive. Surface chunks contain four ensemble members and pressure chunks contain four
  members at all 13 levels.
- `google-weathernext2-forecast-operational-virtual`: the per-initialization archive
  from 2025 onward. Surface and pressure chunks contain one member; pressure chunks also
  contain one level.

Both products retain the same 64 members, 60 leads, 8 surface variables, 6 pressure
variables, float32 values, and full 721 by 1440 grid. Only the operational product has
scheduled updates; the historical product exposes suspended update and validation
resource templates so manual backfills can reuse the deployed image and pod sizing. Each
virtual reference covers one complete compressed source Zarr v2 chunk at byte offset zero.
Target encodings retain the source Blosc LZ4/shuffle codec; the proxy must not decode,
normalize, or split chunks.

The target pressure dimension remains last. For historical chunks, a standard Zarr v3
`TransposeCodec` will map the source `(init, member, lead, level, latitude, longitude)`
chunk to `(init, member, lead, latitude, longitude, pressure_level)` at read time. The
operational pressure grid uses singleton level chunks and will use the same target
dimension order without changing source bytes.

The products expose the source-native ascending latitude and 0–360 longitude as `y` and
`x` dimensions, respectively. Their coordinate metadata retains the CF latitude and
longitude identities. Historical pressure chunks use the standard transpose codec and
expose the source's ascending pressure levels with pressure level last.

The operational dataset preserves the strict `valid_time > 48h` gate. The historical
dataset is fixed and entirely older than the gate.

## Infrastructure prerequisite

Canonical references use
`GET https://wn.dynamical.org/chunks/<exact-source-object-key>`, restricted to
`weathernext_2_0_0/zarr/`; GET and Range stream untouched GCS objects. Discovery uses the
separate paginated GCS passthrough
`GET /objects?prefix=<array-prefix>&matchGlob=<array-prefix>{0,1,...,63}.{lead-index}.*&delimiter=%2F&maxResults=1000[&pageToken=<opaque>]`
and consumes `items[].name`, `items[].size`, `items[].md5Hash`, and `nextPageToken`. The GCS JSON `etag` is generation-based and does not match the chunk route HTTP ETag; the listed base64 MD5 is converted to the quoted hexadecimal ETag enforced by `If-Match`.
Each operational query returns only one variable/init/lead's member and pressure-level
objects, bounded to one page. References contain no plane selector or query. Direct GCS
references are not emitted, so readers and Kubernetes jobs receive no Google credentials.

Both routes are deployed with trusted-path authorization and the operational listing glob restriction. The chunk route streams exact objects and honors matching and stale `If-Match` values with 206 and 412 responses, respectively.

## Tasks

- [x] Prove the native Blosc plus client-side transpose pipeline with bounded source-shaped chunks.
- [x] Split the template and region-job topology into historical and operational products.
- [x] Generate whole-object references with paginated-list lengths and no plane selector.
- [x] Preserve the operational lag boundary in generated references.
- [x] Bound operational discovery to one GCS glob page per variable/init/lead.
- [x] Regenerate and review both templates.
- [x] Run focused dataset tests, common template/virtual tests, ruff, and ty.
- [x] Propagate source object identity through `VirtualRef` into Icechunk manifests.
- [x] Run live integration against the native Worker endpoint.
- [ ] Build and validate bounded historical and operational staging repositories.
- [ ] Benchmark manifest size and construction memory before the full historical run.
- [ ] Run the historical 2022–2024 and operational 2025–cutoff manifest backfills.
- [ ] Enable operational updates and publish both catalog entries with concatenation documentation.
- [x] Update the draft PR.

### 2026-08-24 — Native-chunk redesign

The prior 2.24-billion-plane design is superseded. Native chunk preservation removes
proxy-side decompression and rechunking, while separate products preserve the annual and
operational grids. A bounded synthetic proof decoded untouched source-format Blosc
LZ4/shuffle bytes and used `TransposeCodec(order=(0, 1, 2, 5, 3, 4))` to expose historical
pressure level last.

### 2026-08-24 — Reformatter implementation

Added the two registered dataset products and their independent templates. The historical
template uses native `(1, 4, 1, 721, 1440)` surface chunks and
`(1, 4, 1, 721, 1440, 13)` pressure-last chunks; the operational template keeps singleton
member and level chunks. Both declare the source Blosc LZ4/byte-shuffle codec. The
registered `dynamicalorg.latitude_longitude` filter flips and rolls full spatial chunks,
and historical pressure appends the standard transpose filter. The fixed historical append
axis is capped before 2025.

Discovery stores listed object sizes in the emitted offset-zero refs. Historical requests
list one fixed init prefix at a time; operational requests use a lead-specific GCS glob so
they never accumulate every object in a per-init array. No backfill was started. The
obsolete mixed-plane template was removed.

### 2026-08-24 — Full validation

The historical product uses suspended Kubernetes resource templates rather than active
cron execution, which keeps it manually backfillable through the existing deployment
workflow. Both datasets appear in manual backfill choices, while their resource templates
remain available for image and pod configuration discovery. Scoped formatting, ruff, and
ty checks pass. The full suite passes with 2,323 tests and 12 environment-gated skips.

### 2026-08-24 — Final review fixes

Operational backfill jobs now share a single strict 48-hour cutoff, not only scheduled
updates. Historical validation and manual repair use the final fixed one-day window rather
than entering the operational-only factory. Historical coordinate statistics use fixed
archive dates. Gated native-route tests cover every variable, missing values, unit filters,
pressure transposition, and successive precipitation intervals in both source eras.

### 2026-08-24 — External PR review

Claude reviewed PR 931 in Herdr and posted six inline findings. Operational object globs
now enumerate members explicitly before the lead index, direct invocations use the same
UTC clock for template and publication cutoff, and transient listing requests retry.
Availability is isolated per variable and logs the first missing object, so an incomplete
variable no longer suppresses unrelated variables. The four-init manifest batch remains
to cap construction near 1.4 million virtual refs instead of roughly 11.2 million for a
32-init batch. The redundant serializer assertion was removed. Scoped formatting, ruff,
and ty checks pass; the full suite passes with 2,326 tests and 12 skips.

### 2026-08-25 — Deployed proxy and reference identity

Confirmed both Worker routes are live. GCS JSON listings expose a generation-based `etag`, while the chunk endpoint exposes and enforces the object MD5 as its HTTP ETag. References therefore derive the quoted hexadecimal HTTP ETag from the listed `md5Hash`, retaining listing-only construction while making Icechunk conditional reads valid. Shared `VirtualRef` emission now carries optional ETag checksums into `VirtualChunkSpec`.

Both gated live integrations pass against the proxy, covering every variable, both native layouts, all expected member references, pressure transpose, missing values, and precipitation semantics. A dedicated conditional request check confirms matching and stale ETags return 206 and 412. Scoped ruff and ty pass. The full suite repeatedly reaches unrelated validation tests before a native xdist/process abort.

### 2026-08-25 — Native spatial orientation

PR review rejected a custom latitude/longitude codec because it would complicate access
for consumers. Both products now expose the native ascending latitude and 0–360 longitude
coordinates under `y` and `x`, matching the catalog's projected-grid dimension naming
convention while retaining CF latitude/longitude coordinate metadata. The custom codec
and its package registration were removed; the standard historical pressure transpose
remains. The 395-test focused suite and all three gated live integration tests pass.

### 2026-08-26 — Documentation scope

PR review removed the `docs/virtual_datasets.md` changes because the shared guide should
not carry WeatherNext-specific implementation detail. The dataset design remains recorded
in this PR plan.


src/reformatters/google/weathernext2/forecast_virtual/region_job.py
```diff
@@ -0,0 +1,417 @@
...
+            prefix = f"{store_key_prefix}{var.internal_attrs.source_name}/"
+            if self.source_layout == "historical":
+                queries.append(
+                    ObjectListingQuery(f"{prefix}{coord.annual_init_index}.")
+                )
+            else:
+                queries.append(
+                    ObjectListingQuery(
+                        prefix=prefix,
+                        match_glob=f"{prefix}*.{coord.lead_index}.*",
```
mrshll: **Medium — the glob does not bound the listing to one lead, and for lead 0 it matches the whole array.**

`*` in a GCS `matchGlob` matches any run of non-`/` characters, so `prefix*.{L}.*` matches an object whose key contains `.{L}.` at *any* dotted position, not just the lead position.

Operational keys are `{member}.{lead}.{level}.0.0` (pressure) and `{member}.{lead}.0.0` (surface):

- **`lead_index == 0`** (lead_time 6h, i.e. every init's first lead): every key ends in `.0.0`, which contains `.0.`, so `*.0.*` matches **every object in the array** — 64x60x13 = 49,920 pressure objects, i.e. 50 paginated round-trips instead of the intended 1, per pressure variable per init.
- **any other lead L**: also matches every object whose *pressure-level* index is `L` (64x60 = 3,840 extra objects) on top of the 832 real ones.

So the PR's "bounded to one GCS glob page per variable/init/lead" does not hold. It is not a correctness bug (`locations <= coord_objects.keys()` still selects the right subset), but it multiplies listing cost by ~5-50x, and `discover_available` re-runs on every 1s poll tick of the operational update.

Anchoring the lead component fixes it, e.g. glob on the two members' worth of positions explicitly or list by a per-lead prefix. Since the lead is not the leading key component there is no prefix that isolates it; one option is `matchGlob=f"{prefix}*.{lead_index}.[0-9]*.0.0"`-style patterns per layout, or dropping the glob and listing the whole array once per (var, init) and caching it the way the historical path already does.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
```diff
@@ -0,0 +1,417 @@
...
+    source_layout: ClassVar[SourceLayout] = "historical"
+    manifest_init_split: ClassVar[int] = ROOT_MANIFEST_INIT_SPLIT
+    operational_update_window: ClassVar[Timedelta] = pd.Timedelta("1D")
+
+
+class GoogleWeathernext2ForecastOperationalVirtualRegionJob(
+    GoogleWeathernext2ForecastVirtualRegionJob
+):
+    source_layout: ClassVar[SourceLayout] = "operational"
+    manifest_init_split: ClassVar[int] = PRESSURE_MANIFEST_INIT_SPLIT
```
mrshll: **Medium — batching backfill commits at the *finer* split rewrites every root manifest 8 times.**

The operational dataset splits manifests at 4 inits for `/pressure_level/*` and 32 inits for root arrays. `process_virtual_refs` groups a backfill's coords by `init_index // manifest_init_split`, so with `manifest_init_split = 4` each commit covers 4 inits: the pressure manifests line up exactly (one write each), but each 32-init **root** manifest is touched by 8 consecutive commits and so is rewritten 8 times (~123k refs each, x 8 root arrays).

Batching at the coarser split instead (`ROOT_MANIFEST_INIT_SPLIT = 32`, a multiple of 4) writes *both* families exactly once: a 32-init batch covers one whole root manifest and eight whole pressure manifests. That is what the historical job effectively gets, since its single split is 32.

If 4 was chosen to bound per-commit memory (a 32-init batch would hold ~10.5M `VirtualRef`s: 32 inits x 60 leads x 5,504 refs), that trade-off deserves a comment here, because the name and the surrounding manifest-splitting design read as "batch = one manifest window" and this value silently isn't.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
```diff
@@ -0,0 +1,417 @@
...
+            tmp_store=tmp_store,
+            get_template_fn=get_template_fn,
+            append_dim=append_dim,
+            all_data_vars=all_data_vars,
+            reformat_job_name=reformat_job_name,
+            job_fire_time=job_fire_time,
+        )
+        (job,) = jobs
+        assert isinstance(job, cls)
+        fire_time = job_fire_time or pd.Timestamp.now()
```
mrshll: **Medium — the fallback clock here is local time, while the backfill path's is UTC, so the same 48h gate lands in two different places.**

`_current_publication_cutoff()` (used by `get_jobs`) is `pd.Timestamp.now(tz="UTC").tz_localize(None)`; this line is a bare `pd.Timestamp.now()`, which is naive **system-local** time. On a non-UTC host the two disagree by the UTC offset:

```
$ TZ=America/New_York python -c "..."
utc-localized-none: 2026-08-24 22:20:09
naive now        : 2026-08-24 18:20:09
```

East of UTC (e.g. `TZ=Europe/Berlin`) local-now is *ahead* of UTC, so `fire_time - 48h` is up to a couple of hours later than the real cutoff and this path emits refs to source objects whose valid_time is younger than the 48-hour publication gate the PR describes as the authorization boundary. Westward it is merely inconsistent with the backfill.

Only reached when `job_fire_time` is None (direct calls / tests today), but it is the one clock in this file that isn't UTC — `_current_publication_cutoff()` is right there.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
```diff
@@ -0,0 +1,417 @@
...
+) -> dict[str, int] | None:
+    objects: dict[str, int] = {}
+    page_token: str | None = None
+    while True:
+        params = {"prefix": query.prefix, "maxResults": "1000"}
+        if query.match_glob is not None:
+            params.update({"matchGlob": query.match_glob, "delimiter": "/"})
+        if page_token is not None:
+            params["pageToken"] = page_token
+        response = client.get(OBJECTS_LOCATION, params=params)
```
mrshll: **Low — one transient proxy error kills the whole job.**

`client.get` has no retry and `raise_for_status()` turns any 5xx into an exception. That exception propagates out of `pool.map` -> `discover_available` -> `process_virtual_refs`, and nothing above catches it (`_file_refs_or_skip` only guards `file_refs`). A single 502/504/`ReadTimeout` from `wn.dynamical.org` therefore aborts the entire update or backfill worker.

That is a real exposure for the operational update, which is a *polling* job: it re-enters `discover_available` roughly every second for up to its 30-minute deadline, so it makes thousands of listing requests per fire and only needs one of them to blip. The repo already has `reformatters.common.retry` for exactly this (`http_download_to_disk` uses it); wrapping the per-page GET in it would keep a blip from ending the fire.

(Only 403/404 are treated as "not published yet"; everything else is fatal, which is the right split — it just needs a retry in front of it.)

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
```diff
@@ -0,0 +1,417 @@
...
+        for coord in pending:
+            coord_objects: dict[str, int] = {}
+            for query in self._listing_queries(coord):
+                objects = listed[query]
+                if objects is None:
+                    break
+                coord_objects.update(objects)
+            else:
+                locations = {chunk.location for chunk in self._source_chunks(coord)}
+                if locations <= coord_objects.keys():
```
mrshll, line 331: **Low — availability is all-or-nothing across 14 variables and thousands of objects, and a backfill only sweeps once.**

A coord is accepted only when *every* location it needs is listed: 224 objects for historical, **5,504** for operational (8 root vars x 64 members + 6 pressure vars x 64 members x 13 levels). One absent object — the source is Zarr v2, which is free to omit a chunk that is entirely `fill_value` — drops the whole (init, lead) for all 14 variables.

In a backfill that is permanent: `process_virtual_refs` in backfill mode calls `discover_available` once and returns, so the coord is logged as "not present, skipping" and never retried. `CheckVirtualManifestCompleteness()` (default `(1.0,)`) would then fail on that init forever, with nothing in the message pointing at the single missing object.

Worth at least logging *which* locations were missing when a coord is rejected, so the failure is diagnosable; better would be emitting refs for the objects that do exist and leaving the rest to `fill_value`.

tests/common/common_template_config_subclasses_test.py
```diff
@@ -274,12 +274,14 @@ def test_coordinates_have_single_chunk(
...
-    GDAL flip) and adjust_longitude_range (monotonic -180..180 longitude)."""
+    """Every GRIB-decoding virtual data var's GribberishCodec flips decoded messages into
+    our shared orientation: north_up (first row = largest latitude, matching materialized
+    datasets' GDAL flip) and adjust_longitude_range (monotonic -180..180 longitude)."""
     for var in dataset.template_config.data_vars:
         serializer = var.encoding.serializer
         assert serializer is not None, var.name
+        if serializer["name"] != "gribberish":
+            continue
         assert serializer["name"] == "gribberish", var.name
```
mrshll: **Low — dead assertion.** The `continue` on the line above already guarantees `serializer["name"] == "gribberish"`, so this assert can never fire. Drop it; the guard is the check now.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Fixed in e37e0a9e. The operational glob now explicitly enumerates members `{0,1,...,63}` immediately after the array prefix, with the lead index anchored as the second chunk-key component. The regression test asserts the complete `matchGlob`, covering the lead-0 and pressure-level ambiguity.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Kept at 4 intentionally and documented the bound in e37e0a9e. Each init contributes 350,208 virtual refs across the full product, so a 32-init batch would construct about 11.2M refs in memory versus about 1.4M for 4 inits. The additional manifest rewrites are the preferred tradeoff here.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Fixed in e37e0a9e. Direct invocations now use an explicit UTC clock, and the same `fire_time` drives both template generation and the strict 48-hour publication cutoff. Added a regression test with a fixed UTC clock.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Fixed in e37e0a9e. Each listing page now uses the shared bounded retry helper for transport errors plus HTTP 408, 429, and 5xx responses; 403/404 retain their availability semantics. Added a 502-then-success regression test.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Improved in e37e0a9e. Availability coordinates are now split per variable, so one incomplete variable no longer suppresses the other variables for that init/lead. A missing coordinate also emits the missing count and first exact object key at debug level; completeness remains atomic within each variable.

tests/common/common_template_config_subclasses_test.py
mrshll: Fixed in e37e0a9e by removing the unreachable assertion.

docs/virtual_datasets.md
```diff
@@ -1,6 +1,6 @@
...
 
-A virtual dataset is an Icechunk store whose chunks are references — `(location, offset, length)` pointers into source files (e.g. GRIB messages) on the provider's object store — rather than copied bytes. A per-variable zarr serializer (e.g. `GribberishCodec`) decodes the referenced bytes at read time. Virtual datasets are written by `VirtualRegionJob` subclasses (`src/reformatters/common/virtual_region_job.py`).
+A virtual dataset is an Icechunk store whose chunks are references — `(location, offset, length)` pointers into source files (e.g. GRIB messages) on the provider's object store — rather than copied bytes. The referenced bytes are decoded at read time by the codec pipeline the variable's encoding declares: for a GRIB source, a per-variable zarr serializer (`GribberishCodec`) with `compressors=()`; for a source that is already a Zarr archive, either the standard bytes serializer plus the source's own compressor points at a whole source chunk object, or an authenticated transform proxy returns deterministic bytes matching the target codec pipeline. Virtual datasets are written by `VirtualRegionJob` subclasses (`src/reformatters/common/virtual_region_job.py`).
```
aldenks: Revert this

src/reformatters/google/weathernext2/forecast_virtual/template_config.py
```diff
@@ -0,0 +1,631 @@
...
+                    dtype="int64",
+                    fill_value=0,
+                    chunks=(),
+                    shards=None,
+                ),
+                attrs=CoordinateAttrs(
+                    units=None,
+                    statistics_approximate=None,
+                    grid_mapping_name="latitude_longitude",
+                ),
```
aldenks, line 363: needs a real spatial ref

tests/common/datasets_cf_compliance_test.py
```diff
@@ -109,88 +109,72 @@ def test_cf_latitude_longitude_recognized(
     """
     Ensure latitude and longitude coordinates are recognized as CF coordinates.
     CF requires these to have standard_name and units attributes.
-    For non-projected datasets, they should also have axis attributes.
+    Coordinate semantics, rather than dimension names, distinguish geographic grids.
```
aldenks: I don't understand all the changes in this test_cf_latitude_longitude test, can we revert them? this dataset uses dims x/y, can it be handled like other projected datasets? the changes here may be a generalization, but are probably an overgeneralization for this one odd dataset that we don't want to encode too generally in our common tests. if we need a simple opt out for it here and then have weathernext2 specific template config tests cover these things that's ok, but im not sure we even need that.

tests/common/datasets_cf_compliance_test.py
```diff
@@ -963,22 +948,28 @@ def test_metadata_consistency_across_datasets() -> None:
...
         template_config = dataset.template_config
         for coord_config in template_config.coords:
-            by_coord_name.setdefault(coord_config.name, {})[dataset.dataset_id] = {
+            standard_name = coord_config.attrs.standard_name
+            identity = coord_config.name
+            if identity in {"x", "y"}:
+                identity = (
+                    f"{coord_config.name} ({standard_name or 'no standard_name'})"
+                )
+            by_coord_identity.setdefault(identity, {})[dataset.dataset_id] = {
```
aldenks: why do x/y need new special handling here?

src/reformatters/__main__.py
```diff
@@ -243,6 +251,13 @@ class UpstreamGriddedZarrsDatasetStorageConfig(StorageConfig):
     EcccHrdpsForecastTemporalDynamicalDataset(
         primary_storage_config=SourceCoopZarrDatasetStorageConfig(),
     ),
+    # Google
+    GoogleWeathernext2ForecastHistoricalVirtualDataset(
+        primary_storage_config=UpstreamGriddedZarrsIcechunkDatasetStorageConfig(),
+    ),
+    GoogleWeathernext2ForecastOperationalVirtualDataset(
+        primary_storage_config=UpstreamGriddedZarrsIcechunkDatasetStorageConfig(),
```
aldenks: probably not the right storage location for these

mrshll: Moved in b4dfdfd2. Both products now use `Weathernext2IcechunkDatasetStorageConfig` — a dedicated R2 bucket `s3://dynamical-weathernext2` with its own `weathernext2-storage-options-key` secret — instead of the Upstream gridded zarrs bucket. The one-line icechunk subclass of `UpstreamGriddedZarrsDatasetStorageConfig` this PR had added is removed; nothing else used it. The bucket and secret still need provisioning before a backfill.

tests/common/datasets_cf_compliance_test.py
mrshll: Reverted in b4dfdfd2 — this file is back to its `main` content apart from one `CF_UNITS_VARIANCES_ALLOWLIST` entry (`sea_surface_temperature`, `degree_Celsius`), which follows the existing allowlist pattern for the other Celsius temperatures.

The generalization only existed because the dataset had been renamed to `x`/`y` while keeping geographic coordinate metadata. That rename is undone: both products declare `latitude`/`longitude` again, still carrying the native ascending latitude and 0–360 longitude values, so the unmodified test passes on its geographic branch. Handling it as a projected dataset was the alternative, but `projection_x_coordinate` with `degree_east` units would be false metadata, and it would still have needed an opt-out because the `long_name`/`units` conflict with the projected `x`/`y` in HRRR.

tests/common/datasets_cf_compliance_test.py
mrshll: Removed in b4dfdfd2 along with the rest of the changes to this file — with `latitude`/`longitude` back there is no `x`/`y` metadata conflict to disambiguate.

src/reformatters/google/weathernext2/forecast_virtual/template_config.py
mrshll: Added in b4dfdfd2. `spatial_ref` now declares the same CRS the other 0.25° global products declare — the WMO sphere of the ERA5 grid WeatherNext 2 runs on — with `crs_wkt`, the ellipsoid and datum fields, and `spatial_ref`. Its `comment` records that the source zarrs declare no CRS of their own, so the grid convention is where this comes from.

src/reformatters/__main__.py
mrshll: Follow-up on the bucket name: the provisioned bucket is `dynamical-wn2`, so `base_path` is `s3://dynamical-wn2` as of 620a2873. The `weathernext2-storage-options-key` secret is applied in the cluster and verified — `load_secret` returns the `icechunk.s3_storage` kwargs, and a head_bucket plus put/get/delete probe against the bucket succeed.

tests/common/datasets_cf_compliance_test.py
mrshll: checking with alden - I thought he wanted to use the x/y coord convention to imply the lat lng nuance

mrshll: Understood — x/y stays, and the diff is now an opt-out instead of a rewrite (0d86ebf4). `tests/common/datasets_cf_compliance_test.py` is +16/-1 against `main`: a `GEOGRAPHIC_XY_DATASET_IDS` constant, one condition skipping the projected x/y metadata assertions for those two ids, and one skipping their x/y in the cross-dataset coordinate consistency collection — needed because geographic `long_name`/`standard_name`/`units` on x/y conflict with HRRR's projected x/y. `native_topology_test.py` now asserts the geographic x/y metadata directly, so nothing is unchecked.

The earlier rename to latitude/longitude is reverted: both products declare y/x again over the native ascending latitude and 0–360 longitude values.

src/reformatters/google/weathernext2/forecast_virtual/region_job.py
mrshll: Revisited the split sizes against docs/virtual_datasets.md, measuring refs per init per array from the templates (4df483df).

The historical product was over-split. Its layout packs 4 members and all 13 levels per chunk, so both array groups hold 960 refs per init per array — a 32-init split filled manifests to only 0.48 MiB, a sixth of the single-level reader budget, across 1,918 manifests. Now 128: 1.92 MiB manifests, 490 manifests, and a 1.72M-ref commit batch — the same order as the operational batch discussed on this thread.

Operational is unchanged: root at 32 is 1.92 MiB against the 3 MiB single-level budget, pressure at 4 is 3.12 MiB against 5–8 MiB for vertical groups. The doc's low-hundreds-of-KiB rule targets latency-sensitive datasets; these inits are already 48 hours old at ingest, so commit CPU is not the binding cost. What may matter is manifest count — pressure at 4 adds ~2,190 manifests/year and icechunk scans the manifest list per commit. Pressure at 8 would be 6.25 MiB, still inside the vertical budget, and halve that growth. Holding that until the staging backfill measures real bytes/ref via `repo.list_manifest_files`, since these locations are long and repetitive and split size changes are incremental.

PR #938, opened 2026-08-24
Add the eccc-hrdps-forecast dataset

aldenks:

PR #940, opened 2026-08-25
Reduce HRDPS Datamart enumeration requests

mrshll:
## Summary

Reduce the HRDPS archiver's ECCC MSC Datamart traffic from about 730,000 to about 88,000 requests per day by disabling rclone's per-file HEAD requests. This implements the immediate mitigation from dynamical-org/meta#189 independently of #938.

## Instrumented request reduction

The pre-change probe measured 417 source requests to list one 414-file lead-hour directory. Extrapolated across a 49-lead, 20,021-file initialization, the old traversal costs about 20,100 requests per initialization whether or not any files need copying.

This PR was tested against the fully published 2026-08-23 00Z initialization with rclone request tracing enabled:

```sh
/usr/bin/rclone copy ':http:/20260823/WXO-DD/model_hrdps/continental/2.5km/00/' /tmp/hrdps-no-head-probe --http-url=https://dd.weather.gc.ca --http-no-head --ignore-existing '--filter=+ *.grib2' '--filter=- *' --fast-list --dry-run --transfers=1 --checkers=1 --dump=requests --log-level=DEBUG --log-file=/tmp/hrdps-no-head-probe.log
```

The dry run traversed the complete 20,021-file initialization and emitted:

| Request/result | Count |
|---|---:|
| Directory GETs | 50 |
| HEAD requests | 0 |
| GRIB GETs | 0 |
| Files inventoried | 20,021 |

A separate one-file copy made two directory GETs plus one GRIB GET, produced a 3,613,289-byte GRIB matching ECCC's `Content-Length`, and opened as the expected 2540 x 1290 grid. Repeating that copy made the same two directory GETs but no GRIB GET, confirming `--ignore-existing` still skips existing destinations.

This reduces enumeration from about 20,100 to exactly 50 requests per initialization: about 20,050 fewer requests, a 99.75% reduction (roughly 402x). Production examines 32 date/initialization pairs per day, so measured traversal behavior projects to:

| Daily ECCC traffic | Before | After |
|---|---:|---:|
| Archive enumeration | ~643,200 | 1,600 |
| Archive downloads | ~80,000 | ~80,000 |
| Reformatter updates | 6,624 | 6,624 |
| **Total** | **~730,000** | **~88,000** |

The change therefore removes about 642,000 ECCC requests per day—approximately 88% of total traffic—while preserving the full-fidelity archive and retry window. The remaining traffic is dominated by the files we intentionally download, and remains slightly above MSC's contact threshold.

## Plan

- [x] Add a failing command-contract test for no-HEAD directory traversal and removal of the ineffective age filter.
- [x] Add `--http-no-head`, terminate the source directory path with `/`, and remove `--min-age=1m`.
- [x] Document the corrected request estimate and residual partial-source risk.
- [x] Verify a filtered live copy, an idempotent repeat, and a no-HEAD dry run.
- [x] Run focused tests, the full suite, ruff, and ty.
- [x] Notify PR #938 and report results on meta issue #189.

## Decisions

- Branch from `origin/main`; do not stack on PR #938.
- Keep the schedule, `days_back`, initialization hours, field coverage, and public CLI unchanged.
- Accept the residual partial-source risk rather than adding a completeness gate or AMQP redesign in this change.

No public Python or CLI interfaces change.

Without HEAD requests, rclone cannot apply the old `--min-age=1m` guard because source modification times are unavailable. If ECCC exposes a file before it has finished publishing, `--ignore-existing` would prevent a later run from repairing that object automatically. The archive normally starts after a complete run is available, but this residual risk remains.

## Validation

- `uv run pytest tests/eccc/hrdps/archive_gribs/copy_files_from_eccc_test.py`: 4 passed
- `MPLBACKEND=Agg uv run pytest`: 2,300 passed, 10 skipped
- `uv run ruff format`: clean
- `uv run ruff check --fix`: clean
- `uv run ty check --output-format concise`: clean

The first full suite run selected the Qt backend and aborted because this headless host lacks the xcb plugin. The affected plotting tests and full suite passed with `MPLBACKEND=Agg`; this is an environment issue unrelated to the change.

## 2026-08-25 — Retro

The implementation stayed at the rclone command boundary. The important addition to the issue's suggested flag was the trailing source slash; without it, rclone treats the path as a file in no-HEAD mode. Removing `--min-age` also avoids presenting an age guard that cannot operate without source modification times.

### DIFF
--- a/src/reformatters/eccc/hrdps/archive_gribs/copy_files_from_eccc.py
+++ b/src/reformatters/eccc/hrdps/archive_gribs/copy_files_from_eccc.py
@@ -78,11 +78,11 @@ def _copy_one_init_hour(
     cmd = (
         "/usr/bin/rclone",
         "copy",
-        f":http:{src_path}",
+        f":http:{src_path}/",
         dst_path,
         f"--http-url={MSC_DATAMART_HOST}",
+        "--http-no-head",
         "--ignore-existing",
-        "--min-age=1m",  # Ignore files that are so young they might still be mid-upload.
         "--filter=+ *.grib2",
         "--filter=- *",
         "--s3-no-check-bucket",  # Workaround for reformatters issue #428
--- a/src/reformatters/eccc/README.md
+++ b/src/reformatters/eccc/README.md
@@ -11,6 +11,31 @@ window of GRIB2 files. `hrdps/archive_gribs/copy_files_from_eccc.py` copies file
 the Datamart to a public Source Co-Op bucket using [`rclone`](https://rclone.org),
 preserving the Datamart's own `{date}/{init_hour}` directory structure.
 
+### MSC Datamart request volume
+
+ECCC's [MSC Open Data usage policy](https://eccc-msc.github.io/open-data/usage-policy/readme_en/)
+asks users making at least 86,400 requests per day to contact them and directs systematic
+retrieval away from directory listings.
+
+The HRDPS archive contains about 20,000 files per run. Without `--http-no-head`,
+rclone's HTTP backend sends one HEAD request per entry while traversing the Datamart's
+directory listings. The archive cron examines eight date and initialization-hour pairs
+per invocation and runs four times per day, producing about 640,000 HEAD requests in
+addition to the day's roughly 80,000 file downloads. Reformatter updates add about
+6,600 requests, for a total near 730,000 requests per day.
+
+The archive enables `--http-no-head`, reducing traversal to roughly one GET per
+directory, or about 1,600 enumeration requests per day. Downloads and reformatter
+updates bring the resulting total to about 88,000 requests per day, so contacting MSC
+is still required.
+
+Without HEAD requests, rclone does not know source file sizes or modification times
+while listing. `--ignore-existing` still unconditionally skips paths already present at
+the destination, but rclone cannot use an age filter to avoid a source file that has
+become visible while it is still being published. If such a file were copied, later
+archive runs would not repair it automatically. The archive schedule normally starts
+after a complete run is available, but this residual risk remains.
+
 ### Testing locally
 
 `rclone` must be on `PATH` at `/usr/bin/rclone`. Test with:
--- a/tests/eccc/hrdps/archive_gribs/copy_files_from_eccc_test.py
+++ b/tests/eccc/hrdps/archive_gribs/copy_files_from_eccc_test.py
@@ -28,10 +28,12 @@ def test_copy_one_init_hour_builds_expected_command(mock_run_cmd: MagicMock) -> 
     cmd = mock_run_cmd.call_args[0][0]
     assert cmd[0] == "/usr/bin/rclone"
     assert cmd[1] == "copy"
-    assert cmd[2] == ":http:/20260704/WXO-DD/model_hrdps/continental/2.5km/00"
+    assert cmd[2] == ":http:/20260704/WXO-DD/model_hrdps/continental/2.5km/00/"
     assert cmd[3] == ":s3:bucket/eccc-hrdps-grib/20260704/00"
     assert "--http-url=https://dd.weather.gc.ca" in cmd
+    assert "--http-no-head" in cmd
     assert "--ignore-existing" in cmd
+    assert not any(arg.startswith("--min-age") for arg in cmd)
     assert "--filter=+ *.grib2" in cmd
     assert "--transfers=8" in cmd
     assert "--checkers=4" in cmd


src/reformatters/eccc/hrdps/archive_gribs/copy_files_from_eccc.py
```diff
@@ -78,11 +78,11 @@ def _copy_one_init_hour(
     cmd = (
         "/usr/bin/rclone",
         "copy",
-        f":http:{src_path}",
+        f":http:{src_path}/",
         dst_path,
         f"--http-url={MSC_DATAMART_HOST}",
+        "--http-no-head",
```
mrshll, line 84: **`--http-no-head` stamps every newly archived object with an epoch modification time.**

With `--http-no-head` rclone has no `Last-Modified` for source entries, so `Object.ModTime()` returns the zero time. Verified against a local HTTP source with this rclone:

```
$ rclone lsl ':http:/d/' --http-url=... 
    3000 2026-08-24 22:06:40.000000000 a.grib2
$ rclone lsl ':http:/d/' --http-url=... --http-no-head
      -1 1969-12-31 18:00:00.000000000 a.grib2
```

And the resulting S3 write (`--dump headers` against a stub endpoint) carries:

```
X-Amz-Meta-Mtime: 0
```

A local `--dst-root-path` copy likewise lands with an mtime of `Dec 31 1969`.

Concrete consequence: objects already in the Source Co-op bucket from before this change carry the Datamart's real `Last-Modified`, and everything written after it carries `0`. The archive becomes time-heterogeneous, so any future `rclone sync`/`check`/`--max-age` sweep, bucket lifecycle rule, or "when did we archive this run" question over the bucket silently splits at this commit. Worth either noting in the README section this PR adds, or passing `--no-update-modtime` deliberately so the intent is explicit.

mrshll, line 84: **Unknown source size switches every GRIB from a single `PUT` to a three-request S3 multipart upload.**

Because `--http-no-head` reports `Size() == -1`, rclone takes the streaming (`Rcat`) path, and anything over `--streaming-upload-cutoff` (default 100 KiB — HRDPS GRIBs are ~3.6 MB) becomes a multipart upload. Verified with a realistic 3,613,289-byte file against a stub S3 endpoint:

```
### before (HEAD, known size)
PUT  /bucket/pfx/big.grib2
HEAD /bucket/pfx/big.grib2

### after (--http-no-head)
POST /bucket/pfx/big.grib2?uploads=
PUT  /bucket/pfx/big.grib2?partNumber=1&uploadId=...
POST /bucket/pfx/big.grib2?uploadId=...
HEAD /bucket/pfx/big.grib2
```

(rclone also logs `Copied (Rcat, new)` instead of `Copied (new)`.)

Three concrete effects the PR's request accounting doesn't cover, since it only measures the ECCC side:

1. Destination writes go from 1 upload request per file to 3. At ~80,000 files/day that is ~160,000 extra Source Co-op requests per day.
2. Any interruption now leaves an **incomplete multipart upload** in the bucket: rclone's internal retries, the `retry(max_attempts=2)` wrapper around this call, and the cron's `pod_active_deadline=timedelta(hours=2)` are all ways a transfer dies mid-upload. Those parts accrue storage charges until aborted, and Source Co-op will not clean them up unless the bucket has an abort-incomplete-multipart lifecycle rule.
3. Multipart objects get a `"<md5>-N"` style ETag rather than the content MD5, so any consumer (or future `rclone check --checksum`) that treats the archive's ETag as an MD5 gets a different answer for objects written after this change than for those written before.

Setting `--s3-upload-cutoff`/`--streaming-upload-cutoff` above the largest HRDPS GRIB would keep the single-`PUT` path while retaining the enumeration savings.

src/reformatters/eccc/hrdps/archive_gribs/copy_files_from_eccc.py
```diff
@@ -78,11 +78,11 @@ def _copy_one_init_hour(
...
         "--ignore-existing",
```
mrshll, line 85: **Both guards against archiving a mid-publication GRIB are gone at once, and `--ignore-existing` makes the result permanent.**

The PR is right that `--min-age=1m` is inert under `--http-no-head` — I confirmed the listing reports `1969-12-31` for every entry, so the filter can no longer exclude anything. But the second guard goes too, and that one isn't mentioned in the README section: with `Size() == -1`, rclone's post-transfer verification short-circuits (`sizeDiffers` returns false as soon as either side's size is `< 0`), so `corrupted on transfer: sizes differ` can no longer fire for these copies.

Before this change a file that grew between the HEAD and the GET was caught by that size check and the transfer failed. After it, nothing checks.

Concrete scenario: the cron runs `0 4,10,16,22`, i.e. init+4h, and the README elsewhere records p99 publication at init+3h49m — so the tail of runs is still publishing when the job starts. Apache serves a partially-written `.grib2`, rclone copies it with no age filter and no size check, `--ignore-existing` means every subsequent run skips the path, and the truncated GRIB is permanent in a public immutable archive. The reformatter then reads that file forever.

The module docstring's `--ignore-existing` for idempotency (Datamart files are immutable once published) is now the sole load-bearing assumption, with nothing backstopping it. A cheap replacement for the lost guard would be skipping the initialization matching the current cron hour (only recheck initializations at least one cycle old), or a per-run file-count gate before the copy — the PR's own probe shows a complete run is exactly 20,021 files.

mrshll: Resolved in #941 with a six-hour minimum initialization age. The init+4h cron skips the newest run and the next cron archives it around init+10h, more than six hours after the observed init+3h49m p99 publication time. The test covers the exact six-hour boundary and skips newer/future initializations. The operations documentation also makes the missing post-transfer size comparison and residual exceptionally-late-publication risk explicit.

src/reformatters/eccc/hrdps/archive_gribs/copy_files_from_eccc.py
mrshll: Resolved in #941. I tested the suggested `--no-update-modtime` against a live ECCC no-HEAD copy; the destination still received an epoch modification time, so I did not add an ineffective flag. The ECCC operations documentation now states that new objects have epoch rclone mtimes while pre-change objects retain Datamart mtimes, and warns against using those mtimes across the transition for age or inventory decisions.

mrshll: Resolved in #941 by adding `--streaming-upload-cutoff=32Mi`. A complete public archived initialization contained 20,021 objects and the largest was 7,562,831 bytes, so the cutoff keeps every currently measured unknown-sized GRIB on the single-PUT path instead of multipart upload. The command-contract test pins the flag.

Issue #932, opened 2026-08-24
ECMWF ER Backfill


mrshll:
Dataset code landed in #904. The GRIB archiver cron landed in #922.

## State at open (2026-08-24)

- **Store:** does not exist. No backfill has been run.
- **Crons:** `...-update` (09 UTC) and `...-validate` (12 UTC) are deployed but suspended (`reformat_suspend = True`).
- **GRIB archive:** `s3://us-west-2.opendata.source.coop/dynamical/ecmwf-ifs-grib/ecmwf-ifs-ens-forecast-46-day/` holds 138 contiguous inits, **2026-04-07 → 2026-08-22**. The 06 UTC frontier cron is healthy; the backwards history walk appears to have stopped around 2026-08-20 at 2026-04-07 (inferred from object mod times).
- **`append_dim_start`:** `2023-06-28`, ~1010 inits earlier than the archive currently reaches.
- No validation report, no STAC entry, no prose file, no notebook.

## Store location

`s3://dynamical-ecmwf-ifs-ens/ecmwf-ifs-ens-forecast-46-day-1-5-degree/v0.1.0.icechunk` — a new prefix in the existing bucket, sibling to `ecmwf-ifs-ens-forecast-15-day-0-25-degree/`. No bucket creation needed; `index.html` lists prefixes dynamically, so nothing to edit there.

## Phase 1 — 1-month test backfill

Prove the pipeline end to end on a disposable store before committing to a large backfill.

- [ ] Branch on `main`: `dataset_version = "0.0.1"` (throwaway prefix, keeps v0.1.0 clean) and `append_dim_start = 2026-07-25T00:00`; `uv run main ecmwf-ifs-ens-forecast-46-day-1-5-degree update-template`; PR + merge (backfills run from `main`'s deployed image)
- [ ] `create-new-store` backfill via the **Manual: Backfill** action with an explicit `--append-dim-end 2026-08-23T00:00` (exclusive). The archive's newest init is 2026-08-22; letting the end default to now would declare inits with no source. 29 inits.
  - Shards are 1 init x all leads x all 101 members; `max_vars_per_job = 1` → ~29 x 35 ≈ 1000 jobs. `jobs_per_pod = 1` (ensemble), start `max_parallelism` ~20 and watch scheduling.
  - Rough sizing: ~49 GB/init uncompressed; expect a few hundred GB written and ~800 GB of GRIB read. The test measures the real numbers for Phase 2.
[...]

PR #947, opened 2026-08-25
Fix ECMWF 46-day pressure level axis order

mrshll:
Every `pressure_level` variable in the ECMWF 46-day store is being written entirely as NaN. Found during the Phase 1 test backfill (#932); Sentry `REFORMATTERS-HY` is the same bug.

## Cause

`read_data` stacked levels on `axis=-1`, returning `(latitude, longitude, pressure_level)`. The template declares

```
(init_time, lead_time, ensemble_member, pressure_level, latitude, longitude)
```

so the output slot is `(pressure_level, latitude, longitude)`. Assignment raised

```
could not broadcast input array from shape (121,240,10) into shape (10,121,240)
```

`_read_into_data_array` catches per-coord exceptions and marks the coord `ReadFailed`, so the pod exited 0 and the Kubernetes job reported no failures. The surface variables were unaffected, which is why the run looked healthy.

## Fix

- Stack on `axis=0`.
- Check the trailing two axes for the grid shape instead of the leading two. The old check passed for the wrong order; the new one covers both the single-level and multi-level branches and fails inside `read_data` rather than surviving to the assignment.

## Tests

Two gaps let this through, both closed here.

**The unit test asserted the bug.** `test_reads_a_pressure_level_message_into_its_output_level` expected `values.shape == (121, 240, len(levels))` — the shape the implementation produced, not the one the template declares — so it passed throughout. It now derives the expected axes from `DAILY_CONFIG.dims[group]`, the same declaration the write path keys on.

**Nothing exercised a vertical group end to end.** `forecast_46_day_dynamical_dataset_test.py` had no backfill test, unlike every comparable dataset. Added one that backfills a single initialization of one root and one pressure level variable and pins the global mean at each level:

```
1000  925   850   700    500     300     200     100     50      10  hPa
10.9  7.4   4.5  -2.8  -17.3   -41.8   -55.2   -65.7  -62.4   -49.7  degC
```

A tropospheric lapse profile through the 100 hPa tropopause, warming into the stratosphere above it — a shape that only comes out right if levels are on the axis the template declares.

Both tests verified red/green: restoring `axis=-1` fails the unit test with `Expected a (121, 240) grid, found (240, 10)` and turns the integration test's whole group to NaN, reproducing the production failure including the swallowed `Read failed` log.

One detail worth flagging for reviewers: the integration test leaves `init_time` at production chunk geometry and shrinks only the other dims. Shrinking the append dim caps it at two shards across the full declared 1153-init axis, so a single job reaches for initializations that predate the GRIB archive.

## Checks

`ruff format`, `ruff check`, `ty check` clean. `pytest tests/ecmwf`: 267 passed, slow tests included.

### DIFF
--- a/src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
+++ b/src/reformatters/ecmwf/ifs_ens/forecast_46_day_region_job.py
@@ -212,10 +212,10 @@ class EcmwfIfsEns46DayRegionJob(
             bands = iter(present)
             values = np.stack(
                 [missing if level is None else next(bands) for level in coord.levels],
-                axis=-1,
+                axis=0,
             )
-        assert values.shape[:2] == GRID_SHAPE, (
-            f"Expected a {GRID_SHAPE} grid, found {values.shape[:2]}"
+        assert values.shape[-2:] == GRID_SHAPE, (
+            f"Expected a {GRID_SHAPE} grid, found {values.shape[-2:]}"
         )
         mask_source_fill_value_inplace(values, data_var.internal_attrs)
         return values
--- a/tests/ecmwf/ifs_ens/forecast_46_day_dynamical_dataset_test.py
+++ b/tests/ecmwf/ifs_ens/forecast_46_day_dynamical_dataset_test.py
@@ -1,10 +1,15 @@
+import numpy as np
+import pandas as pd
 import pytest
+import xarray as xr
 
 from reformatters.common import validation
+from reformatters.common.types import DatetimeLike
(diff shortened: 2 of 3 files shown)


mrshll (commented):
No findings. I traced the returned array shape through `out.loc[coord.out_loc()]`; stacking pressure levels on axis 0 matches the template’s `(pressure_level, latitude, longitude)` slot, while the trailing-grid assertion covers both root and pressure-level reads. The focused 17-test region-job suite and the new end-to-end backfill test pass locally, as do ruff and ty on the changed files.

PR #952, opened 2026-08-26
Honor overwrite-chunks on virtual datasets

aldenks:
A virtual backfill decided its work by asking the store which source files it already had refs for, regardless of what the caller asked for. `--overwrite-chunks` reached the safety guards and the metadata write, but never the write loop:

```python
remaining = [
    coord
    for job in jobs
    for coord in job.filter_already_present(job.source_file_coords(), readonly_store)
]
```

`grep overwrite` over `virtual_region_job.py` returned only an unrelated comment.

## Why it matters

It makes a whole class of correction impossible. A variable whose refs all point at the **wrong source file** can never be repointed, because its own stale refs mark every file as already done. The rewrite is skipped, no commit is made, and **the job reports success.**

This is not hypothetical. #950 changed four HRRR analysis variables to source forecast hour 1 instead of hour 0, where HRRR writes them as constant-zero messages. Two `--overwrite-chunks` backfills ran to completion — 37/37 and 50/51 workers, zero failures — and changed nothing:

```
lightning_threat_1m                   2024-07-15T18:00  nonzero=0  (still empty)
aerosol_optical_thickness_atmosphere  2024-07-15T18:00  nonzero=0  (still empty)
lightning_threat_2m  (control, untouched)               nonzero=18576  max=6.478
```

The only way to make progress was to delete 417,372 refs by hand first, so that `filter_already_present` had nothing to find.

## The change

`process_worker_jobs` takes `overwrite_chunks` across the worker-processing seam, and the virtual variant uses every coord rather than the filtered set when it is set. `set_virtual_refs` already replaces a ref at a chunk index, so rewriting is exactly what the flag promises. The materialized variant accepts and ignores it — it has always rewritten its whole region.

Resuming is unchanged: without the flag, an interrupted backfill still skips what it already wrote, and an operational update still skips files it has ingested.

## Test

`test_overwrite_chunks_rewrites_already_present_refs` pins both halves: a plain re-run makes no new commit, and a re-run with `overwrite_chunks=True` writes the refs again and the values still read back correctly.

`uv run pytest -m "not slow"` 2118 passed; ruff and ty clean.

## Related

Two other sharp edges from the same session, not addressed here:

- `representative_var` can pick a variable that does not exist over the filtered range, killing a job mid-run — separate PR.
- `append_dim_end` is frozen at submission, so an operational update landing before worker 0's guard makes a job permanently unrecoverable with a "trimming an existing store" error. Every retry reuses the stale value.

### DIFF
--- a/src/reformatters/common/materialized_region_job.py
+++ b/src/reformatters/common/materialized_region_job.py
@@ -126,6 +126,8 @@ class MaterializedRegionJob(
         store_factory: storage.StoreFactory,
         branch_name: str,
         worker_index: int,
+        *,
+        overwrite_chunks: bool = False,  # noqa: ARG003 - always rewrites its region
     ) -> dict[str, list[SourceFileResult]]:
         """Write all of this worker's jobs to ``branch_name`` in a single commit."""
         # One commit per worker; an empty job set would make an empty icechunk
--- a/src/reformatters/common/materialized_region_job.py
+++ b/src/reformatters/common/materialized_region_job.py
@@ -127,9 +127,12 @@ class MaterializedRegionJob(
         branch_name: str,
         worker_index: int,
         *,
-        overwrite_chunks: bool,  # noqa: ARG003 - always rewrites its region
+        overwrite_chunks: bool,
     ) -> dict[str, list[SourceFileResult]]:
         """Write all of this worker's jobs to ``branch_name`` in a single commit."""
+        assert overwrite_chunks, (
+            "Materialized jobs rewrite every chunk of their region; see rewrites_whole_region"
+        )
         # One commit per worker; an empty job set would make an empty icechunk
         # commit (which raises), so the caller must filter empty workers out.
         assert worker_jobs, "process_worker_jobs requires at least one job"
--- a/src/reformatters/common/materialized_region_job.py
+++ b/src/reformatters/common/materialized_region_job.py
@@ -127,7 +127,7 @@ class MaterializedRegionJob(
         branch_name: str,
         worker_index: int,
         *,
-        overwrite_chunks: bool = False,  # noqa: ARG003 - always rewrites its region
+        overwrite_chunks: bool,  # noqa: ARG003 - always rewrites its region
     ) -> dict[str, list[SourceFileResult]]:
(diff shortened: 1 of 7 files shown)


src/reformatters/common/materialized_region_job.py
```diff
@@ -126,6 +126,8 @@ def process_worker_jobs(
         store_factory: storage.StoreFactory,
         branch_name: str,
         worker_index: int,
+        *,
+        overwrite_chunks: bool = False,  # noqa: ARG003 - always rewrites its region
```
aldenks: Seems like this should be a required arg? For materialized the logical default is True and it _must_ be True, for virtual its logically default false and true is supported? But the conflicting defaults makes me think we should have it be a required arg so no defaults come into play 

assert it is true in the func impl here for materialized region job

src/reformatters/common/virtual_region_job.py
```diff
@@ -313,11 +313,16 @@ def process_worker_jobs(
...
     ) -> dict[str, list[SourceFileResult]]:
         """Drive the whole worker's virtual write loop on ``branch_name``.
 
-        Gathers the not-already-present source files across all the worker's jobs
-        against a single readonly view, then runs one write loop over their union:
+        Gathers the source files across all the worker's jobs against a single readonly
+        view, then runs one write loop over their union. Files whose refs the store
+        already holds are skipped unless ``overwrite_chunks`` asks for a rewrite --
+        without that, a variable whose refs all point at the wrong source file can
+        never be corrected, because its own stale refs mark every file as done.
```
```suggestion
        already holds are skipped unless ``overwrite_chunks`` asks for a rewrite.
```

src/reformatters/common/region_job.py
```diff
@@ -458,10 +458,15 @@ def process_worker_jobs(
...
         worker_index: int,
+        *,
+        overwrite_chunks: bool = False,
     ) -> dict[str, list[SourceFileResult]]:
         """Process one worker's region jobs against ``branch_name`` and
         return the per-variable source file results.
 
+        ``overwrite_chunks`` rewrites data the store already holds instead of
+        resuming: a variant that skips already-written work must not skip it.
+
```
aldenks: Cut

PR #957, opened 2026-08-26
Keep cron job names within the kubernetes length limit

mrshll:
The deploy of #931 failed at `kubectl apply`: all four WeatherNext 2 cron job names exceed Kubernetes' 52-character cap for `CronJob.metadata.name`.

```
google-weathernext2-forecast-historical-virtual-update     54
google-weathernext2-forecast-historical-virtual-validate   56
google-weathernext2-forecast-operational-virtual-update    55
google-weathernext2-forecast-operational-virtual-validate  57
```

### Why it escaped tests

`CronJob.name` already carried `max_length=52`, but `ReformatCronJob` and `ValidationCronJob` each redeclared `name` with only a `pattern`. In pydantic a redeclared field **replaces** the parent's constraints rather than adding to them, so the length cap silently applied to no cron job at all. The first place it could fail was the cluster.

### Fix

- Share the constraint as `CronJobName` and nest it in both subclasses, so the cap holds and the `-update`/`-validate` patterns still apply.
- Add a `DynamicalDataset.cron_job_name_prefix` hook, defaulting to `dataset_id`. The two WeatherNext 2 datasets override it to drop the redundant `google-` prefix.
- Thread that prefix through `rename_cronjob_for_staging`, which previously assumed cron job names start with `dataset_id`.

`dataset_id`, store paths, and the published product ids in dynamical-org/meta#188 are unchanged. The other 52 cron jobs keep their existing names.

```
weathernext2-forecast-historical-virtual-update     47
weathernext2-forecast-historical-virtual-validate   49
weathernext2-forecast-operational-virtual-update    48
weathernext2-forecast-operational-virtual-validate  50
```

### Tests

No new test was needed for the length itself — restoring the constraint makes the existing `test_operational_kubernetes_resources_are_valid` fail on exactly the two offending datasets, since it constructs every registered dataset's cron jobs. Verified it goes red before the rename and green after.

`test_cronjob_names_are_consistent` now checks the prefix is a genuine tail of `dataset_id` and that each cron job still carries the full `dataset_id`, preserving its traceability intent. Added a staging case covering a prefix that differs from `dataset_id`.

Full suite: 2385 passed, 13 skipped.

Unblocks the historical backfill in dynamical-org/meta#188.

### DIFF
--- a/src/reformatters/google/weathernext2/forecast_historical_virtual/dynamical_dataset.py
+++ b/src/reformatters/google/weathernext2/forecast_historical_virtual/dynamical_dataset.py
@@ -44,9 +44,14 @@ class GoogleWeathernext2ForecastHistoricalVirtualDataset(
         )
     )
 
+    @property
+    def cron_job_name_prefix(self) -> str:
+        # With the "google-" prefix these names exceed the CronJobName limit.
+        return self.dataset_id.removeprefix("google-")
+
     def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
         update = ReformatCronJob(
-            name=f"{self.dataset_id}-update",
+            name=f"{self.cron_job_name_prefix}-update",
             schedule="0 0 * * *",
             pod_active_deadline=timedelta(hours=6),
             image=image_tag,
@@ -57,7 +62,7 @@ class GoogleWeathernext2ForecastHistoricalVirtualDataset(
             suspend=True,
         )
         validate = ValidationCronJob(
-            name=f"{self.dataset_id}-validate",
+            name=f"{self.cron_job_name_prefix}-validate",
             schedule="0 0 * * *",
             pod_active_deadline=timedelta(minutes=30),
             image=image_tag,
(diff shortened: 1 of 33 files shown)


src/reformatters/google/weathernext2/forecast_historical_virtual/dynamical_dataset.py
```diff
@@ -44,9 +44,14 @@ class GoogleWeathernext2ForecastHistoricalVirtualDataset(
...
 
+    @property
+    def cron_job_name_prefix(self) -> str:
+        # With the "google-" prefix these names exceed the CronJobName limit.
+        return self.dataset_id.removeprefix("google-")
+
     def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
         update = ReformatCronJob(
-            name=f"{self.dataset_id}-update",
+            name=f"{self.cron_job_name_prefix}-update",
```
aldenks, line 54: lets revert this entire PR :D and simply...

```
        cron_job_name_prefix = self.dataset_id.replace("weathernext2", "wn2")
        update = ReformatCronJob(
            name=f"{cron_job_name_prefix}-update",
```
and similar for validation cron job and in the operational's dynamical_dataset.

PR #968, opened 2026-08-27
Record two WeatherNext 2 variable quirks found while validating the historical archive

mrshll:
Validating `google-weathernext2-forecast-historical-virtual` turned up two intrinsic variable facts the metadata did not carry. Both were verified against the store rather than assumed, and both apply to the historical and operational datasets.

The validation tooling changes this work also needed are now split out into #976; this PR is metadata only.

### Specific humidity holds small negative values

0.09% to 0.45% of cells at a given time, down to -6e-4 kg/kg, present in every year and at both short and long leads. Total precipitation already documents the same class of raw-model artifact; specific humidity now says so too.

### Sea surface temperature is a daily field

Across three inits and two points it steps once per 24 hours and varies negligibly between steps — up to 2312x the mean step size at the daily boundary versus within it — so a reader looking for a sub-daily signal finds none.

This is not rounding: virtual chunks keep full float32 precision, and observed steps run down to 3e-5 degrees.

## Notes

- Both are additions or extensions to a `comment` attribute on a dataset that is in neither the production nor the staging catalog, so the backward-compatibility policy does not apply.
- `_pressure_var` gained a `comment` parameter, which it previously hard-coded to `None`.

## Testing

- `uv run main DATASET_ID update-template` re-run for both the historical and operational datasets; the only template changes are the two comments.
- `uv run pytest tests/common/common_template_config_subclasses_test.py tests/common/datasets_cf_compliance_test.py tests/google` — 412 passed.
- `uv run ruff format --check`, `uv run ruff check`, `uv run ty check` all clean.

https://claude.ai/code/session_01Nu85CtgWZjouCWCWvU32yK

### DIFF
--- a/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
+++ b/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
@@ -55,7 +55,7 @@
     "short_name": "sst",
     "standard_name": "sea_surface_temperature",
     "units": "degree_Celsius",
-    "comment": "NaN over land where sea surface temperature does not apply. The underlying field changes once per 24 hours.",
+    "comment": "NaN over land where sea surface temperature does not apply. Value changes once every 24 hours.",
     "step_type": "instant",
     "coordinates": "expected_forecast_length spatial_ref valid_time",
     "_FillValue": "AAAAAAAA+H8="
--- a/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
+++ b/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
@@ -55,7 +55,7 @@
     "short_name": "sst",
     "standard_name": "sea_surface_temperature",
     "units": "degree_Celsius",
-    "comment": "NaN over land where sea surface temperature does not apply.",
+    "comment": "NaN over land where sea surface temperature does not apply. The underlying field changes once per 24 hours.",
     "step_type": "instant",
     "coordinates": "expected_forecast_length spatial_ref valid_time",
     "_FillValue": "AAAAAAAA+H8="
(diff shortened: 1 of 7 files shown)


docs/validation.md
mrshll, line 1: i assume we want to drop this change.

src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
```diff
@@ -55,7 +55,7 @@
     "short_name": "sst",
     "standard_name": "sea_surface_temperature",
     "units": "degree_Celsius",
-    "comment": "NaN over land where sea surface temperature does not apply.",
+    "comment": "NaN over land where sea surface temperature does not apply. The underlying field is daily: values step once per 24 hours and vary negligibly between steps, so there is no sub-daily signal to read.",
```
aldenks: on both datasets.
```suggestion
    "comment": "NaN over land where sea surface temperature does not apply. The underlying field changes once per 24 hours.",
```

src/reformatters/google/weathernext2/forecast_virtual/template_config.py
```diff
@@ -638,6 +643,9 @@ def _pressure_data_vars(
             long_name="Specific humidity",
             units="1",
             standard_name="specific_humidity",
+            comment=(
+                "Small negative values are raw model artifacts; set values < 0 to zero."
```
aldenks:
```suggestion
                "Small negative values are raw model artifacts."
```

src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/sea_surface_temperature/zarr.json
mrshll: Applied, on both datasets — edited in `template_config.py` and re-ran `update-template` for the historical and operational stores.

---

src/reformatters/google/weathernext2/forecast_virtual/template_config.py
mrshll: Applied.

Note this leaves `specific_humidity` reading "Small negative values are raw model artifacts." while the pre-existing `total_precipitation_surface` comment still ends "...; set values < 0 to zero." Happy to trim that one to match if you'd like them consistent — I left it alone to avoid widening this PR.

---

docs/validation.md
mrshll: Dropped — all `docs/validation.md` changes are gone. This PR is now metadata only; the validation tooling changes moved to #976, which carries no doc changes either.

One consequence worth flagging: #976 adds two CLI flags (`--decode-samples`, `--value-samples`) that are now documented only in `--help`. Say the word if you'd like a minimal Options entry for each there.

---

PR #978, opened 2026-08-28
Add precipitation, 10 m wind, 2 m extremes and CAPE to ECMWF ER 46 day

mrshll:
Adds the six ECDS variables the 46-day daily dataset could always have carried, and
records the sequence needed to get them into the archive, the store, and validation.

## Why they were missing

`SINGLE_LEVEL_LEAD_TIMES` records each variable's *native* step axis. That record is
accurate for all 38 single-level variables — the error was reading it as the only steps
ECDS will serve. The 47 daily steps (0, 24, … 1104) are a strict subset of the 185
six-hourly steps, so a variable filed under "6-hourly" is fully retrievable at daily
steps. Verified against the live constraints endpoint and by retrieving and decoding
real messages: 274 GRIB messages across 7 retrievals, plus the 54-variable manifest
enumerated for `origin=ecmwf` at three initialization dates spanning the archive era.

Everything else in those tables checked out — `MAXIMUM_LEAD_HOUR`,
`CONTROL_ONLY_VARIABLES`, all 38 step classifications, all 6 pressure-level level sets
including specific humidity's 7-level restriction, the isentropic entry, and the
13-accumulation / 2-instantaneous split. One genuine blind spot remains out of scope
here: the source's `oceanic` level type (9 variables) is absent from `LevelType` and has
never been requested or inspected.

## Variables added

| Dataset variable | ECDS variable | How it is read |
|---|---|---|
| `precipitation_surface` | `total_precipitation` | Daily steps of a step-0 accumulation, deaccumulated to a 24 h mean rate |
| `wind_u_10m` / `wind_v_10m` | `10_m_u/v_component_of_wind` | Instantaneous at each daily step |
| `maximum_temperature_2m` | `maximum_2_m_temperature_in_the_last_6_hours` | **Max of the four 6 h windows tiling each step** |
| `minimum_temperature_2m` | `minimum_2_m_temperature_in_the_last_6_hours` | **Min of the four 6 h windows** |
| `average_convective_available_potential_energy_atmosphere` | `convective_available_potential_energy` | 24 h mean, as published |

The 2 m extremes are the only ones needing new machinery. Each source message covers the
6 hours *ending* at its step, so reading daily steps alone would return the 18Z–00Z
window and call it a daily extreme. `SubStepReduction` names the messages tiling a step
and how they combine; `_message_byte_ranges` fetches them in one ranged request and
`read_data` reduces them.

`FIELD_KEYS` already carried all six with correct `statistical_process` discriminators,
so the inventory and index need no change — confirmed on real blobs, including that the
index tells maximum from minimum where GDAL's element, description and unit are
identical for both.

Naming follows the repo's cross-dataset consistency test: `tmax`/`tmin` and
"Maximum/Minimum temperature" to match GFS and GEFS, with the 24-hour window stated in
`comment`. CAPE takes the `average_` prefix for the same reason `average_temperature_2m`
does — an instantaneous variable of that name exists in sibling datasets — and
`_atmosphere` because its GRIB surfaces are ground-to-TOA, like `total_column_water_atmosphere`.

## Template change is purely additive

Six new nodes; no pre-existing node or root attribute changed. Not a breaking change.

## The archive starts holding 6-hourly data — deliberate

Worth stating plainly, because it is a change in what the archive *is* rather than just
how large it is. Today it is 100% daily. Init 2026-08-10, listed from the bucket:
12 blobs, 22.26 GB, every one of them 47 daily steps or 46 daily means.

This PR takes it to 16 blobs and ~31.3 GB per init, of which **5.19 GB is sub-daily**:

| Added | Per init | Cadence |
|---|---|---|
| daily-mean group re-fetch (CAPE rename) | 3.88 GB | daily |
| `total_precipitation` + `10_m_u/v_component_of_wind` | 3.12 GB | **6-hourly, 185 steps** |
| `maximum/minimum_2_m_temperature_in_the_last_6_hours` | 2.07 GB | **6-hourly, 184 steps** |

`MATERIALIZED_PRODUCT_ECDS_VARIABLES` declares the 6-hourly and daily consumer
manifests separately. `ECDS_VARIABLES` is their union, so overlapping fields are
archived once. `request_shards.py` still controls the source lead-time axes because
product cadence alone is insufficient: daily max/min require four 6-hour source windows.

The 2 m extremes force it: the source publishes no daily extreme, only 6 hour windows, so
`maximum_temperature_2m` cannot exist without all 184 of them. Reading only the daily
steps would return the 18Z–00Z window mislabelled as a daily maximum.

Precipitation and 10 m wind do not force it — the daily dataset reads 47 of their 185
steps — but they stay at native resolution for two reasons. Routing them to daily steps
would merge them into the existing 15-variable daily-point group, changing its digest and
forcing a re-fetch of all 15 across every initialization, which costs more than the
0.79-vs-3.12 GB it saves unless new machinery keeps them in their own group. And
`plans/186`'s proposed 6-hourly dataset is exactly these five variables at exactly these
resolutions (3 at 185 steps, 2 at 184), with `forecast_46_day_6_hourly_1_5_degree` already
scaffolded as an empty directory — so archiving natively now means that dataset needs zero
further ECDS retrieval, where daily-only would mean a second multi-week pass through the
queue for the sake of ~$40/month of storage.

## Archive state and the restart

The archive is **197 initializations deep** — 2026-02-11 to 2026-08-26, contiguous, no
gaps — against a declared axis of 1,156. It is filled from both ends: the 06 UTC
`archive-grib-files` cron on prod Kubernetes extends the frontier, and a history walk on an
EC2 instance works backwards. Both write the same prefix and both derive blob names from
`ECDS_VARIABLES`.

Because 959 initializations have not been archived at all, restarting the history walk now
with the full variable set is much cheaper than adding the new variables later: every init
from here is retrieved once, and only the 197 already done need a top-up.

| | Per init | Total |
|---|---|---|
| 959 never archived, all 41 variables | 27.65 GB | **26.51 TB** |
| 197 already archived, new blobs only | 8.13 GB | **1.60 TB** |
| | | **28.11 TB to retrieve** |
| 14-var daily-mean blobs superseded by the CAPE rename | 2.74 GB | 0.54 TB to delete |

Finishing the walk with the old 35 variables was already 21.35 TB, so **the six new
variables add 6.77 TB** on top of work already committed.

`archive_initialization` lists the destination and skips any selection whose exact file name
is already there, so the walk can simply cover the whole axis newest-first: the 197 done
initializations cost only their three missing blobs, and nothing is re-retrieved by accident.

### The one hard ordering constraint

**Deploy this to prod Kubernetes before restarting the EC2 walk.** Both writers compute blob
names from `ECDS_VARIABLES`, and adding CAPE renames the daily-mean group
(`…-52703092` → `…-c21fee3e`). If the walk runs the new code while the cron still runs the
old, the two writers disagree and the archive ends up with the daily-mean group under two
different names depending on which process wrote each initialization. Same image, both
places, before the walk restarts.

CAPE joining its natural group is now the right call rather than something to engineer
around: at 197 archived initializations the rename costs 0.54 TB of re-fetch and 0.54 TB of
orphans, roughly 2% of the run. An earlier commit gave CAPE its own request to dodge that;
it has been dropped, because the saving does not justify a permanent structural exception
and the re-fetch produces exactly the blob an original 15-variable request would have.

## CAPE — included; the boundary is a dataset question, not an archive one

CAPE ships, and its retrieval cost is genuinely small: **0.77 TB, 2.7% of the 28.11 TB run**
(0.19 TB as one extra variable in requests already being made for the 959 unarchived
initializations, 0.58 TB re-pulling the renamed group for the 197 already archived), plus
0.54 TB of orphans to delete. No archiver mechanism is needed —
`archive_initialization` applies the same selections to every initialization.

What re-loading does *not* fix is that CAPE names two different physical quantities.
`plans/186` finding 6, which this PR missed on its first pass:

> **CAPE carries a silent definition change at 2024-11-12** (CAPE → MUCAPE, same ECDS name,
> identical constraints signature). … Either annotate in the variable `comment` plus a
> validation-report review note, or drop it — but do not add it silently.

At the CY49R1 upgrade ECMWF replaced CAPE with MUCAPE under an unchanged ECDS name and an
unchanged constraints signature, so no API check can see it — which is why the step-axis
stability checks here came back clean and told us nothing. Over the declared axis that is
**503 initializations on the old definition and 653 on the new**; all 197 currently archived
are after the boundary.

**This does not block the archive.** The archive is a faithful mirror of the source, so it
holds CAPE for every initialization the source offers, exactly as it holds everything else.
The decision is what the *dataset* exposes, and it is due before the store backfill in
stage 3, not before the walk restarts:

1. **Expose the whole era** with the boundary as a validation-report review note. Per
   AGENTS.md a version-boundary change belongs in review notes rather than a static
   `comment`, which would go stale. Leaves one variable meaning two things.
2. **Expose only from 2024-11-12**, named for MUCAPE — one physical definition throughout.
   Needs an `available_from`-style mechanism; GEFS has one at `gefs_config_models.py:36`
   but it is provider-specific and would need adding here.

Either way the metadata this PR currently carries is wrong for the majority of the archive:
MUCAPE is a most-unstable-parcel quantity, so `standard_name =
"atmosphere_convective_available_potential_energy"` and `long_name = "Convective available
potential energy"` describe only the pre-2024-11-12 half. That needs settling with the
option above.

The CAPE→MUCAPE change is taken from `plans/186`'s reading of the cycle notes; I have not
re-verified it against ECMWF documentation or the data. It is testable — MUCAPE should sit
systematically above surface-parcel CAPE, so a distribution shift at the boundary would
corroborate it, and the archive will hold both sides once the walk crosses 2024-11-12.

## Operational

- **Update cron** worker count is `num_variable_groups()`, so it rises from 35 to 41
  automatically. `max_vars_per_job=1` keeps each job to one variable's array; the two extremes
  read four messages per slot instead of one, same output array.
- **The frontier cron's run grows ~22.3 → ~27.7 GB per initialization** inside the same 06 UTC
  start and 6 hour deadline. Time the first few runs.
- **Validate cron** will fail on the new variables until the store backfill lands — expected,
  the known add-a-variable failure mode.
- **Store backfill comes last**, once the archive is complete, and covers all 41 variables in
  one pass. The partially written store (166 initializations) is superseded by it.

## Validation to redo

Beyond the standard per-variable pass, three things are specific to this change:

- **The extremes are the new failure mode.** Confirm `minimum_temperature_2m` ≤
  `average_temperature_2m` ≤ `maximum_temperature_2m` everywhere. A reduction bug that
  silently returned a single window would still look physically plausible on its own plot and
  only shows up against the mean. Pinned as a test here, worth re-checking on the store.
- **Precipitation against convective.** `precipitation_surface` ≥
  `precipitation_convective_surface` at every cell, since convective is a component of the
  total. Cheap, and catches a scaling or deaccumulation error.
- **10 m wind is a 00Z snapshot, not a daily statistic.** Every other root variable here is a
  24-hour statistic; add a review note so a reader does not compare it against the means as if
  it were one.

## Verification evidence

- ECDS constraints for `origin=ecmwf`: 54 variables across four level types, per-variable step
  axes, at 2023-06-28, 2024-06-28 and 2026-08-24 (identical at all three). Structural only —
  it cannot see the CAPE definition change.
- Real messages decoded for statistic and window via the GRIB2 product definition templates:
  accumulation from step 0 for precipitation; 6 h max/min windows ending at the step; the daily
  means as an average of four 6-hourly samples; instantaneous 10 m wind and pressure-level u/v/w.
- Archive state and layout read from the bucket: 197 initializations, 12 blobs each, 35
  variables, message order lead → member → variable (alphabetical), confirmed on a real
  64,400-message index.
- A message retrieved alone is byte-identical (same SHA-256) to the same message retrieved in
  a multi-variable request, so blob composition does not affect message bytes.
- Test fixture extended with 13 real control messages for 2026-08-10T00Z, so every new
  variable's read path is covered by real data. The daily extremes are asserted equal to the
  pixel-wise max/min of their four windows.


src/reformatters/ecmwf/ifs_ens/forecast_46_day_1_5_degree/template_config.py
```diff
@@ -506,6 +570,52 @@ def data_vars(self) -> Sequence[EcmwfIfsEns46DayDataVar]:
...
+                    comment="Minimum temperature at 2 metres over the previous 24 hours.",
+                ),
+                internal_attrs=EcmwfIfsEns46DayInternalAttrs(
+                    ecds_variable="minimum_2_m_temperature_in_the_last_6_hours",
+                    grib_element="TMP",
+                    grib_description='2[m] HTGL="Specified height level above ground"',
+                    grib_unit="[K]",
+                    keep_mantissa_bits=7,
+                    add_offset=-273.15,
+                    sub_step_reduction=SubStepReduction(
```
mrshll: is this the correct usage of the SubStepReduction @aldenks

PR #984, opened 2026-08-29
Move WeatherNext 2 variable quirks into comment attrs

mrshll:
Validating `google-weathernext2-forecast-operational-virtual` surfaced two variable quirks that had been living in the historical dataset's validation report review notes. Both are intrinsic, always-true facts about the variables, so per the comment-vs-review-note rule in AGENTS.md they belong in each variable's `comment` attr, where they travel with the data instead of going stale in a report.

## Changes

- **`pressure_level/specific_humidity` gains a `comment`**: "Small negative values are raw model artifacts; set values < 0 to zero." Wording matches the existing sentence on `total_precipitation_surface`, which has the same artifact.
- **`sea_surface_temperature`'s `comment` gains a sentence**: "Value changes once every 24 hours." — it is a daily field, clearly visible as 24-hour steps in the validation report's time series. Its existing land-NaN sentence is unchanged.
- **`_pressure_var` gains the `comment` parameter** `_root_var` already had; it previously hard-coded `comment=None`, so no pressure-level variable could carry one.

Templates regenerated for both WeatherNext 2 datasets, which share this template config. The only attribute diffs are the two comments above.

## Validation

- `uv run ruff format && uv run ruff check --fix && uv run ty check` — all pass
- `uv run pytest tests/common/common_template_config_subclasses_test.py tests/common/datasets_cf_compliance_test.py` — 391 passed

Both quirks were confirmed against real data in this validation run rather than carried over on trust: specific humidity reaches −7.6e-4 in the sampled 500 hPa field, and sea surface temperature's point time series steps in flat 24-hour segments.

The next operational update will push these attrs to the store. The validation report already reflects the new wording, so it runs a few hours ahead of the store until then.

Draft validation report: https://dataset-validation-reports.dynamical.org/google-weathernext2-forecast-operational-virtual/drafts/v0.1.0_2026-08-28T21-28/validation_report.html

https://claude.ai/code/session_01VLiGAuwZuzVoD9SNoGxD1K

---
_Generated by [Claude Code](https://claude.ai/code/session_01VLiGAuwZuzVoD9SNoGxD1K)_

### DIFF
--- a/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/pressure_level/specific_humidity/zarr.json
+++ b/src/reformatters/google/weathernext2/forecast_historical_virtual/templates/latest.zarr/pressure_level/specific_humidity/zarr.json
@@ -64,6 +64,7 @@
     "short_name": "q",
     "standard_name": "specific_humidity",
     "units": "1",
+    "comment": "Small negative values are raw model artifacts; set values < 0 to zero.",
     "step_type": "instant",
     "coordinates": "expected_forecast_length spatial_ref valid_time",
     "_FillValue": "AAAAAAAA+H8="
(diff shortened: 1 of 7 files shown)


mrshll:
Superseded by #968, which is merged and carries the same two comment attrs on both the historical and operational templates:

```
sea_surface_temperature            NaN over land where sea surface temperature does not apply.
                                   Value changes once every 24 hours.
pressure_level/specific_humidity   Small negative values are raw model artifacts; set values < 0 to zero.
```

#968 also went through review, so it drops the `docs/validation.md` changes (moved to #976) that both PRs originally carried.

PR #989, opened 2026-09-01
Fix new integration naming convention

aldenks:
## Summary

New integration names now mirror the dataset_id policy:

- Materialized datasets add no kind word to module or class names.
- Virtual datasets append _virtual to the module and Virtual to class names; callers pass the base variant and --kind virtual adds the suffix.
- Variants already ending in _virtual are rejected with a kind-specific error instead of producing a misleading or doubled suffix.
- Every concrete DynamicalDataset subclass uses the XxxxDataset suffix, including the ECCC and teaching classes.
- Teaching names mirror generated names: Example* for materialized integrations and ExampleVirtual* for virtual integrations.

| Example | Before | After |
| --- | --- | --- |
| Materialized noaa-gfs-forecast | src/reformatters/noaa/gfs/forecast/; NoaaGfsForecastTemporalDynamicalDataset | src/reformatters/noaa/gfs/forecast/; NoaaGfsForecastDataset |
| Virtual noaa-gfs-forecast-virtual | src/reformatters/noaa/gfs/forecast/; NoaaGfsForecastSpatialDynamicalDataset | src/reformatters/noaa/gfs/forecast_virtual/; NoaaGfsForecastVirtualDataset |

No shipped dataset uses the old Temporal or Spatial class-name infix. Concrete and teaching DynamicalDataset subclasses now consistently use the Dataset suffix rather than DynamicalDataset.

## Validation

- uv run ruff format
- uv run ruff check --fix
- uv run ty check
- uv run pytest -m "not slow" -n 4 (2256 passed)

### DIFF
--- a/src/reformatters/common/initialize_new_integration.py
+++ b/src/reformatters/common/initialize_new_integration.py
@@ -49,11 +49,11 @@ def initialize_new_integration(
     provider = _sanitize_identifier(provider)
     model = _sanitize_identifier(model)
     variant = _sanitize_identifier(variant)
-    if variant == "virtual" or variant.endswith("_virtual"):
+    if "virtual" in variant:
         message = (
-            "variant must omit the 'virtual' suffix when --kind virtual is used"
+            "variant must omit 'virtual' when --kind virtual is used"
             if kind is DatasetKind.virtual
-            else "variant cannot use the 'virtual' suffix with --kind materialized"
+            else "variant cannot use 'virtual' with --kind materialized"
         )
         raise typer.BadParameter(
             message,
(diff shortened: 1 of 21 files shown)


src/reformatters/common/initialize_new_integration.py
```diff
@@ -57,14 +49,26 @@ def initialize_new_integration(
     provider = _sanitize_identifier(provider)
     model = _sanitize_identifier(model)
     variant = _sanitize_identifier(variant)
+    if variant == "virtual" or variant.endswith("_virtual"):
```
aldenks: if "virtual" in variant (this will catch any out of order drops of virtual in the variant too)

Issue #825, opened 2026-07-31
GEFS virtual (map-optimized) datasets: 10 day, 16 day, 35 day, analysis


mrshll:
Tracks the **GEFS virtual (map-optimized) datasets** — four datasets, all starting 2020-09-23T12 (the v12 current-archive start), sourced from `s3://noaa-gefs-pds`.

| `dataset_id` | `name` | grid | leads | inits |
| --- | --- | --- | --- | --- |
| `noaa-gefs-forecast-10-day-0-25-degree-virtual` | NOAA GEFS forecast, 10 day, 0.25 degree, virtual | 0.25° | 0–240 h / 3 h | 6 h |
| `noaa-gefs-forecast-16-day-0-5-degree-virtual` | NOAA GEFS forecast, 16 day, 0.5 degree, virtual | 0.5° | 0–384 h | 6 h |
| `noaa-gefs-forecast-35-day-0-5-degree-virtual` | NOAA GEFS forecast, 35 day, 0.5 degree, virtual | 0.5° | 0–840 h | 24 h (00z) |
| `noaa-gefs-analysis-0-25-degree-virtual` | NOAA GEFS analysis, 0.25 degree, virtual | 0.25° | `time`, 3 h | — |

The forecasts carry `ensemble_member` (31); the analysis is control-member only. The 0.5° datasets carry a `pressure_level` group; the 0.25° ones are root-only.

- [ ] `noaa-gefs-analysis-0-25-degree-virtual`
- [ ] `noaa-gefs-forecast-10-day-0-25-degree-virtual`
- [ ] `noaa-gefs-forecast-16-day-0-5-degree-virtual`
- [ ] `noaa-gefs-forecast-35-day-0-5-degree-virtual`

**Design decisions and verified source facts are in the most recent comment.** The original per-dataset source explorations (2026-07-20 / 2026-07-31) are preserved as comments; #826 and #762 were folded into this issue.

mrshll:
Scope decision: where the source shifts resolution mid-archive, split into two virtual datasets at native resolutions rather than interpolating or dropping one regime — here that means a 0.25° 0–240h dataset (graduating the existing 10-day spatial work) plus a 0.5° 0–840h dataset. Also: the `b` files' scale (~6×10⁹ refs) is a design problem to solve (manifest split design), not grounds for exclusion.

aldenks:
## Source exploration — GEFS 35-day forecast (2026-07-31)

[...]

Issue #823, opened 2026-07-31
GFS virtual (map-optimized) datasets: forecast + analysis


mrshll:
Tracks the **GFS virtual (map-optimized) datasets** — two datasets, both 0.25°, sourced from `pgrb2.0p25` + `pgrb2b.0p25` on `s3://noaa-gfs-bdp-pds`.

| `dataset_id` | `name` | shape |
| --- | --- | --- |
| `noaa-gfs-forecast-virtual` | NOAA GFS forecast, virtual | `init_time` × `lead_time` × `latitude` × `longitude`, 0–384 h, 6 h inits |
| `noaa-gfs-analysis-virtual` | NOAA GFS analysis, virtual | `time` × `latitude` × `longitude`, hourly |

Both carry a root group plus a `pressure_level` group, and start at 2021-05-01 to match their materialized twins.

- [ ] `noaa-gfs-forecast-virtual`
- [ ] `noaa-gfs-analysis-virtual`

**Design decisions and verified source facts are in the most recent comment.** The original per-dataset source explorations (2026-07-31) are preserved as comments; #824 was folded into this issue.

aldenks:
## Source exploration — GFS forecast (2026-07-31)

_Preserved from this issue's original body; the body is now a family tracker._

**Verdict**: NEEDS JUDGMENT — mechanically a near-copy of `noaa-gefs-forecast-10-day-spatial-dev` (identical grid, bucket family, index parser, codec), but three real decisions: variable-inventory scope (25 → ~216 arrays), sub-1 hPa *float* pressure levels break the shared int-cast level lookup, and ~10⁹ refs puts manifest sizing at the top of the documented envelope.

**Proposed dataset**: `dataset_id="noaa-gfs-forecast-virtual"`, `name="NOAA GFS forecast, virtual"` (materialized: `noaa-gfs-forecast` / "NOAA GFS forecast").

**Source**

- `s3://noaa-gfs-bdp-pds/gfs.YYYYMMDD/HH/atmos/gfs.tHHz.pgrb2.0p25.fLLL` (+ `.idx` sidecar), us-east-1, anonymous, obstore-listable — same shape as the HRRR/GEFS virtual datasets.
[...]

PR #990, opened 2026-09-01
Extract the shared NOAA virtual GRIB-index region job

aldenks:
This PR ships **no dataset**. It contains three distinct kinds of change, listed separately below so they can be judged separately:

1. **An extraction** — shared NOAA virtual machinery lifted out of HRRR (behavior-preserving).
2. **Two fixes to pre-existing behavior** in that shared code (each a proven no-op for HRRR).
3. **A rename** in `noaa_grib_index.py` (no behavior change).

---

# 1. Extraction

Three NOAA virtual dataset suites (HRRR, and the GFS and GEFS suites now being built) need the same NODD-S3 discovery and GRIB-index → virtual-chunk-reference machinery, so this lifts it out of `noaa/hrrr/virtual_region_job.py` into a new shared `noaa/noaa_virtual_region_job.py`. `NoaaVirtualSourceFileCoord[DATA_VAR]` carries the lead time, the variables a file packs and the `.idx` sidecar URL; `NoaaVirtualRegionJob[DATA_VAR, COORD]` implements `discover_available` (obstore listing, index required) and `file_refs` (parse the idx, match each message by `(element, level, window)`, derive its byte range from the next message's start) against two class attributes a consumer supplies — `source_location_prefix` and `source_bucket_region`. HRRR keeps what is genuinely HRRR: its bucket constants, `hrrr_virtual_chunk_containers()`, the two truncated uploads it skips, the https→`s3://` URL rewrite, and the forecast/analysis subclasses.

The one deliberate semantic change in the extraction is that `_message_lookup` no longer coerces vertical level labels with `int()`. GFS pressure levels are float64 and GEFS's are int64 hPa, and the label must reach `pandas.Index.get_indexer` as the variable's own coordinate holds it. The `grib_index_level` format strings gain `:g` (`"{level} mb"` → `"{level:g} mb"`) so an integer and a float coordinate both render the way source indexes spell levels (`1000`, not `1000.0`; `0.01` stays `0.01`).

**Evidence HRRR's output is unaffected.** All three HRRR virtual stores publish `pressure_level` and `model_level` as **int64** (1000→50 mb and 1→50 hybrid) — confirmed in the checked-in templates and, independently during review, by opening the three live stores through the production STAC catalog. `pandas.Index.__iter__` already boxes an int64 index to Python `int`, so the removed `int()` was a no-op on those coordinates to begin with, and `format(1000, "g") == "1000"` makes `:g` a no-op on every level either index holds. An exhaustive replay of the old and new `_message_lookup` over all three HRRR virtual template configs × all data vars × leads 0–48 × every level produced identical lookup keys and identical emitted labels (91,967 entries per config). `grib_index_level` is an internal attr and is not serialized into `templates/latest.zarr`, so there is no template change and no regeneration.

---

# 2. Two fixes to pre-existing behavior

Both were surfaced by review as pre-existing warts and are fixed here rather than in a dataset PR: they live in the shared file this PR exists to review, GFS hits the first on essentially every source file, and deferring them would move HRRR-touching risk into a PR whose subject is a different dataset.

### 2a. Duplicate index messages are now first-wins

**What changed.** When two index lines both match the same variable *and* the same output cell, only the first is turned into a ref. Previously both were emitted at the same chunk index and the second silently overwrote the first — not numerically wrong, since such messages carry identical values, but nondeterministic-looking and inconsistent with the materialized path, which takes `matches[0]`.

The dedupe key is `(variable, out_loc)`, **not** the index key. One message legitimately filling several variables is a different case and is untouched: at GFS leads ≤ 6 the 6-hour-bucket variable and the run-total variable match the same `(element, level, window)` key and both must get a ref. `test_file_refs_duplicate_messages_keep_the_first_and_still_fill_every_variable` pins exactly that combination — a doubled line whose key maps to two variables yields one ref per variable, both at the first line's byte range.

**Why it matters now.** GFS publishes two byte-distinct but bitwise-identical `APCP`/`ACPCP` messages at every lead ≤ 6, so this fires constantly for the next consumer.

**Evidence HRRR is unaffected.** Across **102 real HRRR `.idx` files** pulled from the archive (2016, 2018, 2021, 2023 and 2026 inits × `sfc`/`prs`/`nat` × leads 0–48), comprising **62,856 index lines** and **51,945 matched refs**, the number of files containing any `(variable, out_loc)` matched by more than one index line was **0**. HRRR never takes this branch.

### 2b. The stale-byte-range guard now checks every message

**What changed.** The guard that rejects a file whose index byte ranges fall outside the data file used to run only on *matched* messages, inside the `if not matches` branch. Its consequence has always been file-level (`return []` discards the whole file), so the detection scope now matches the consequence's scope: every message's range is checked, matched or not. A stale index whose corruption happens to land on a message no variable requested used to sail through and contribute refs from the rest of the file.

**Evidence HRRR is unaffected.** Over the same 102 real index files and 62,856 lines, the number of files containing any line the hoisted guard rejects was **0**. No HRRR file that is ingested today would be dropped.

Both fixes are pinned by tests that fail against the pre-fix source and pass after it (verified by stashing only the source change: 2 failed, 15 passed → 17 passed).

---

# 3. Rename: `_lead_time_str` → `grib_index_window_str`

Pure rename in `noaa/noaa_grib_index.py`, no behavior change — the function body is untouched. It renders the idx *window* field, not a lead time, and this PR adds the third cross-module consumer of what was an underscore-private name. Every reference across `src/`, `tests/` and docs is updated; `grep -rn _lead_time_str` returns zero hits.

Flagging it because a concurrent GEFS effort is looking at this same function for a real behavioral reason — if that lands first, this rename rebases on top of it.

---

# Reviews

Two independent reviews (a Claude sub-agent and codex `$review-agent`) ran against the extraction diff before any fixes; findings were synthesized into one ranked list. The two fixes in section 2 came out of that synthesis.

**Raised by both, taken.** `NoaaVirtualSourceFileCoord.data_vars` was typed `Sequence[DataVar[NoaaInternalAttrs]]`, decoupled from the DataVar type parameter. `isinstance(hrrr_var, DataVar[NoaaInternalAttrs])` is `False` for every provider DataVar, so a consumer that inherits the field instead of redeclaring it silently gets a pydantic-revalidated deep copy of every DataVar per coord — the shape of a past 15 GB → 3.7 GB memory regression. The coord is now generic (`NoaaVirtualSourceFileCoord[NOAA_DATA_VAR]`), `_message_lookup` is typed on the same variable, and a test pins that a coord parameterized on a non-HRRR DataVar type holds the caller's objects (`is`), not copies.

**Raised by codex, taken.** The float64-level test proved nothing about the generic plumbing because it used HRRR's DataVar, coord and job throughout. It now runs a distinct `OtherDataVar` / `OtherSourceFileCoord` / `OtherVirtualRegionJob` triple — the actual second-consumer shape — over a float64 `pressure_level` coordinate including a fractional level.

**Raised by Claude, taken.** (a) The source-generic `file_refs` tests (byte ranges, one-message-fills-two-variables, element alternatives, `anl` window, and the skip paths) still lived in the HRRR 48-hour dataset's test file; they moved to `tests/noaa/noaa_virtual_region_job_test.py` per the repo's `<module>_test.py` convention, leaving the HRRR-specific ones behind. (b) A docstring claimed refs carry the vertical index's dtype — false, pandas boxes to Python scalars — deleted. (c) `NoaaHrrrVirtualRegionJob`'s docstring still described ref building it no longer does. (d) A test compared `{level:g}` against a `{level}` form that exists nowhere in the repo; it now pins the rendered strings themselves. (e) HRRR variable names trimmed out of a provider-generic docstring. (f) The `{level:g}` convention documented in `docs/virtual_datasets.md`.

**Declined.** Hoisting `hrrr_virtual_chunk_containers()` and the https→`s3://` rewrite into the shared module — the per-dataset containers function is an established repo-wide pattern and the second consumer does not exist yet to shape the abstraction. An assert bounding `:g` at 6 significant digits — no NOAA vertical level comes within three orders of magnitude of that cliff. Renaming HRRR's materialized `get_idx_url` to match the base `get_index_url` — real, but it touches the materialized paths and belongs in its own change.

---

# Checks

`uv run ruff format`, `uv run ruff check`, `uv run ty check` all clean. `uv run pytest tests/noaa/hrrr tests/common tests/noaa/noaa_virtual_region_job_test.py tests/noaa/noaa_grib_index_test.py` → **1321 passed**.

Part of #823 (and shared with the GEFS virtual datasets, #825).


AGENTS.md
```diff
@@ -57,7 +57,7 @@ deploy/                      # Docker and kubernetes configs
 └── aws/                     # nodepool.yaml, create_new_aws_open_data_bucket.sh
 ```
 
-- **Shared provider utilities** Check `src/reformatters/<provider>/` for shared modules (e.g., `ecmwf/ecmwf_grib_index.py`, `noaa/noaa_utils.py`).
+- **Shared provider utilities** Check `src/reformatters/<provider>/` for shared modules (e.g., `ecmwf/ecmwf_grib_index.py`, `noaa/noaa_utils.py`, `noaa/noaa_virtual_region_job.py`).
```
aldenks: Undo

PR #991, opened 2026-09-01
Add the shared GEFS virtual base and noaa-gefs-analysis-0-25-degree-virtual

aldenks:
First of the four datasets in #825, plus the shared base the other three will hang off.

## What this adds

**Shared base**, mirroring how the HRRR virtual family is factored:
`noaa/gefs/virtual_template_config.py` (`NoaaGefsVirtualTemplateConfig`) and
`noaa/gefs/virtual_region_job.py` (`NoaaGefsVirtualRegionJob`,
`NoaaGefsVirtualSourceFileCoord`). `source_file_types: frozenset[GEFSSourceFileType]`
selects both the grid and the variable catalog: the resolution falls out of the existing
`FILE_RESOLUTIONS` map, and `get_shared_template_dimension_coordinates` /
`get_shared_coordinate_configs` now take a resolution so the 0.5° datasets reuse them
unchanged (at 0.25° they reduce to the exact prior expression, so no materialized
template changed). Only the `s` catalog is populated; `a`/`b` lands with PR 3.

**Dataset** `noaa-gefs-analysis-0-25-degree-virtual` — control member only, 3-hourly
`time` from 2020-09-23T12, 38 variables. A separate first commit renames
`GEFSDataVar`/`GEFSInternalAttrs` to `NoaaGefs*` per the repo's naming policy.

## Source verification — measured, not inherited

Read literal `.idx` inventories from `s3://noaa-gefs-pds` at 2020-09-23T12,
2021-07-20T06/T12, 2022-10-18T06/T12, 2024-06-01, 2025-01-01 and 2026-08-28, across
f000/f003/f006/f009/f012/f015/f024/f120/f240, and decoded messages with gribberish.

- **Windows**: GEFS `s` uses a 6h-resetting bucket at 3h lead granularity — `0-3`, `0-6`,
  `6-9`, `6-12`, `12-15`, `18-24`, `114-120`, `234-240`. `_lead_time_str` renders
  **488/488** observed (element, lead) window strings exactly across three eras. **No
  change to the frozen `noaa_grib_index.py`.** #825's alternating 3h/6h claim is
  **verified** for the analysis's leads rather than left as inference.
- **No run-total sibling.** Every lead carries exactly 38 messages; there is no
  `0-N day acc fcst` beside the bucket. Differs from GFS — PR 2 needs no run-total variant.
- **#825 is wrong about MSLET.** The four late-arriving variables do *not* share one
  date. MSLET entered at **2021-07-20T12** (bisected: absent at T06, present at T12);
  only CPOFP, HGT@cloud-ceiling and VIS entered at `GEFS_B22_TRANSITION_DATE`. Counts
  go 34 → 35 → 38. Checked against idx, not against the issue.
- **`CLWMR`/`CLMR` does not apply** — neither spelling appears in any `s` file.
- **TSOIL is Kelvin** (216–321 K) despite GDAL labelling it `[C]`; converted explicitly.
- All 38 variables decode through the real read path with plausible values and units.

## Deliberate deviation from "zero source messages dropped"

#825 says the GEFS design drops no source messages. That is **deliberately false here**:
`HGT@surface` is excluded. The `s` file publishes it only at lead 0, unlike every other
message in it, so serving it means a mostly-NaN array plus a lead-0-only code path for
one variable. `pgrb2b` publishes the same quantity at ordinary leads, so the 16-day and
35-day datasets will carry it plainly at 0.5°. The 38-message inventory and the 38
variables match only by coincidence: 38 messages at lead ≥3, minus HGT@surface.

## Raw semantics, not materialized ones

APCP is served as the bucket accumulation under `total_precipitation_surface`, never
reusing `precipitation_surface` (a rate in the materialized product). TMAX/TMIN take
lead 3 or 6 rather than the source's degenerate zero-length `0-0 day max fcst` at lead 0.

## Manifest split

`split_size = 4 * 365 * 8 = 11680` (four years of 3-hourly steps); each array holds one
ref per step, so a split is also its manifest's ref count. **Measured 19.67 bytes/ref on
this dataset's own manifests** (1200 refs in one array, via `list_manifest_files`) — not
HRRR's 16.4, which the docs explicitly say not to carry over. A full manifest is
11680 × 19.67 = **224 KiB**: above icechunk's 1000-ref compression floor, far inside the
3 MiB reader budget, ~0.85 s of commit CPU across all 38 arrays. 76 manifests today.

## Comment claims that were executed, not proofread

Each names the wrong world it rules out.

| Claim | Wrong world ruled out | Result |
|---|---|---|
| `soil_temperature_0_10cm` / `volumetric_soil_moisture_0_10cm` "NaN over water, where there is no soil" | "the mask is a soil-model domain, not water" — open ocean cannot tell these apart | **Holds.** All four large lakes (Caspian, Superior, Victoria, Baikal) are NaN, while glaciated land (Greenland 245.3 K, Antarctica 235.2 K) keeps values — a soil-model domain would plausibly have dropped permanent ice. Area-weighted NaN 71.2% vs Earth water ~71%. Mechanism is a **GRIB bitmap → physical NaN**, no sentinel (`fill_value` already NaN), so per AGENTS.md the comment describes the NaN and adds no masking instruction. |
| `snow_water_equivalent_surface` / `snow_thickness_surface` "NaN over **open** water" | "'open' is doing no work and the mask is just water" | **Holds.** WEASD NaN is a strict subset of TSOIL NaN; 135,457 cells are TSOIL-NaN but carry WEASD (sea ice: Arctic 88N = 164, Weddell = 84). "water" would have been wrong. |
| `total_precipitation_surface` window + differencing clause | "the pair straddling a cycle change shares a start" | **Holds.** Adjacent pairs sharing a window start are exactly (03,06) (09,12) (15,18) (21,00). The earlier boundary-crossing framing was wrong; enumeration caught it. |
| Window start computed in code (`init_time`) | "off-by-one at 00/06/12/18 — attributed to the window it opens, not the one it closes" | **Holds.** Over a full day, `init_time` == most recent boundary **strictly before** T for every valid time. |
| `"avg"` sentence × 12 variables (a template, checked once against its least-typical members) | "right for 11, wrong for the 12th" | **Found the error.** The four categoricals are `statistical_process='average'` in the source but decode to strictly {0,1}; the averaging sentence contradicted `flag_values`. Now they carry only `0=no; 1=yes`, matching the published twin. The remaining 8 are all genuinely `average`. |
| `geopotential_height_cloud_ceiling` "Mask values above 19,000m" | "it is a single sentinel needing a `fill_value`" | **Holds.** 2,435 distinct values above 19,000; max 20000.22; no NaN. |
| `percent_frozen_precipitation_surface` "Mask values < -0.1" | — | **Nuance, kept deliberately.** Negatives run continuously from −50.0 to −0.0000061, so −0.1 leaves a few near-zero negatives. Wording is verbatim from the published materialized twin, which AGENTS.md says to match. |
| `available_from` dates | "the exploration write-up recorded the wrong date" — only idx can catch this | **Found the error** (MSLET, above). |

## Reviews

Three independent reviews: codex (`$review-agent`) and two Claude passes. **No P0/P1 from any.**

**Taken:**
- *(all three)* `representative_var` was duplicated, with the shared base holding the inferior copy — and the leaf's `assert` broke the documented `--filter-variable-names <late-added-var>` workflow. Now one implementation in the base: prefer era-stable, prefer instant, fall back rather than assert.
- *(codex + Claude, independently)* `file_refs` silently dropped a requested variable absent from the index. Now asserts. Paired with the `available_from` fix below, "every requested variable is in the file" is a real invariant rather than a coincidence.
- *(mine)* `available_from` judged against `init_time`, not valid time.
- *(Claude)* Shared base ignored `grib_element_alternatives` — inert today, silent data loss for PR 3's `CLWMR`/`CLMR`.
- *(Claude)* Categorical window comment (above).
- *(Claude)* `_WINDOW_COMMENTS` hard-coded analysis-specific UTC hours inside the shared base; moved to the dataset that knows its time axis.
- *(codex + Claude)* Snow scaling was not load-bearing — every asserted value was 0.0 or NaN, both fixed points of ÷1000. Now asserts 0.146 m on the Greenland ice sheet.
- *(codex + Claude)* 33 of 38 variables were exercised only against config-derived expectations. Added 18 real archived `.idx` fixtures spanning both inventory transitions.
- *(Claude)* bytes/ref was imported, not measured; base-class docstring described `a`/`b` behaviour it does not implement; TSOIL comment's justification clause was factually wrong (ECMWF datasets *do* ship soil temperature in Celsius); one comment narrated the verification act.

**Declined:**
- *(codex P2)* "`discover_available` should advance only through consecutive complete positions." Declined: `max(authorized)` is a deliberate trade — the alternative reintroduces the permanent block that the "a cycle the archive never published must not block the file beside it forever" clause exists to prevent, and `hrrr/analysis_virtual` makes the identical trade. Changing it here alone would leave the virtual family with two gate semantics. The second Claude review independently reached the same conclusion and traced the scenarios without finding a livelock. **Worth raising as a family-wide question, not a PR-1 fix.**
- *(Claude)* `resolution.strip('0')` is fragile for a hypothetical `1p00`. Declined: `FILE_RESOLUTIONS` is a closed 3-entry map and the idiom matches the existing `GefsSourceFileCoord.get_url`; diverging here would leave two idioms.

## Mutation checks

Each test was verified to have power by breaking the source and confirming a *named*
test failed, then restoring (verified byte-identical):

| Mutation | Caught by |
|---|---|
| TMAX `step_type` max → instant | `test_extreme_temperatures_are_not_read_from_the_degenerate_lead_0_window` |
| `representative_var` drops the era-stable filter | `test_representative_var_skips_a_variable_the_archive_added_later` |
| `representative_var` drops the instant preference | `test_representative_var_prefers_instant_over_windowed` |
| A variable's `grib_index_level` perturbed | `test_every_requested_variable_maps_to_a_real_message` (12 era fixtures) |
| `available_from` gated on valid time | `test_available_from_is_judged_against_the_cycle_not_the_valid_time` |

The last one **survived the first round** — the fix had no test with power over it,
because no shipped variable is both windowed and late-added. Added a test constructing
that combination, which the forecast datasets will reach with real variables.

Two further "checks with no power" were caught and fixed during this pass: the era-fixture
parametrize set was silently empty (an empty `parametrize` *skips* rather than fails — now
asserted at 18), and the fixture test was handing its own fixtures to `file_refs`, which
unlinks the index it is given, so the test **deleted its own fixtures**.

## Operations

Both cron jobs ship **suspended**; a later PR unsuspends after the backfill. Update
`51 3,9,15,21` (f006, the longest lead used, publishes ~init+3h48m), validate 30 min
later plus a 10 minute margin, at `31 4,10,16,22` — the update stops polling 30
seconds before its deadline, so a validator firing exactly at the deadline could read
the store mid-commit. Single writer.

## Checks

`ruff format`, `ruff check`, `ty check` clean; **572 passed** across `tests/noaa/gefs/`,
`common_template_config_subclasses_test.py` and `datasets_cf_compliance_test.py`,
including the slow backfill+update integration test. `generate-manual-workflows` ran.

Part of #825.

## Review round 2 (codex, both PRs read as one clump)

No P0/P1. Findings landing on this PR:

| Finding | Verdict |
|---|---|
| **P2** Hoist the identical 18h `operational_update_window` into `NoaaGefsVirtualRegionJob` and delete the leaf | **Declined**, with a better fix. The two 18h values *are* 18h for the same reason — three update cron fires, so two consecutive missed runs self-heal — but the derivation is per dataset (3 × *that* dataset's own cron cadence), and the cadence lives in `dynamical_dataset.py`, not the region job. The `hrrr` virtual family makes this explicit: one shared base, three leaves, three different values (6h / 12h / 14h). Decisive point: the coming `-16-day-` and `-35-day-` forecast datasets subclass the forecast intermediate, and a 35 day 0.5° `a`+`b` run publishes over a much longer burst behind a much longer pod deadline — its recovery budget is the single number most likely to need to differ, and a silently inherited 18h is exactly how that goes wrong unnoticed. Instead the prose claim is now a check: `test_operational_update_window_spans_three_update_fires` reads the cadence off the update cron's schedule and pins the window to three of them. That catches the reviewer's failure mode *and* keeps the two datasets free to diverge. |
| **P2** Use one canonical definition for the ~20 GEFS quantities shared with `common_gefs_template_config.py` | **Declined.** Measured the actual overlap: 19 names in common, and on those, `short_name` / `long_name` / `units` / `standard_name` / `step_type` / `grib_element` / `grib_index_level` are already identical — the only differing field is `comment` (the window wording), which must differ. The deliberate divergences show up as *different names* (`precipitation_surface` the rate vs `total_precipitation_surface` the raw accumulation), so they are not in the shared set at all. And the stated failure mode — "a correction reaches materialized GEFS but not both virtual products" — is already a *test failure*, not a silent drift: `test_metadata_consistency_across_datasets` enforces same-`name` ⇒ same `short_name`/`long_name`/`standard_name`/`units` across every implemented dataset. A shared quantity table would be a second, parallel mechanism for an invariant a check already holds, threaded through two internal-attrs models whose required fields are disjoint (`grib_description`/`index_position`/`gefs_file_type`/`keep_mantissa_bits` vs `source_file_type`/`available_from`/serializer/filters), for 19 of 65 definitions. The `hrrr` materialized/virtual pairs already define their catalogs separately for the same reason. |
| **P3** Delete the MSLET bisect narration | **Taken.** "Bisected against the published idx files." records how the date was found, which AGENTS.md excludes. The date and the source-inventory fact stay. |
| **P3** The `_s_file_data_vars` docstring claims the omitted surface `HGT` "is served" by the 0.5° `pgrb2b` datasets | **Taken.** Those datasets do not exist, so the claim was false regardless of what happens to the `source_file_types` hook. Now states only the omission and its reason. |
| **P3** Remove the nonfunctional `source_file_types` hook | **Declined** on this PR (the false docstring half is fixed above). The field is load-bearing — `resolution_degrees` derives the grid from it and both leaves declare it — and `_catalog_data_vars`'s `assert` already declares the base s-file-specific with a message naming what is missing, which is what the finding asks for. The `-16-day-` and `-35-day-` 0.5° `a`+`b` datasets need exactly this field; removing and re-adding it is churn. `_SPATIAL_RESOLUTIONS` gains its 0.5° entry with them. |

### Mutation checks

| Mutation | Sole catcher |
|---|---|
| `operational_update_window` 18h → 12h | `test_operational_update_window_spans_three_update_fires` (also the existing literal pin in `region_job_test.py`) |
| Update cron 6-hourly → 8-hourly, **both schedule literals updated to match** | `test_operational_update_window_spans_three_update_fires` — the realistic drift, where the literal assertions are "fixed" and the recovery quietly stops covering three fires |

`uv run main noaa-gefs-analysis-0-25-degree-virtual update-template` produced **no diff** —
byte-identical, confirming this round's source changes are comment-only.

---

## Registry-wide check: every vertical group states its manifest split

`tests/datasets_test.py::test_virtual_group_manifest_splits_are_explicit`, parametrized over `DYNAMICAL_DATASETS`.

A group's arrays carry their vertical dimension's length times the refs of a root array, so a split sized for root arrays leaves that group's manifests oversized by that factor. Neither a missing map key nor a scalar `split_size` errors at construction — the value is simply inherited, silently, and the only symptom is an oversized manifest.

This dataset is root-only, so it cannot trip the check today; the exposure is in the `-10-day-`, `-16-day-` and `-35-day-` datasets that add groups.

**Verified the check fires rather than passing vacuously.** It flags a map carrying `pressure_level` alongside an unlisted `height_above_mean_sea_level`, passes the same map once the entry is added, and catches a new group added to this family's map later.

### Scope: it is about the value being *stated*, not about the value being *small*

The right split for a group is whatever satisfies the reader budget at the lowest manifest count `M` — **not** the root split divided by the group size. That division is the argument that *grouping is not worse than not grouping* for the same data, a comparison between two layouts; it is not a sizing rule, and reusing it as one produces splits far finer than any budget requires. The check therefore asserts only that a group's split is written down. An unstated value is a defect even when it happens to be correct.

### One named exemption

`google-weathernext2-forecast-historical-virtual` applies one split to every array including its `pressure_level` group, and its region job batches writes on that same constant (`manifest_init_split`), so read split and write batching must move together. Whether that shared value meets the group's reader budget has not been audited — that is its owner's call. Exempted **by name**, with the coupling recorded, rather than by weakening the rule to "scalars are fine": a scalar gives a group the same split as a root array, which is precisely the condition this check exists to catch, so a blanket scalar exemption would lose the check exactly where it is needed.

### Merge ordering with #992

This check gates `noaa-gfs-analysis-virtual` (#992), whose analysis map omits its `height_above_mean_sea_level` group while its forecast sibling lists it. That entry is being added there. **Whichever of the two PRs merges second will be red until it lands** — no action needed beyond merging them in either order once #992 carries the entry.

---

## Built on the shared NOAA virtual region job (#990)

`NoaaGefsVirtualSourceFileCoord` and `NoaaGefsVirtualRegionJob` now subclass #990's `NoaaVirtualSourceFileCoord` / `NoaaVirtualRegionJob`. Discovery, `.idx` parsing, byte-range ref building and the message lookup are the base's; what stays here is the GEFS file-path layout, the bucket `ClassVar`s, the era-stable `representative_var`, and the requested-variable completeness check. 132 lines deleted.

**Base extension (9 lines).** `NoaaVirtualRegionJob._check_refs_complete(coord, refs)`, a no-op hook called only on `file_refs`' success path. It is necessary rather than convenient: the base returns `[]` both when it *skips* a file (unparseable index, byte ranges outside the data file) and when nothing matched, so a subclass wrapping `super().file_refs(...)` cannot tell those apart and would silently accept a file missing its only requested variable. Both reviewers confirmed the hook is unreachable on the skip paths and that no subclass-only approach distinguishes them without duplicating the base's loop.

### Two behaviour changes inherited from the base

**1. Every index message's byte range is now checked, not just matched ones.** The deleted GEFS code validated a message's range only after it matched a requested variable; #990's base validates every message before the lookup, and discards the whole file on any bad range. So a corrupt or degenerate range on a message GEFS never requested now discards a file that GEFS previously ingested without trouble. In backfill mode `process_virtual_refs` sweeps once, so such a file is not retried and leaves a NaN row.

This is deliberate in the base and commented there — a corrupt range anywhere condemns the index — but it is a real change for GEFS and was not called out in the migration commit. Checked against all 18 GEFS `.idx` fixtures and the four `tests/noaa/fixtures/ge*.idx`: none has a duplicate offset or a non-increasing span, so nothing in the current archive is affected.

**2. Two element spellings for one chunk are first-wins, not an assertion.** GEFS previously asserted when a file carried two `grib_element_alternatives` spellings for the same chunk. The base de-duplicates on `(var.path, out_loc)`, taking the earlier message. This supersedes the assert rather than losing it: the base's key is level-aware, where the old path-only assert would have false-fired on the second of 26 pressure levels. The test now pins that exactly one ref survives, at the first message's offset, instead of pinning the refusal. No GEFS variable declares an element alternative today, so this is inert in the current catalog.

### Known limitation, for the vertical-group datasets

`_check_refs_complete` keys on `data_var.path`, so it catches a variable that received *no* message. It does not catch partial level coverage — a future pressure-level product delivering 1 of 26 levels would pass. Aligning the key with the base's `(path, out_loc)` needs a group dataset to specify and test against, so it belongs with PR 3 / PR 4 rather than here.

---

## Manifest split sizing, as measured

The config comment states only the timeless basis and points at `docs/virtual_datasets.md` "Manifest splitting". The measurements behind `split_size = 4 * 365 * 8` are recorded here instead, because manifest byte sizes are dated facts that move as variables are added:

| | measured |
|---|---|
| refs per manifest | 11,680 (one ref per 3-hourly step) |
| manifest size | ~224 KiB, at 19.7 bytes/ref on this dataset's own manifests |
| arrays | 38 |
| **refs per commit** | **0.44 M** |
| commit wall, 4 cores | 0.53 s |

Reference point: operational `noaa-hrrr-forecast-48-hour-virtual` sustains **12.1 M refs/commit at p50 2.8 s end to end**, over 12,936 samples across 88 inits. This dataset sits well under that.

Two bounds the value clears with room: the manifest is far inside the 3 MiB single-level reader budget, and 11,680 refs is 11.7× icechunk's 1,000-ref `min_num_chunks` threshold, below which location compression disengages.


src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/templates/latest.zarr/zarr.json
```diff
@@ -0,0 +1,2282 @@
+{
+  "attributes": {
+    "dataset_id": "noaa-gefs-analysis-0-25-degree-virtual",
+    "dataset_version": "0.1.0",
+    "name": "NOAA GEFS analysis, 0.25 degree, virtual",
+    "description": "Weather analysis from the Global Ensemble Forecast System (GEFS) operated by NOAA NWS NCEP, served as references to the source GRIB messages. Each time step is the control member's shortest available forecast lead. Coverage begins with the GEFS v12 0.25 degree archive; the materialized noaa-gefs-analysis reaches further back through coarser and reforecast sources, which cannot share one grid here.",
```
aldenks:
```suggestion
    "description": "Weather analysis from the Global Ensemble Forecast System (GEFS) operated by NOAA NWS NCEP.",
```

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    region_job_class: type[NoaaGefsAnalysis025DegreeVirtualRegionJob] = (
+        NoaaGefsAnalysis025DegreeVirtualRegionJob
+    )
+
+    icechunk_virtual_config: IcechunkVirtualConfig = Field(
+        default_factory=lambda: IcechunkVirtualConfig(
+            containers=gefs_virtual_chunk_containers(),
+            # Four years of 3-hourly steps. Refs per commit = arrays x refs per active
+            # split; see "Manifest splitting" in docs/virtual_datasets.md.
```
aldenks:
```suggestion
            # Four years of 3-hourly steps.
```

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+                split_size=4 * 365 * 8, dim="time"
+            ),
+        )
+    )
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        # f006, the longest lead an analysis step uses, publishes ~init+3h48m.
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{self.dataset_id}-update",
+            schedule="51 3,9,15,21 * * *",
```
aldenks: we should start 3 mins before we expect it, e.g. +3h45m. adjust val run to match

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,87 @@
...
+            suspend=True,
+        )
+
+        return [operational_update_cron_job, validation_cron_job]
+
+    def validators(self) -> Sequence[validation.Validator]:
+        return (
+            # A time is ingested when its own shortest-lead file publishes, hours
+            # after the cycle that covers it starts.
+            validation.CheckCurrentData(max_delay=timedelta(hours=13)),
```
aldenks: if f006 comes by ~init+3h48m. as the comment claims, shouldn't this be a delay of about 4h20m after init?

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/template_config.py
```diff
@@ -0,0 +1,95 @@
...
+    """Virtual GEFS analysis: the control member's shortest available lead at each
+    3-hourly valid time."""
+
+    source_file_types: frozenset[GEFSSourceFileType] = frozenset({"s"})
+    window_comments: dict[str, str] = {
+        "avg": "Average value in the last 6 hour period (00, 06, 12, 18 UTC) or 3 hour period (03, 09, 15, 21 UTC).",
+        "accum": (
+            "Total accumulated in the last 6 hour period (00, 06, 12, 18 UTC) or 3 hour "
+            "period (03, 09, 15, 21 UTC). Subtracting the value at an earlier time with "
+            "the same window start gives the exact total between those two times."
```
aldenks:
```suggestion
            "period (03, 09, 15, 21 UTC)."
```
our users are smart

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/template_config.py
```diff
@@ -0,0 +1,95 @@
...
+            dataset_id="noaa-gefs-analysis-0-25-degree-virtual",
+            dataset_version="0.1.0",
+            name="NOAA GEFS analysis, 0.25 degree, virtual",
+            description=(
+                "Weather analysis from the Global Ensemble Forecast System (GEFS) "
+                "operated by NOAA NWS NCEP, served as references to the source GRIB "
+                "messages. Each time step is the control member's shortest available "
+                "forecast lead. Coverage begins with the GEFS v12 0.25 degree archive; "
+                "the materialized noaa-gefs-analysis reaches further back through "
+                "coarser and reforecast sources, which cannot share one grid here."
```
aldenks:
```suggestion
                "Weather analysis from the Global Ensemble Forecast System (GEFS) "
                "operated by NOAA NWS NCEP."
```

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/dynamical_dataset.py
aldenks: You are right, and the number is now 4h20m.

Worst case is a 00/06/12/18 hour, which waits on its own cycle: for T = C, `discover_available` releases it only once f000 of init C is ready, and the first update fire after C is C+3h45m (the run also picks up T = C+3h from f003 of init C; a non-cycle hour therefore lands ~45m after its timestamp). With `pod_active_deadline=30m` that run has always stopped writing by C+4h15m, so a position lands ~4h15m after its own timestamp at the latest. 4h20m gives that five minutes of margin and sits below the 6h cycle spacing, so a wholly missed cycle is already due at the next validation run (C+4h25m) rather than one run later.

The old 13h was two cycles of slack past that.

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/templates/latest.zarr/zarr.json
aldenks: `templates/latest.zarr/zarr.json` is generated, so I made the change in `template_config.py` and regenerated with `uv run main noaa-gefs-analysis-0-25-degree-virtual update-template`. The regenerated diff carries exactly this string.

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/template_config.py
aldenks: Trimmed the `accum` entry as suggested. Left `avg`/`max`/`min` alone: the `avg` string is byte-identical to the published noaa-gefs-analysis and noaa-gefs-forecast-35-day comment, and `max`/`min` are parallel constructions of it.

src/reformatters/noaa/gefs/analysis_0_25_degree_virtual/dynamical_dataset.py
aldenks: Applied — 13h → 4h20m. Derived it two ways and both agree the value is right: at a 04:25 validate the newest position is C+3h (f000 and f003 of cycle C both land inside the poll window), aged 1h25m, and the *due* position is C since floor_3h(04:25 − 4h20m) = C. The hard bound is 4h25m, so 4h20m leaves five minutes of margin.

One consequence worth recording rather than arguing: this also makes a single failed update pod raise an alarm, not just a wholly missed cycle. Both leave the store at C−3h, aged 7h25m at 04:25, so the check cannot tell them apart — a pod that fails at 03:45 and heals at 09:45 will still page in between. We have that shape in MRMS today (validate at :13 against a ~6 min update, one lost pod trips CheckCurrentData and self-resolves). Leaving the value as you asked since prompt detection of a missed cycle is the point; flagging it so the first spurious page is recognised rather than investigated.

PR #992, opened 2026-09-01
Add the NOAA GFS analysis virtual dataset

aldenks:
Adds `noaa-gfs-analysis-virtual`, a virtual (chunk-reference) Icechunk dataset over the GFS 0.25 degree archive, together with the shared GFS virtual template config and region job that the forecast variant will build on.

**Stacked on #990** (`noaa-virtual-region-job-extraction`), and this PR's base is that branch, not `main`. Earlier revisions of this description claimed the stack while the branch was actually cut from `main`; that was wrong and is now fixed — the branch is rebased onto #990 and `git merge-base --is-ancestor` confirms it.

**GFS consumes the shared module.** `NoaaGfsVirtualRegionJob` subclasses `NoaaVirtualRegionJob` from `reformatters.noaa.noaa_virtual_region_job`, so there is exactly one NOAA implementation of the idx→refs walk rather than a second copy of it. `noaa/gfs/virtual_region_job.py` lost 121 lines to this. What stays GFS-specific is only what the shared base does not express:

- `owns_index_message` — the prefer-`pgrb2` skip of the 41 messages `pgrb2b` repeats byte for byte. This is the one thing the base could not host, so #990's `file_refs` gained a single `owns_index_message(coord, element, level)` hook defaulting to `True`; HRRR's behaviour is unchanged.
- `carried_by` — which of the two products publishes a given variable.
- `source_location_prefix` / `source_bucket_region` — the NODD bucket, via the base's own ClassVars.
- `NoaaGfsVirtualSourceFileCoord.get_url` — the `s3://` rewrite, matching HRRR's shape.
- `representative_var` — the per-product ingestion probe.

The `(variable, out_loc)` first-wins dedupe and the stale-byte-range guard now come from the shared base rather than from a GFS copy; the guard's semantics tighten slightly in the process, since the base checks every message's range rather than only matched ones.

## The dataset

Hourly analysis on the source's own 0.25 degree latitude/longitude grid, one GRIB message per chunk, `time` from **2021-03-22T12:00** — the first cycle of the 0.25 degree archive, verified directly (00Z and 06Z that day and all of 2021-03-21 return 404; 12Z and 18Z return 200). Extending an append dim earlier later is a breaking change, so this is the inclusive start rather than the materialized twins' 2021-05-01.

Each hour takes every variable at the shortest lead the source publishes it at, matching the materialized `noaa-gfs-analysis`: a variable with hour-0 values comes from the cycle its hour falls in (leads 0-5), any other from the cycle before the preceding hour (leads 1-6). So a windowed variable's window always opens at the synoptic hour strictly before its valid time, which is what the metadata comments say in UTC terms.

**293 arrays** — 277 at the root and 16 in a `pressure_level` group on a 57-level float64 coordinate. That is the shared 295-array catalog minus two forecast-only running totals, which at leads <= 6 render the same window string as the 6 hour buckets and would be byte-identical arrays.

## Two source files

Every cycle publishes `pgrb2` (743 messages) and `pgrb2b` (349), and **both are fully in scope** — zero source messages are dropped. 41 `(element, level)` pairs appear identically in both; a `pgrb2b` index skips exactly those and the array is filled from `pgrb2`. The constant is pinned against real indexes at three eras and four leads, and a mutation test (removing one pair) fails 18 tests.

A companion `carried_by()` predicate decides which product publishes a given variable, from the 16 `pgrb2b`-only level strings, the two `pgrb2b`-only surface elements, and the two isobaric elements only `pgrb2` supplies. It is pinned the same way, against every variable in both products' real inventories.

## What was verified rather than assumed

Everything below was measured by decoding real messages with gribberish, not reasoned from documentation.

- **`soil_type_surface` flag table — shipped with no `flag_values`/`flag_meanings`.** The published field is not categorical: SOTYP takes a continuum of values from 0 to 16 quantized to 0.01, so interpolation to this grid leaves values between the class codes. Publishing a flag table would be wrong. The comment names the 16 category STATSGO classification and says values are not exact class numbers. Adding the attributes later is non-breaking.
- **`UFLX`/`VFLX` sign — `standard_name` left unset.** Measured against the 10 m wind: `sign(flux) == sign(wind)` on 0.4% of cells with `|wind| > 2 m/s`, correlation −0.82; Southern Ocean westerlies `u10 = +8.1` give `uflx = −0.20`. NCEP's convention is the *upward* flux of eastward momentum, the opposite of CF's `surface_downward_eastward_stress`, and CF has no upward counterpart. The comment states the convention.
- **`ice_water_mixing_ratio` — `standard_name="cloud_ice_mixing_ratio"` kept.** The string is in the CF table with canonical units `1` and the description "the ratio of the mass of ice to the mass of dry air", which is what GRIB 0-1-23 ICMR carries in the FV3 microphysics.

Also measured, correcting or confirming the authored metadata:

- **Reflectivity floor is −20 dBZ, not HRRR's −10.** 56% of REFC cells sit at exactly −20 (65% and 88% for REFD at 1000 m and 4000 m, 63% at both model levels), second-lowest value −19.99. HRRR's REFC really does floor at −10. The five GFS reflectivity arrays now carry the floor comment with GFS's own value.
- **Freezing-level heights mark "at or below ground" with an exact 0** (23.4% of cells for the 0 °C isotherm, 22.4% for the highest tropospheric freezing level; second-lowest 0.16 m). Both now declare `fill_value=0.0` and carry HRRR's wording, so a CF-aware reader sees NaN. RH at the same levels has no such marker and is unchanged.
- **`ice_cover_surface` is a continuous fraction here and a binary mask in HRRR — both are right.** GFS ICEC has 1001 distinct values with 10.8% strictly between 0 and 1; HRRR's has exactly two, `{0, 1}`. So HRRR's `flag_values=(0, 1)` is correct for HRRR and would be wrong here. Ours carries no flag attributes and a comment saying it takes any value between 0 and 1. Note `test_metadata_consistency_across_datasets` compares only short_name, long_name, standard_name and units — not `step_type`, `comment`, `flag_values` or `flag_meanings` — so a divergence like this is invisible to CI.
- **Cloud ceiling and percent-frozen wording differs from HRRR's because the sources differ.** GFS's `HGT:cloud ceiling` declares no bitmap (section 6 indicator 255) and returns a finite value everywhere, marking "no ceiling" with a drifting ~20000 m value; HRRR's carries a real bitmap (indicator 0, 74.9% NaN). GFS's CPOFP has 501 distinct negative values with no single marker, so nothing can be normalized. Our mask-range wording, copied verbatim from the published materialized GFS, is right for GFS.
- **The NaN-meaning comments were executed against the data.** "NaN over water" matches `LAND == 0` on every one of 1,038,240 cells for all nine variables in that family. "Only over land and sea ice" holds exactly as stated (the valid set is a subset of land ∪ ice; 1,755 marginal-ice cells with a median 4.4% ice fraction are excluded, which the wording does not claim otherwise). "NaN where the terrain is above this level" agrees with the surface height on 99.73-99.99% of cells, the disagreements being cells whose terrain sits within a few hundred metres of the level. "NaN where this potential vorticity surface does not exist" is confirmed by latitude: the +2 PVU surface is 97% NaN south of 60S, 92% at 30-60S and 0% NaN anywhere in the Northern Hemisphere. The 11 cloud-family comments were reworded to what is actually true — every cell with zero cover is NaN, and so are some edge cells with small cover.
- **Grid alignment cross-checked against the published materialized twin.** `temperature_2m` at four widely separated longitudes on 2026-08-28T18:00 matches `noaa-gfs-analysis` to within its rounding quantum (25.844/25.875, 11.844/11.875, 16.144/16.125, 22.544/22.5). A wrong longitude roll would move every value half a globe. Those four values are asserted in the integration test.

## Naming and units policies

**Names** follow an ordered pair: (1) if an equivalent variable already ships in the catalog, match its name and metadata exactly, even where that name under-specifies the aggregation; (2) otherwise encode the aggregation, with `instantaneous_` used only where a published *windowed* variable already owns the plain name.

**This leaves a genuine oddity, stated here so a reviewer finds it explained rather than having to reconstruct it.** `precipitation_rate_surface` is instantaneous while `precipitation_convective_surface` sits beside it as a 6 hour average — the opposite reading of "no prefix". Likewise `total_cloud_cover_atmosphere` and the four `categorical_*_surface` arrays are averages here while the same names are instantaneous in other datasets. It is inherited from the catalog, not introduced here: these variables match opposite published sources.

Worth knowing before concluding we diverged from a settled norm: **the catalog is already split against itself on `total_cloud_cover_atmosphere`, including inside one model family.** It ships as `step_type="avg"` in materialized noaa/gfs, materialized noaa/gefs, `ecmwf/ifs_ens/forecast_46_day_1_5_degree` and the `example_materialized` teaching template, and as `step_type="instant"` in noaa/hrrr, eccc/hrdps, ecmwf/aifs_ens, ecmwf/aifs_single/forecast, dwd/icon_eu and `ecmwf/ifs_ens/forecast_15_day_0_25_degree`. ECMWF IFS ENS publishes it as avg in the 46-day and instant in the 15-day: same provider, same model family, same short_name, opposite quantities.

**Choosing to match our own materialized GFS twin is deliberately flagged and pending your decision.** The reasoning: a materialized/virtual pair over one source is a like-for-like comparison we built on purpose, so a semantic mismatch there produces a silently wrong result, whereas cross-model name agreement is weaker than it looks when the quantities differ by model, grid and resolution anyway.

**Units** follow the same ordered pair. Three divergences from GDAL parity are deliberate:

- **POT stays kelvin**, matching HRRR virtual's `potential_temperature_2m`, though GDAL does convert it. Potential temperature is conventionally kelvin.
- **TSOIL and ICETMP are converted** to Celsius although GDAL does not convert them, TSOIL matching `ecmwf/ifs_ens`'s soil temperatures. Both carry a comment, because a reader comparing against a GDAL-based read of the same message would otherwise see an unexplained 273.15 offset.

**Standing warning, worth carrying to any GRIB dataset: GDAL's unit LABEL is not a usable signal.** Measured exhaustively over all 743 + 349 bands: GDAL relabels **10** GFS elements kelvin to Celsius but converts only **6** (TMP, TMAX, TMIN, DPT, POT, APTMP). TSOIL, ICETMP, LFTX and 4LFTX are relabelled and not converted. Inferring conversion from the label is silently wrong by exactly 273.15 for four of them. Compare values under `GRIB_NORMALIZE_UNITS=YES` vs `NO`, never labels.

## One naming question for you, not acted on

`average_precipitation_rate_surface` (the raw PRATE 6-9 hour average) may be a rule-1 miss. Measured: it and the materialized twins' deaccumulated APCP are the **same physical quantity** — PRATE's window average against `APCP/(3h)` over the matching window gives correlation 0.9999 and a ratio of means of 1.0048.

**But I recommend not renaming.** The materialized `precipitation_surface` deaccumulates to the *previous step* — its own comment says "Average precipitation rate since the previous forecast step", a 1 hour mean — while this array is the mean over the lengthening 1-6 hour window. Same quantity type, different window, so one name would mean two things. Happy to change it if you disagree.

## Manifest splitting

`split_size={r"^/pressure_level/": 3_000, r"^/height_above_mean_sea_level/": 20_000, None: 30_000}` on `time`.

**This is a repartitioning of the same references, not an efficiency gain.** The store holds the same refs either way and total stored manifest bytes are essentially unchanged; the split only picks *where on that partitioning* we sit.

| | manifests | refs each | size each |
|---|---:|---:|---:|
| before | ~4,050 | 4,096 | 0.04 MiB |
| after | ~800 | 30,000 root / 171,000 `pressure_level` / 160,000 height | 0.27 / 1.59 / 1.48 MiB |

Before, the manifests were too small to be worth being objects. After, every class sits inside the reader budgets. **Bytes rewritten per commit do rise, about 6.8×** — that is exactly what taking refs per commit from 1.6 M to 10.8 M means, and it is deliberate, to meet the reference below. Stored bytes unchanged and rewritten-bytes up are both true and are different quantities.

**The reference is one measured configuration, not a family.** `noaa-hrrr-forecast-48-hour-virtual` carries a comparable refs-per-commit and is measured in production at **p50 2.8 s end to end** — source publish to reader-visible, including discovery, idx parse, ref build and commit — over **12,936 samples across 88 inits**. HRRR's 18-hour forecast and analysis carry comparable refs per commit but have **never been latency tested**; they corroborate, they are not further measurements. The group splits are HRRR analysis's own group manifest sizes divided by our level counts, and the catch-all is the value it runs: a configuration match, not a derivation.

Figures above are recomputed against our own measured bytes/ref (9.58 root, 9.72 group) from the shipped template. Counting refs per commit uniformly across templates puts this dataset at 10.8 M and the measured HRRR config at 10.5 M; a differently-scoped count used during review put them at 11.6 M and 12.1 M. The ranking and the conclusion are the same either way, and no figure here depends on which count is used.

**Every vertical group needs its own entry.** The catch-all is sized for root arrays, so a group that falls through gets a window multiplied by its level count, with no symptom other than an oversized manifest. #991 adds the registry-wide invariant enforcing this.

A note for reviewers expecting a template diff: `manifest_split` feeds the icechunk **repository** config, not zarr metadata, so changing it correctly produces **no** change under `templates/latest.zarr`. Verified by regenerating: `manifest_split` is referenced nowhere in the template code path.

## Cron jobs

Both ship `suspend=True`; a follow-up enables them after the backfill. The update fires at init+3h50m, measured from real `Last-Modified` headers (f000 and f006 both land 3h35m-3h51m after their cycle), and polls from there until the cycle's files land.

## Reviews

Two independent reviews were run over the branch diff before any fixes, and synthesized into one pass.

**Taken:**

1. **`representative_var` could pick a probe from the other product** (codex P2, reproduced with `wind_gust_surface` + `temperature_457m_amsl`). A variable-filtered job spanning both products offered the same mixed variable list to each, so the pgrb2b coord probed a pgrb2-only variable and the write loop's probe-coverage assert aborted. Fixed with `carried_by()`: a job now reads only the products carrying its variables and probes one each file fills. Regression test uses codex's exact pair.
2. **Duplicate refs at leads 1-6** (codex P3). The bucket and running total of APCP and ACPCP render one window string there and the index carries two identical messages, so the analysis's single bucket array matched twice and two `VirtualChunkSpec`s were submitted for one chunk. Values were correct because the messages are byte-identical, but it violated the one-ref-per-position invariant. `file_refs` now keeps the first, tested at leads 1, 3 and 6 against real indexes.
3. **Reflectivity floor and freezing-level marker** — the two missing comments, resolved by measurement as described above.
4. **`ice_cover_surface`, cloud ceiling, percent-frozen** — verified; ours are right and so are HRRR's, for different source behaviour.

**Declined:**

- **Renaming `average_precipitation_rate_surface`** — different window under the same name; raised above for your decision.
- **The `instantaneous_`/`average_` prefixes and the `step_type` clashes they create** — a consequence of the settled naming rule; surfaced above for your decision rather than changed.
- **Fixing HRRR's `ice_cover_surface` flag attributes** — measured correct for HRRR's own data, so there is nothing to fix.

Codex found no P0 or P1 issues and independently confirmed the 41-message preference, the pressure-level union, the ScaleOffset math, grid orientation, manifest splitting, the leading-edge gate and synoptic-hour lead selection.

**One unexplained-then-explained blip.** A `GroupNotFoundError` on `analysis_virtual/templates/latest.zarr` failed one test run mid-session and did not reproduce. Rather than leave it, I reproduced it deliberately: `update_template()` rewrites the checked-in store with `mode="w"`, and a test that reads it during that rewrite finds the directory present but the root `zarr.json` gone. It is a local-workflow race between regenerating the template and a concurrently running test, not a defect. Three clean runs of the same 290 tests pass.

## Second review pass (codex, both PRs reviewed together)

Two rounds over #990 + this stack as one body of work. Round 2 returned no P0/P1.

**Taken**
- *A single-variable backfill of any `pressure_level` variable crashes* (P2, both reviewers, reproduced). Fixed in #995; see the merge-order note below.
- *The first analysis hour is written incomplete* (P1). At 2021-03-22T12Z the windowed variables need the 06Z cycle, which predates the archive, while 13Z is complete and authorizes 12Z's instantaneous-only files. `append_dim_start` moved to **2021-03-22T13:00**, the first hour that can ever hold all of its variables, with a regression test. Note the reviewer's suggested alternative — a contiguous complete prefix — would have deadlocked the first backfill, since 12Z never completes.
- *Three copies of the GRIB-index download stub* (P3). Consolidated onto `tests/noaa/grib_index_fixtures.stub_grib_index_download`; the shared and HRRR test modules now use it.
- *Production comments naming test functions* (P3). Removed.
- *Decoder investigation in published metadata* (P3). Five soil/ice temperature comments described how GDAL mislabels the raw kelvin message. That describes a different decoder, not the quantity a reader of this archive gets; removed, keeping the durable NaN semantics.
- *Change-time narration* (P3). "from now through a twenty year archive" and "a follow-up enables them" removed.

**Declined**
- *Stack the GFS work on the shared extraction* (P2) — already done; the finding is an artifact of diffing a stacked branch against `main`, where the shared module legitimately appears in both diffs. `git merge-base --is-ancestor` confirms a single lineage.
- *Generate the repeated level families from one specification* (P2) — the catalog is a reviewed data table whose value is that each of its 295 rows can be checked against the source inventory one-to-one. Generating rows would make that correspondence indirect, which is the property the verification process depends on.
- *Rename `get_idx_url` to `get_index_url` across NOAA* (P3) — a real inconsistency, but it lives in the materialized code paths that neither PR touches.

### Round 3 — convergence

A third round returned no P0/P1, repeated its two earlier findings (family generation, `get_idx_url` naming — both declined again on the same grounds), and raised three comment-accuracy points, all taken:

- **A comment on the completeness validator was false.** It claimed "every ingested position is whole"; `discover_available` deliberately releases an earlier incomplete hour once a later one is complete, which is exactly the interior gap the validator exists to catch. Rewritten to say what the validator is actually for. This is the same behaviour the archive-boundary fix above turns on, so the comment was contradicting a property this PR had already established.
- **The index-fixture docstring narrated the debugging story.** Reduced to its contract: a fresh copy per call, because `file_refs` deletes the index it is given.
- **The respelling test's docstring claimed an archive audit the test does not enforce** (that four v16.x upgrades were separately diffed and found inventory-neutral). Removed; the docstring now claims only what the assertions pin.

Repeats plus comment-accuracy findings, with no new substantive defect, is the convergence signal — review stopped here.

## Testing

`ruff format`, `ruff check`, `ty check` clean; **2,318 non-slow tests pass**, 142 in `tests/noaa/gfs` including the slow ones.

Coverage worth naming: the whole 293-array catalog is run against **real indexes at three eras** and every message is required to reach exactly one array position (696/741/306/308 refs at f000 and f009 per product), which is what would catch a wrong level string, element spelling or rendered window string. The hour-0 overrides are derived from real f000 indexes rather than asserted from a literal. The lead rule is enumerated over all 24 hours of a day, because the off-by-one at 00/06/12/18 UTC produces plausible output rather than an error. The integration test runs `backfill_local` then `update` over four variables spanning both products and the pressure group, with snapshot values including a 875 hPa level only `pgrb2b` carries.

A shared `tests/noaa/grib_index_fixtures.py` hands each test a fresh copy of a real index, because `file_refs` deletes the index it is given — a test that wrote its fixture once at the download path would be consumed by the first call and pass meaninglessly ever after.

---

## Visibility: the analysis convention for a ragged leading edge

`discover_available` holds the append frontier back, and the reason is specific to the *shape* of an analysis rather than to GFS's two products.

An analysis `time` position is assembled from **several different lead times** — a variable with hour-0 values comes from f000 of the cycle its hour falls in, a windowed variable from a longer lead of an earlier cycle. Those leads publish hours apart. Without a gate the newest `time` positions would be systematically ragged: the long-lead contributions for an hour arrive well before that hour's own f000, so the leading edge would show a shifting subset of variables that is an artefact of lead timing rather than of what the source has published.

The convention is therefore: **hold data from leads beyond f000 until that position's f000 is available.** `noaa-hrrr-analysis-virtual` applies the same rule; this dataset inherits it rather than inventing an exception.

**What the gate costs here, measured:** median **8 s**, p90 **894 s**, max **908 s**, with **85.7% of positions visible within a minute**. The median is tight because the analysis takes short leads; the tail is not, because a position waiting on the previous cycle waits for that cycle to publish, which is a quarter-hour-scale event however short the leads are.

This is an analysis-shaped rule and does not generalise. The forecast sibling in #995 has **no** gate and should not: there, a partially-filled `init_time`/`lead_time` position is the intended behaviour, because readers want the latest available data as each file lands.

## The height-above-mean-sea-level group

The 24 root arrays `{temperature,wind_u,wind_v}_{305,457,610,914,1829,2743,3658,4572}m_amsl` become **3 arrays in a `height_above_mean_sea_level` group**, on a coordinate of those 8 heights in metres.

`ice_growth_rate_10m_amsl` deliberately stays a root suffix: different element, and 10 m is not one of the 8. The grouping is driven off `(element, level type)`, never off the name — an `_amsl` or "mean sea level" string match would sweep in `PRMSL` and `MSLET`, which use mean sea level as a *reduction target* rather than a height above it.

**Coordinate metadata.** `long_name="Height above mean sea level"`, `standard_name="height_above_mean_sea_level"`, `units="m"`, `positive="up"`, `axis="Z"`. Not `altitude`: CF defines `altitude` as height above the **geoid**, while `height_above_mean_sea_level` is height above the time-mean sea surface. The GRIB2 source level type is 102, "specified altitude above mean sea level", so CF has an exact match and `altitude` would assert a geoid reference the source does not claim — the same class of error as `x_wind` on a lat/lon grid. Both names are in `tests/common/cf-standard-name-table.xml`, the table CI checks against.

**Why it qualifies.** The family is fully dense — all 24 (element, height) combinations present — verified at 8 dates including the archive's literal first init 2021-03-22T12Z. The same three quantities also existing in `pressure_level` is not a conflict: the group sets the zarr path, so `pressure_level/temperature` and `height_above_mean_sea_level/temperature` are distinct arrays of one quantity on two vertical coordinates.

**GFS-specific: the family splits across products.** 305/457/610/914/4572 m come from `pgrb2b` and 1829/2743/3658 m from `pgrb2` — and note **4572 m is the topmost yet sits in `b`**, so this is not a clean high/low cut. A constant probe level per (group, product) is valid because each product's level inventory is uniform across all three elements, verified at three eras; `_PROBE_VERTICAL_LEVEL` pins `pgrb2`→1829 m and `pgrb2b`→305 m, parametrized per (group, product, era, lead). Swapping the pair fails all 24 height cases.

**What grouping costs.** A virtual chunk is one GRIB message, so the vertical chunk size is 1: grouping changes neither the number of refs nor the bytes they occupy. What it changes is array count — 24 arrays become 3 — and commit cost is fixed-plus-marginal with the fixed term scaling per array. Both terms therefore fall or stay flat and neither can rise, since grouping adds no arrays and no refs. On this dataset the height group contributes **0.11 s** of a ~2.37 s commit.

The approval basis is unchanged: density (24/24 across every era in both models), CF coordinate correctness, structural consistency with the existing `pressure_level` group, and no cost penalty. The first three are untouched by any of the cost-model work.

## How the two products divide the catalog

Nine vertical families make up everything that is not a plain surface or single-level field. Measured from the catalog and `carried_by`, at three eras:

| family | arrays | from `pgrb2` | from `pgrb2b` |
|---|---:|---:|---:|
| isobaric (`pressure_level` group) | 16 | 16 | 14 (different levels of the same arrays) |
| potential-vorticity surfaces | 48 | 12 | 36 |
| height above mean sea level | 25 | 10 | 15 |
| mb-above-ground layers | 37 | 12 | 25 |
| below-ground soil layers | 12 | 12 | 0 |
| sigma levels | 10 | 10 | 0 |
| cloud-layer levels | 25 | 25 | 0 |
| tropopause / max wind | 13 | 13 | 0 |
| hybrid model levels | 6 | 6 | 0 |

**Every family `pgrb2b` touches, it shares with `pgrb2`.** `pgrb2b` never introduces a vertical family of its own — it extends four families `pgrb2` already establishes, and contributes nothing to the other five. So dropping `pgrb2b` would thin four families rather than remove any; and no reader of a family has to know which product a given member came from.

The extension is by *member*, not by array: in the four shared families `pgrb2b` supplies different surfaces, heights or layers of the same quantity, and only in the isobaric group do the two products supply different levels *of one array*.

## Why the potential-vorticity surfaces are in scope

Three independent reasons, of which only the third is specific to GFS:

1. They are a dense, comparable set — eight surfaces × six quantities — so they behave like any other vertical family rather than a scattering of one-offs.
2. They are cheap: 48 of 295 arrays, and a virtual dataset pays no rewrite cost for an array it references.
3. **GFS-specific:** the family is split across the two products (12 arrays from `pgrb2`, 36 from `pgrb2b`). Dropping `pgrb2b`'s share would leave a family that looks complete at ±2 PVU and silently has nothing at the other six surfaces — a worse failure than not carrying the family at all. This reason does not transfer to a model whose products do not split a family.

## Why refs are matched on index element strings

The GRIB2 parameter triple (discipline, category, number) is the stable identifier — it did not change across the CLWMR→CLMR respelling, as the control below asserts. It is also **absent from the `.idx` sidecar**, which carries only the element string, the level string and the window string.

Reading the triple requires fetching and parsing the message itself. For a virtual dataset whose entire premise is that the build never reads the data — it reads a ~50 KB sidecar and writes byte ranges — fetching every message to identify it would defeat the point and cost roughly the whole archive in transfer.

So string matching is **forced**, not chosen, and `grib_element_alternatives` is a necessary consequence rather than a workaround for sloppy matching: when the source respells an element, the string that identifies it changes while the quantity does not, and the alternatives tuple is the only place that can be recorded.

`test_the_respelled_element_still_selects_the_same_grib_parameter` is the one control on that mechanism. It fetches the selected message on each side of the transition and asserts it decodes to discipline 0, category 1, number 22. A string-level check cannot catch an alternatives tuple that picks up a spelling belonging to a *different* parameter; mutating the tuple to `("ICMR",)` makes the control fail with `(0, 1, 23)`, which is what makes it load-bearing.

**What that control does not show.** Identical parameter octets on both sides mean the respelling did not change *which parameter* the message carries. They do **not** mean the respelling changed no published data: the control reads the parameter section, not the data section, and it reads one level of one variable at one lead. A concurrent change to the values, the packing, or the level coverage would be invisible to it. The evidence that nothing else changed is separate — the full `(element, level, window)` symmetric-difference inventory diff across the eras, which is empty once the rename is filtered out.

## ⚠ Merge-order note — two defects in this PR's code are fixed in #995

#995 (the forecast variant, stacked on this one) found two defects in code **this** PR introduces, and fixes them there:

1. **A single-variable backfill of any `pressure_level` variable crashes.** The representative probe pins 1000 hPa, which `pgrb2b` never publishes. This affects the analysis dataset too, and it bites precisely the variable-filtered targeted re-backfills.
2. **`convective_precipitation_*_surface` metadata was wrong.** `ecmwf-aifs-single-forecast-virtual` already owns that name with `cp` / "Convective precipitation"; this catalog carried the NCEP `acpcp` / "Convective precipitation (water)". A rule-1 (match-the-published-equivalent) miss. Corrected in #995, which also regenerates this dataset's template.

**Consequence:** merge the stack in order (#990 → #992 → #995), and **do not run the analysis backfill between #992 and #995** — the probe crash would hit the first variable-filtered pass. Both fixes can be moved down into this branch instead if you would rather each PR stand alone; say the word.

Part of #823.


src/reformatters/noaa/gfs/virtual_region_job.py
```diff
@@ -0,0 +1,171 @@
+from collections.abc import Mapping
```
aldenks: Pickup reviewing here

PR #993, opened 2026-09-01
Add the NOAA GEFS 10 day 0.25 degree virtual forecast dataset

aldenks:
**Stacked on #991** (`gefs-virtual-analysis`), which is not yet merged. This branch is cut from it, so the diff against `main` also contains #991's shared base and analysis dataset. The surface to review here is `git diff gefs-virtual-analysis...HEAD`. Targeting `main` so Alden can sequence the merges; if #991 gains review changes to the shared base, this rebases onto its updated head.

Adds `noaa-gefs-forecast-10-day-0-25-degree-virtual`: every message of the 0.25° `pgrb2s` file, all 31 ensemble members, leads 0–240h in 3h steps, inits every 6h from 2020-09-23T12. Both cron jobs ship `suspend=True`.

## How the forecast intermediate is factored

`NoaaGefsForecastVirtual{TemplateConfig,SourceFileCoord,RegionJob}` sit between the shared GEFS virtual base and the 10-day, mirroring HRRR's `NoaaHrrrForecastVirtualRegionJob` (which has two real subclasses). They own everything the 16-day and 35-day will share: the `(init_time, ensemble_member, lead_time, latitude, longitude)` shape, the forecast coordinate set, the lead-time-phrased window comments, and the `(init × member × lead × file type)` coord product. The 10-day subclasses add only `forecast_length`, `dataset_attributes`, and an 18h `operational_update_window`.

Deliberately **not** added: hooks with one implementation and no second caller. Review flagged three (`_forecast_resolution`, `_expected_forecast_length_values`, `_expected_forecast_length_statistics`) and I inlined them — a later PR extracts them in the commit that needs them. `lead_times()` stays, because `dimension_coordinates` and the tests both call it and the 16-day/35-day genuinely need to override it (their leads change spacing at 240h). Grouping by `var.internal_attrs.source_file_type` is the seam where those datasets will resolve `s` vs `a`/`b` per lead.

## Verified against real source (contradicting #825 where noted)

- **81 leads (0–240h step 3h) × 31 members** (`gec00` + `gep01..gep30`), from a full S3 listing of the `pgrb2sp25` prefix. `geavg`/`gespr` are statistics, excluded.
- **9548/9548** `(element, level, window)` triples the catalog requests matched real archived idx, across 3 eras × all 81 leads × the whole catalog. All 401 distinct rendered window strings appear in the real 2026 idx.
- **Member inventories are byte-identical** across `gec00`/`gep01`/`gep30` in all six eras sampled — so the 34 → 35 → 38 counts PR 1 measured on the control member hold for every member.
- **Publication timing contradicts the natural assumption.** GEFS `s` does not trickle from init+0 the way HRRR/GFS do: `f000` lands ~init+3h47m and the last member's `f240` ~init+5h37m, in one ~110-minute burst (4 inits sampled: f000 3h46m–3h49m, end 5h30m–5h37m). This is what sets the cron shape — fire at init+3h45m rather than near init+0, which would idle for 3¾ hours. #825 gave f240 at ~init+5h29m; measured 5h31m–5h37m.
- `expected_forecast_length` is constant 240h: the archive-start init already carries `f240`.

## `discover_available` for a forecast: ungated

The analysis gates on own-cycle completeness because each of its `time` steps is assembled from two files published hours apart, so releasing the windowed half alone would extend `time` to a step carrying only half its variables. A forecast has no such assembly: each `(init, lead, member)` cell is filled by exactly one file, and a lead is complete the moment its file lands. Gating an init until all 2511 files arrived would trade per-file visibility for a ~110-minute wait and buy nothing, since the cells released early are individually whole. The ragged leading edge is expected forecast semantics, and it exists only inside the update's own poll window: `CheckVirtualManifestCompleteness` clamps to the store's append-dim extent, so an init the source has not started is skipped rather than failed, and by the time validation fires the burst has closed. This uses the shared base's obstore listing unchanged; the open `max(authorized)` question on the analysis gate is analysis-only and does not reach here.

## Manifest split arithmetic

Objective `1.115e-9·M² + k·Σ(arrays × MiB)` with `k_root = 0.077`, subject to ≤3 MiB per single-level manifest (no vertical group in the `s` product).

- Refs per array per init = 81 leads × 31 members = **2511**; 38 arrays.
- **Measured bytes/ref: 17.75**, from backfilling one full split-aligned 16-init window (40176 refs in one manifest → 713,113 B = **0.680 MiB**), read back with `list_manifest_files`. The brief expected ~14.7 from HRRR; GEFS's longer per-member URLs compress less well. At ~2500 refs/manifest the same store measures 16.5–17.7, so this is not a small-manifest artifact.
- Setting `c₂ = 0.077 × 38 × 2511 × 17.75 / 2²⁰ = 0.1244 s` per init of split, the optimum is `N³ = 2 × 1.115e-9 × 38² × A² / c₂`: **N ≈ 12.9** at today's A ≈ 8766 inits, **N ≈ 19.4** at a five-year A ≈ 16000.
- Chose **N = 16** (four days), within ~6% of optimal across that whole horizon: 2.47 s/commit today, 3.40 s at five years.
- Reader budget: 40176 × 17.75 = 0.68 MiB ≤ 3 MiB ✓. Compression floor: 40176 ≥ 1000 refs ✓. Manifest count M = 38 × 548 ≈ 21k, recorded in the code comment beside the bytes figure.

## Reviews

Both reviews returned **no P0 and no P1**. Review A independently re-verified the arithmetic that could corrupt data — idx matching over all 20 fixtures, era gating inclusivity, that lead 240 renders `234-240 hour acc fcst` rather than the `0-10 day` running-total form, `out_loc`/probe-cell agreement, ScaleOffset direction, the 18h→3-init window — and found it correct.

**Taken**

| # | From | Finding | Fix |
|---|---|---|---|
| 1 | A (P2) | The completeness comment promised guarantees the 0.95 tier cannot give. One whole member is 81/2511 = 3.2% missing → 96.8% present → **passes**. A variable-shaped gap is invisible: presence is probed per source file via `representative_var`. | Rewrote the comment around what the check proves, and added the true fact: a missing member passes only while its init is newest, then fails at the next fire under the 1.0 tier. Test now asserts the hole (`81 < allowed_missing`). |
| 2 | A (P2) + self | Forward-looking docstrings narrating subclasses that don't exist; three hooks with one implementation. | Stripped every forward-looking sentence; inlined the three hooks. Kept the two intermediate classes (brief + HRRR precedent). |
| 3 | B (P3) | The idx fixture test asserted refs == `coord.data_vars`, but both sides come from the same `available_from` filter — moving `GEFS_B22_TRANSITION_DATE` later would drop three variables and still pass, contradicting the test's own docstring. | Expected count now derived from the **fixture file**: idx lines minus only the messages this dataset declines (`HGT@surface`; the degenerate `0-0 day` TMAX/TMIN at lead 0). |
| 4 | B (P3) | Validation cron's 30-minute `pod_active_deadline` unpinned. | Asserted. |
| 5 | A (P3) | `_stalled_cycles_before_alerting` docstring said "tolerates" but returned the alerting count, and re-implemented the due-position arithmetic. | Now drives the real `CheckCurrentData.check` via a `ValidationContext`; named and documented correctly. |
| 6 | A (P3) | "A cycle that published nothing still fails" was backwards — an un-ingested cycle has no append-dim position, so completeness *skips* it. | Corrected; names `CheckCurrentData` as the check that catches it. |
| 7 | A (P3) | Discovery test asserted method identity, not behaviour. | Now asserts a partly-published init releases every available file. |
| 8 | A (P3) | `GEFS_ENSEMBLE_MEMBERS` / `GEFS_S_FILE_LEAD_FREQUENCY` are source facts sitting in the virtual template module, while `GEFS_S_FILE_MAX` already lives in `gefs_config_models.py`. | Moved, and used in `forecast_35_day` so one way of doing things remains. Its template regenerates **byte-identical** (verified). |
| 9 | A (P3) | Manifest count worth recording beside the bytes figure. | Added. |
| 10 | review follow-up | Validator fired at exactly update-fire + deadline, but `_virtual_poll_deadline` is fire + deadline − 30s, so it could read the store mid-commit. | Validator now fires 10 minutes later (06:25), matching HRRR-48h. `max_delay` is since derived from that offset — see "Operational validator bounds, derived" below. **#991 adopted the same margin.** |
| 11 | self | Cron comment claimed "~35 minutes of slack"; measured 38–44 min. | Reworded to a claim that stays true. |

**Declined**

- **`virtual_region_job.py`'s `assert not lookup` fires for any variable declaring `grib_element_alternatives`** — one lookup key per spelling, only the matched one popped. Real: reproduced here with a synthetic `("GUSTold",)` alternative on `wind_gust_surface` (AssertionError against a real fixture; without it, one ref, fine). Latent for now — no shipped GEFS virtual variable declares alternatives (`{}` across all 38). **Declined here and routed to #991**: those are PR 1's lines, the finding was raised on PR 1's review, and this branch is stacked on it, so it inherits the fix on rebase whereas fixing both would collide. Contributed a correction to the fix shape — one key maps to a *list* of variables, so the invariant must be tracked per-variable with `.get()`, not per-key with `pop`, or a message shared by two variables satisfies only one.
- **Decode-health sample size** (~5.4k GRIB decodes per fire — the largest in the catalog). Fits the 30-minute deadline; recorded as the first knob to turn (`sampled_leads`) if validation ever runs long. No change.

## Mutation checks

32 mutations, **all killed** once aimed at the file the behaviour actually lives in. Each was applied to one source line, the named test confirmed failing, then restored with `git checkout HEAD -- <path>`.

| Mutation | Test that failed |
|---|---|
| lead step 3h → 6h | `test_lead_times_cover_the_s_file` |
| `forecast_length` 240h → 384h | `test_lead_times_cover_the_s_file` |
| 31 members → 30 | `test_all_thirty_one_ensemble_members` |
| dim order: `ensemble_member` after `lead_time` | `test_forecast_time_structure` |
| chunks no longer span longitude | `test_one_chunk_per_message_spans_only_the_grid` |
| window reset 6h → 3h | `test_every_lead_time_carries_the_window_its_comment_promises` |
| window reset 6h → 12h | `test_differencing_only_holds_within_a_window_start` |
| accum comment drops the window-start qualifier | `test_window_comments_are_phrased_in_lead_time` |
| forecast attrs fall back to analysis wording | `test_dataset_attributes` |
| `valid_time` no longer adds lead | `test_template_carries_the_forecast_coordinates` |
| only the control member gets a file | `test_every_init_lead_member_triple_gets_exactly_one_file` |
| lead 0 keeps the windowed variables | `test_lead_zero_carries_only_the_instant_variables` |
| `available_from` ignored | `test_a_variable_is_not_sourced_before_its_available_from` |
| control member named like a perturbed one | `test_source_file_coord_url_names_the_member` |
| update window 18h → 24h | `test_operational_update_jobs_single_polling_job` |
| manifest split 16 → 32 | `test_manifest_split_holds_four_days_of_inits` |
| `max_delay` 6h20m → 6h30m, and → 6h10m | `test_validators` |
| leading tier restored to `(0.95, 1.0)` | `test_validators` |
| update fires an hour later | `test_operational_kubernetes_resources` |
| validator margin removed (back to 06:15) | `test_operational_kubernetes_resources` |
| validation `pod_active_deadline` dropped | `test_operational_kubernetes_resources` |
| cron name abbreviation changed | `test_cron_job_names_fit_the_kubernetes_limit` |
| APCP idx level wrong | `test_every_requested_variable_maps_to_a_real_message` |
| B22 transition moved to 2030 | `test_every_requested_variable_maps_to_a_real_message` |
| MSLET availability moved later | `test_every_requested_variable_maps_to_a_real_message` |
| forecast discovery gated to complete inits | `test_forecast_discovery_releases_a_partly_published_init` |
| `out_loc` omits `ensemble_member` | `test_file_refs_span_each_message_and_end_at_the_file_end` |
| TSOIL loses its K→C filter (in-code config) | `common_template_config_subclasses_test` |
| TSOIL loses its K→C filter (checked-in template) | `test_backfill_local_and_operational_update` |
| moved constants: members 31 → 30, lead step 3h → 6h | `test_all_thirty_one_ensemble_members`, `test_lead_times_cover_the_s_file` |

Three mutations survived round one, and each taught something:

- A **dim-order swap** survived the chunk test — correctly, since chunks are `(1,1,1,721,1440)` regardless of how the three size-1 dims are ordered. The order claim belongs to `test_forecast_time_structure`, which kills it. Re-aimed.
- **Removing TSOIL's K→C filter from one per-array `zarr.json`** survived the e2e test, because the checked-in template also carries codecs in its **consolidated root document**. Mutating both copies kills it (295.86 K vs 22.71 °C). The in-code config change is caught separately by `common_template_config_subclasses_test`.
- A **`get_url` pattern** never applied (my escaping), not a survival; killed once fixed.

Also checked explicitly for the three zero-power shapes: the idx fixture set is guarded by `assert len(_IDX_FIXTURES) == 20`, and removing the fixtures **errors at collection** rather than skipping; the fixture test copies each file before `file_refs` unlinks it, so fixtures are byte-identical after two consecutive runs.

## Comment claims executed

Every comment asserting a computational property was run as a script, with the wrong world it rules out:

| Claim | Wrong world ruled out | Result |
|---|---|---|
| Windowed values span 6h at leads 6, 12, 18… and 3h at 3, 9, 15… | A running-total variable (window always `0-lead`), or lead 240 rendering the `0-10 day` day-form window | 0 violations over 15 windowed vars × 81 leads |
| Those exact window strings exist in the archive | A window the catalog renders but the source never publishes | 0 absent from the real 2026 idx |
| "Subtracting the value at an earlier lead time **with the same window start**" | An *unqualified* differencing sentence — which would be **false**: only 40 of 3160 lead pairs share a start, and adjacent leads 6 and 9 do not (`0-6` vs `6-9`) | Qualification confirmed load-bearing; kept it rather than following the brief's "unqualified" phrasing |
| 0.95 ≈ all but the last four lead times of every member | A tier that silently allows five leads' worth | Allowed exactly 125 missing; 4×31 = 124 passed, 5×31 = 155 failed. Superseded by `(1.0,)` — see below |
| One whole member "still fails" | — | **Refuted at 0.95**: 81 missing → 96.8% → passed. Now true, at the `(1.0,)` the review settled on. |
| The source lays that tail down in its final ~20 minutes | A tail spanning hours | 17–22 min across 4 inits |
| `max_delay=6h20m` alerts on the first stalled cycle | A threshold that alerts on the normal case, or tolerates one | Confirmed by driving the real check; the 1-vs-2 boundary sits at exactly 6h25m |
| 31 members are `gec00` + `gep01..gep30` | `geavg`/`gespr` counted as members | Confirmed against the S3 listing |
| f000 ~init+3h47m, f240 ~init+5h37m, deadline clears it by over half an hour | A source that publishes from init+0 (which would make the fire time wrong by 3¾ h) | Confirmed; slack 38–44 min |
| 16 × 81 × 31 = 40176 refs at 17.8 B/ref = 0.68 MiB | Carrying HRRR's 16.4 B/ref over untested | Measured on this dataset's own manifests |

## Shared / frozen code

`noaa/noaa_grib_index.py` untouched — `_lead_time_str` renders every window this dataset needs, including `234-240 hour acc fcst` at lead 240, verified against real idx. The one shared-base change I wanted and did not make is the `grib_element_alternatives` assert above, routed to #991.

## Checks

`ruff format`, `ruff check --fix`, `ty check` clean. `pytest -m "not slow"` 2432 passed; `tests/noaa/gefs/` 213 passed including the real-source `backfill_local` + `update` integration test with snapshot assertions for all five variables it processes at lead 0/3/6/240 and members 0/30 (lead 0's accumulation is NaN; ocean cells NaN for soil and snow). Template regenerated and in sync; `generate-manual-workflows` ran.

Not run: any backfill of the real store. The dataset is unpopulated and both crons are suspended.

## Follow-up: the window wording is now machine-checked

The `window_comments` wording was correct for all three GEFS forecast lead domains, but correct by luck: it enumerates two lead time sequences, and every GEFS lead happens to be a multiple of 6 hours or 3 hours past one. A variant with different spacing would ship a silently wrong comment — a 4 hourly tail past 240 gives 2 and 4 hour windows, 24 of 36 tail leads unenumerated.

`NoaaGefsForecastVirtualTemplateConfig._validate_window_comments_describe_every_lead_time` now checks, for every lead time the config serves, that the window running from the last whole multiple of the 6 hour reset frequency *strictly below* the lead (so lead 6 spans 0–6) has a length the wording enumerates. It is a `model_validator`, so the 16 and 35 day datasets inherit it rather than needing their own, and a bad domain fails at config construction.

**Wording and check share one source.** `_WINDOW_LEAD_TIME_SEQUENCES` maps window length to its lead time sequence; the four comment strings interpolate the phrase built from it and the assert tests its keys. Round 2 showed the original version of this claim overreached, and the mapping is now *derived* rather than written out — see "Round 2" below. Emitted comment text is byte-identical; the template regenerates with no diff.

Tests (`tests/noaa/gefs/virtual_template_config_test.py`), with the wrong world each rules out:

| Test | Wrong world ruled out | Result |
|---|---|---|
| `test_ten_day_lead_times_are_described` | The guard rejecting this dataset's own domain (3 hourly 0–240, 81 leads) | Passes |
| `test_sixteen_day_lead_times_are_described` | A guard the 16 day domain (3 hourly to 240, then 6 hourly to 384, 105 leads) trips over — that dataset does not exist yet, so nothing else would catch it | Passes |
| `test_thirty_five_day_lead_times_are_described` | Same for the 35 day domain (6 hourly tail to 840, 181 leads) | Passes |
| `test_lead_times_the_window_comments_cannot_describe_are_rejected` | A guard that cannot fire: a 4 hourly tail from 244 | Raises naming lead 244 and its 4 hour window |
| `test_forecast_length_shorter_than_the_first_window_is_described` | An off-by-one at the first window, where lead 3 is a partial and lead 0 has no window at all | Passes |

Mutation-checked: deleting the validator fails `test_lead_times_the_window_comments_cannot_describe_are_rejected` by name (`DID NOT RAISE`) and leaves the other four green. Each domain test asserts its lead count so an empty or mis-built domain cannot pass vacuously.

`noaa/noaa_grib_index.py` still untouched. The analysis virtual config's parallel wording ("00, 06, 12, 18 UTC") is unguarded — it is not a subclass of the forecast config, and extending the check there was out of scope for this change.

`ruff format`, `ruff check --fix`, `ty check` clean; `tests/noaa/gefs/` 218 passed; `tests/common/common_template_config_subclasses_test.py` and `datasets_cf_compliance_test.py` 425 passed; `update-template` produces no diff.

## Review round 2 (codex, both PRs read as one clump)

No P0/P1. Findings landing on this PR:

| Finding | Verdict |
|---|---|
| **P3** Validate the emitted window *sequences*, not only their duration keys | **Taken — the reviewer is right and the claim it contradicts was mine.** The guard checked that every lead's window *duration* was a key in `_WINDOW_LEAD_TIME_SEQUENCES`; the sequence labels (`"6, 12, 18, ..."`) were free strings. So "an edit to either cannot leave them disagreeing" was false: they cannot disagree about which *durations* exist, but they could absolutely disagree about which *leads* belong to which duration. Rather than narrowing the claim in a comment, the sequences are now read off the lead time grid — `_window_lead_time_sequences()` walks `GEFS_S_FILE_LEAD_FREQUENCY` steps over three `WINDOW_RESET_FREQUENCY` periods and groups the leads by `_window_length()`, the same function the guard uses — so naming a lead under the wrong window is no longer expressible. The class-level comment that restated the two sequences in prose is gone with it. Emitted text and template are byte-identical. |
| **P3** Pin the forecast validation margin relationally and fix its stale arithmetic | **Taken, both halves.** The `6h15m` was stale (03:45 + 2h30 + 10 min = 06:25, so the newest init is **6h25m** old). Rather than just correcting the number, `test_validators` now derives it from the validation cron's own schedule and asserts `validation_fire - newest_init == 6h25m`, and `_stalled_cycles_before_alerting` takes that derived fire instead of a hardcoded one. `test_operational_kubernetes_resources` now also asserts the ten-minute relationship — `validation fires == update fires + pod_active_deadline + 10 min` — mirroring #991's sibling test instead of pinning independent literals. |
| **P2** Hoist the identical 18h `operational_update_window` into the shared base and delete the forecast leaf | **Declined**, with the same reasoning and the same substitute check as on #991: the two 18h values are 18h for the same reason (three update cron fires), but each is derived from *its own* cron cadence, and the coming `-16-day-` / `-35-day-` subclasses of this PR's forecast intermediate are exactly where a silently inherited 18h would be wrong. `test_operational_update_window_spans_three_update_fires` now pins the relationship instead. |
| **P3** Defer the forecast intermediate until it has a second production subclass | **Declined.** The general rule is right — an intermediate with one subclass usually has not earned its place — but the premise does not hold here. `noaa-gefs-forecast-16-day-0-5-degree-virtual` and `-35-day-` are specified with their source facts already measured, they subclass this intermediate, and a decision is already taken to put a discovery-withholding override on it. `tests/noaa/gefs/virtual_template_config_test.py` already exercises both of their lead domains through the intermediate's guard, which is only possible because the guard lives there — collapsing the layer into the 10 day leaf would delete two of the five tests' reason to exist and move ~220 lines out and back within one dataset cycle. The parallel `hrrr` family reached the same shape (`NoaaHrrrVirtualRegionJob` → `NoaaHrrrForecastVirtualRegionJob` → per-length leaves). |
| **P3** Remove the nonfunctional `source_file_types` hook / **P3** delete the MSLET bisect narration | Both land on #991's lines; triaged there (docstring claim fixed, narration deleted, field kept). |

### Mutation checks

| Mutation | Sole catcher |
|---|---|
| `_WINDOW_LEAD_TIME_SEQUENCES` hardcoded again with the two sequences **swapped**, and all three literal string assertions in `template_config_test.py` "fixed" to match | `test_window_comments_name_the_lead_times_that_carry_each_window` — this is exactly the wrong world the old key-membership guard could not see |
| `pod_active_deadline` 2h30 → 2h20 with its own literal updated, validation schedule left alone | the derived margin assertion in `test_operational_kubernetes_resources` (`[25, 385, …] != [15, 375, …]`) |
| Update cron 6-hourly → 8-hourly with both schedule literals updated | `test_operational_update_window_spans_three_update_fires` |
| Validation cron 06:25 → 06:35 with its literal updated | `test_validators` (the 6h25m pin) and the margin assertion |

`ruff format`, `ruff check`, `ty check` clean; `tests/noaa/gefs/` **220 passed** including the
slow real-source integration test. `uv run main noaa-gefs-forecast-10-day-0-25-degree-virtual
update-template` produced **no diff** — byte-identical, which is the proof that deriving the
window sequences was a pure definition-site refactor.

One knowingly-duplicated three line test helper: `_fire_minutes` now exists in both this PR's
and #991's dataset test. It collapses to one shared helper when the stack is rebased; a
cross-branch import would have left this branch red until then.

---

## Manifest split sizing, as measured

The config comment states only the timeless basis. The measurements behind `split_size = 16` are recorded here instead, because manifest byte sizes are dated facts that move as variables are added:

| | measured |
|---|---|
| refs per manifest | 16 × 81 leads × 31 members = 40,176 |
| manifest size | 0.68 MiB, at 17.8 bytes/ref on this dataset's own manifests |
| arrays | 38 |
| **refs per commit** | **1.5 M** |

Reference point: operational `noaa-hrrr-forecast-48-hour-virtual` sustains **12.1 M refs/commit at p50 2.8 s end to end**, over 12,936 samples across 88 inits. This dataset sits well under that.

Two bounds the value clears with room: the manifest is far inside the 3 MiB single-level reader budget, and 40,176 refs is 40× icechunk's 1,000-ref `min_num_chunks` threshold, below which location compression disengages.

---

Carrying #991's review over. Both bounds were round numbers; both are now derived from the
publication lags this PR measured.

### `CheckCurrentData` — `max_delay` 12h → **6h20m**

Two bounds squeeze it, and only one of them is the one you first reach for:

| | |
|---|---|
| `pod_active_deadline` | **2h30m** |
| worst-case moment init *T* becomes present | init + 6h15m |
| validate cron fires | init + **6h25m** |

- **Lower bound, > 6h15m.** The binding instant is not the validate fire — at the fire a healthy
  store already holds init *T*. It is the last instant at which *T* is legitimately still absent:
  the end of its own update run. Just before init+6h15m the newest position present is *T*−6h, so
  anything ≤ 6h15m false-alarms on a perfectly healthy store.
- **Upper bound, ≤ 6h25m.** For the validation run that *follows* a missed cycle to catch it, *T*
  must already be due when that run fires. Beyond 6h25m the miss goes unreported until the next
  fire, six hours later.
- **Chosen 6h20m** — 5 min above the lower bound, 5 min below the hard bound.

Verified against the real check rather than re-derived: `test_validators` drives
`CheckCurrentData` over stalled-cycle scenarios and pins that **the first** stalled cycle alerts
(was: the second). The 1-vs-2 boundary sits at exactly 6h25m, so 6h20m is inside it with margin.

### `CheckVirtualManifestCompleteness` — `(0.95, 1.0)` → **`(1.0,)`**

Stated in files, since that is what the check counts:

| | |
|---|---|
| expected source files per init | 81 leads × 31 members × 1 product = **2,511** |
| last file of a run lands | init + **5h37m** |
| poll window closes | init + 6h15m |
| validate fires | init + **6h25m** |
| `0.95` silently tolerated | 2511 − ⌈0.95 × 2511⌉ = **125 missing files** |

The whole burst closes 48 minutes before validation fires, and the update polls through it. So
there is nothing left for a leading tier to excuse: every init the store reached must be whole,
which is the plain default.

At `0.95` a whole ensemble member absent from the newest init (81 files) passed — the PR body
above records that as **Refuted** under the old tier. At `(1.0,)` it fails.

## Cron job naming

The dataset id plus `-validate` is 54 characters against kubernetes' 52, so the cron names
abbreviate. They now drop `-degree` and **keep** the dashed resolution, rather than borrowing the
provider's `p` spelling:

| | before | after | len |
|---|---|---|---|
| update | `noaa-gefs-forecast-10-day-0p25-virtual-update` | `noaa-gefs-forecast-10-day-0-25-virtual-update` | 45 |
| validate | `noaa-gefs-forecast-10-day-0p25-virtual-validate` | `noaa-gefs-forecast-10-day-0-25-virtual-validate` | **47** |

Checked in code, not by eye — `CronJobName` is `max_length=52` and
`test_cron_job_names_fit_the_kubernetes_limit` asserts it. `.github/workflows/manual-create-job-from-cronjob.yml`
was regenerated by the `generate-manual-workflows` step.

URLs, and `FILE_RESOLUTIONS`. Only our own identifier changed.

## Metadata trims

- **Accumulation window comment** — dropped "Subtracting the value at an earlier lead time with the
  same window start gives the exact total between those two lead times." Only the `accum` string
  changed; `avg`/`max`/`min` are untouched, since `avg` is byte-identical to published
  `noaa-gefs-analysis` / `noaa-gefs-forecast-35-day` and `max`/`min` are parallel constructions
  that `datasets_cf_compliance_test` holds to each other.
- **Dataset description** — cut to the one sentence the published materialized siblings already
  use, byte-for-byte: `Weather forecasts from the Global Ensemble Forecast System (GEFS) operated
  by NOAA NWS NCEP.`
- **Manifest split comment** — reduced to the single orienting line, matching #991.

Template regenerated; the diff is exactly those two strings and nothing else.


src/reformatters/noaa/gefs/forecast_10_day_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+    )
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        # The dataset id plus "-validate" exceeds the 52 character cron job name limit.
+        cron_job_name_prefix = self.dataset_id.replace("-0-25-degree", "-0-25")
+        # The whole run publishes in one burst: f000 lands ~init+3h47m and the last
+        # member's f240 ~init+5h37m. Fire just before the burst starts and poll through
+        # it; the deadline clears the observed end by over half an hour and still ends
+        # well before the next cycle's fire.
```
aldenks:
```suggestion
        # A run publishes ~init+3h47m through ~init+5h37m.
        # Fire just before the first files become available and stop 30m after expected completion.
```

src/reformatters/noaa/gefs/forecast_10_day_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,89 @@
...
+            memory="7G",
+            secret_names=self.store_factory.k8s_secret_names(),
+            workers_total=1,
+            parallelism=1,
+            suspend=True,
+        )
+        validation_cron_job = ValidationCronJob(
+            name=f"{cron_job_name_prefix}-validate",
+            # The update's fire plus its pod_active_deadline, so the run being
+            # validated has always stopped writing.
```
aldenks:
```suggestion
            # The update's fire plus its pod_active_deadline
```

src/reformatters/noaa/gefs/forecast_10_day_0_25_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,89 @@
...
+        return (
+            # A cycle that published nothing is caught here rather than by the
+            # completeness check below, which skips append dim positions the store
+            # does not reach.
+            validation.CheckCurrentData(max_delay=timedelta(hours=6, minutes=20)),
+            # The whole run publishes before validation fires, so every init the store
+            # reached must be whole.
+            validation.CheckVirtualManifestCompleteness(),
+            validation.CheckVirtualDecodeHealth(),
+        )
```
aldenks, line 81:
```suggestion
        return (
            validation.CheckCurrentData(max_delay=timedelta(hours=6, minutes=20)),
            validation.CheckVirtualManifestCompleteness(),
            validation.CheckVirtualDecodeHealth(),
        )
```

src/reformatters/noaa/gefs/forecast_10_day_0_25_degree_virtual/template_config.py
```diff
@@ -0,0 +1,34 @@
...
+)
+
+
+class NoaaGefsForecast10Day025DegreeVirtualTemplateConfig(
+    NoaaGefsForecastVirtualTemplateConfig
+):
+    """Virtual GEFS 10 day forecast: every 0.25 degree pgrb2s message of all 31
+    ensemble members, out to the 240 hour lead where that file ends."""
+
+    source_file_types: frozenset[GEFSSourceFileType] = frozenset({"s"})
```
aldenks:
```suggestion
    source_file_types: tuple[GEFSSourceFileType] = ("s",)
```
is there really a benefit to using a set vs a tuple at n < 10?

this is a suggested change to the base class type and all children who set it

aldenks: Applied, and you were right that nothing relied on set semantics. Two deviations from the literal suggestion worth flagging:

**Annotation is `tuple[GEFSSourceFileType, ...]`, not `tuple[GEFSSourceFileType]`.** The latter is a 1-tuple type, so it would reject the `("a", "b")` the 0.5 degree children declare.

**Order is the source's own: `s`, `a`, `b`** — the order `GEFSSourceFileType = Literal["s", "a", "b"]` and `FILE_RESOLUTIONS` already use. So `("s",)` and `("a", "b")`. This matters now in a way the frozenset hid: the two dispatch comparisons in `_catalog_data_vars` compare by order, so a child declaring `("b", "a")` would fall through to the assert. It fails loudly at import with the offending value, which seemed better than adding a normalizing validator.

Scope: this is the **config-level** field only. `NoaaGefsVirtualInternalAttrs.source_file_types` — the per-variable one behind `_S_FILE`/`_A_FILE`/`_B_FILE`/`_A_AND_B_FILES` and ~300 call sites — stays a frozenset, since your comment was anchored on the config declaration and those constants are still used per variable.

Children updated: the base plus `noaa-gefs-analysis-0-25-degree-virtual` (already merged, so I checked its template regenerates **byte-identical** — it does, `source_file_types` is not serialized and `resolution_degrees` goes through a set comprehension), the 10 day here, the 16 day in #1017, and the shared a/b base in #1019. The 35 day inherits and needed no change.

PR #995, opened 2026-09-01
Add the NOAA GFS forecast virtual dataset

aldenks:
Adds `noaa-gfs-forecast-virtual`, the last code PR of the GFS virtual series (#823).

**Stacked on #992**, which is stacked on #990, and this PR's base is `gfs-virtuals` rather than `main`. Earlier revisions of this description claimed a stack that did not exist — all three branches were cut from `main` independently. That is fixed: #990 → #992 → #995 is now a real ancestor chain on the remote, and the diff to read is against `gfs-virtuals`.

This dataset's region job inherits the shared NOAA idx→refs implementation through `NoaaGfsVirtualRegionJob`, which #992 rewrote to subclass `NoaaVirtualRegionJob` from `reformatters.noaa.noaa_virtual_region_job`. Nothing in this PR reimplements `file_refs` or `_message_lookup`, and the window-string helper is referenced by its post-#990 name, `grib_index_window_str`.

## The dataset

`init_time` × `lead_time` × `latitude` × `longitude`, plus a `pressure_level` group — 295 arrays (279 root + 16 isobaric), one virtual chunk per GRIB message on the source 0.25° grid, from both `pgrb2` and `pgrb2b`.

- 6-hourly inits from **2021-03-22T12:00**, the first cycle of the 0.25° archive (00Z and 06Z that day are 404; 12Z and 18Z are 200 — probed, not taken from a note).
- **209 lead times**: hourly f000–f120 then 3-hourly f123–f384. The spacing changes inside the one `lead_time` dim and every position is populated; f121 and f122 do not exist at any era.
- Every one of `pgrb2`'s 743 messages and `pgrb2b`'s 308 non-duplicate messages reaches an array position at every lead > 0 (696 / 306 at f000).

## The run-total variables, and the shared index line

`total_precipitation_run_total_surface` and `convective_precipitation_run_total_surface` are **forecast-only** — the analysis excluded them because at leads ≤ 6 they render the same window string as the 6 hour bucket. Here they are distinct quantities:

| lead | bucket | running total |
|---|---|---|
| f003 | `0-3 hour acc fcst` | `0-3 hour acc fcst` (one message) |
| f009 | `6-9 hour acc fcst` | `0-9 hour acc fcst` |
| f024 | `18-24 hour acc fcst` | **`0-1 day acc fcst`** |
| f120 | `114-120 hour acc fcst` | **`0-5 day acc fcst`** |
| f123 | `120-123 hour acc fcst` | `0-123 hour acc fcst` |
| f384 | `378-384 hour acc fcst` | **`0-16 day acc fcst`** |

At leads 1–6 the index carries **two byte-identical messages** under one key and both variables take a ref to the first: one index line fanning out to several variables is fine, one variable taking two refs for one chunk is not. Tested at leads 1, 3 and 6 across three eras: both arrays filled, one ref each, same byte offset; at f009 the offsets differ. The integration test pins it end to end — 8.5625 for both at f003, 5.5625 vs 23.25 at f009.

## Window comments: enumerated, not stated

The wording deliberately differs from the analysis sibling's (which phrases windows in UTC times, having no `lead_time` dim), and it **enumerates window lengths rather than restating the reset rule** — keeping the rule in prose as well as in `_lead_time_str` is what produced three successive off-by-one drafts.

- Bucket: `Accumulated over the 6 hour window containing this step, which restarts every 6 hours of forecast lead time: the window is 6 hours long at lead times divisible by 6 and shorter at every other lead time (1 to 5 hours where lead times are hourly, 3 hours where they are 3-hourly). Subtracting the value at an earlier lead time with the same window start gives the exact total between those two lead times.`
- Running total: `Accumulated from the forecast initialization time to this step, so the window lengthens with lead time and never resets. Subtracting the value at an earlier step gives the exact total between those two steps.`
- Average / maximum / minimum: the bucket's first sentence with `Averaged over` / `Maximum value over` / `Minimum value over`, no differencing sentence.

The differencing sentence is qualified for the bucket and unqualified only for the running total, and that split keys on `window_reset_frequency`, not on the dataset kind. Enumerated over the real 208 published leads: the bucket shares a window start in **344 of 21,528 lead pairs (1.6%)**, and 63 of the 207 adjacent pairs do not — an unqualified sentence there would be false for almost every pair a reader tries. The running total shares one in 21,528 of 21,528. Both facts are pinned by a test, as are the length claims (hourly leads produce {1,2,3,4,5,6}; the 3-hourly tail produces only {3,6}, so "1 to 6 hours" would be false for 88 of 208 leads).

## Visibility: each file's refs are written as it arrives

This dataset has no completeness gate, by design. A file's refs are committed as soon as that file lands, so the currently-publishing `init_time`/`lead_time` position is normally partially filled — some lead times present, others not yet, and a level column drawing on both products may have one product's levels before the other's.

That is the intended behaviour for a forecast, not a gap: readers want the latest available data, and "the `pgrb2` file is here so its refs are written, `pgrb2b` has not landed yet" is a correct description of a forecast still publishing. Withholding it would trade freshness for a tidiness no forecast reader asked for.

The per-file latency promise in `docs/virtual_datasets.md` therefore **holds** for this dataset: a source file's data is visible within seconds of publication, with no sibling wait in the path.

**Cadence is the framework default: one commit per discovery sweep, paced at most once per second.** The `tick_interval` override this dataset previously carried is removed — it was the only one in the codebase, so every virtual dataset now runs the 1 s default. It was wrong on its own terms: the loop sleeps `tick_interval − elapsed` clamped at zero, so the tick is a *floor on polling cadence*, not a cap on batch size. Under load the sleep is zero and the loop already self-batches, picking up whatever arrived while the previous commit ran. The override bought deliberate idle while published files sat undiscovered, for an effect that was already free.

**The real consequence, so it is not a surprise in review:** each source file is committed exactly once, taking this dataset from ~24 commits per init to up to ~418, each rewriting the full active window of the arrays it touches. **Update-pod CPU per init rises roughly proportionally.** That is accepted, and it is not a projection — `noaa-hrrr-forecast-48-hour-virtual` runs exactly this shape today: 147 files, 1 s tick, ~147 commits, p50 2.8 s end to end, inside a comparable poll deadline.

The analysis sibling in #992 *does* gate, for a reason specific to its shape rather than to the two-product split — see that PR.

## Manifest sizing

`split_size={r"^/pressure_level/": 16, r"^/height_above_mean_sea_level/": 16, None: 128}` on `init_time`. Unchanged.

**Sized so refs per commit (arrays touched × refs per split) sits in line with the operationally tuned HRRR virtual datasets.** Counting uniformly across shipped templates, this dataset is at **9.9 M** against **10.2–10.7 M** for the three HRRR virtual datasets. The one production latency measurement we have is `noaa-hrrr-forecast-48-hour-virtual` at a comparable refs-per-commit: **p50 2.8 s end to end**, source publish to reader-visible, over 12,936 samples across 88 inits. HRRR's 18-hour forecast and analysis carry comparable figures but have never been latency tested — they corroborate rather than measure.

Coarser is also protective: fewer manifests means a smaller manifest scan. On that scan term, superseding an earlier revision of this description which said it was not real — that was wrong: the `1.115e-9` coefficient was fitted to an assumed law whose exponent our rigs never tested, so the coefficient should not be relied on, but the scan itself is present and material at high manifest counts, with a real-archive observation of commits going from 11 s to 150 s at M ≈ 257,000. Splits are sized to keep manifest counts low, not to balance a modelled term.

Measured full manifests: 0.24 MiB root, 1.77 MiB `pressure_level`, 0.25 MiB `height_above_mean_sea_level`.

A note on our local rig, which earlier revisions of this description leaned on for absolute predictions: it disagrees with production by 5–20× on the closest analogue, so it is cited here only for the relative core-scaling result it was designed to measure, and not for commit-time predictions.

**Every vertical group needs its own entry.** The catch-all is sized for root arrays, so a group falling through gets a window multiplied by its level count, with no symptom other than an oversized manifest. #991 adds the registry-wide invariant enforcing this.

## Reviews

Two independent reviews (a Claude sub-agent and codex via `$review-agent`) over the diff against `gfs-virtuals`.

**Raised by both, taken:**
- **The vertical-group probe level (P2).** Both reproduced the crash independently. Fixed as above, with tests at both levels of the stack.
- **The window comment's boundary wording (P2/P3).** Both flagged that "the most recent multiple of 6 hours" reads as the current step at f006/f012; the Claude review additionally showed "lengthens from 1 to 6 hours" is false past f120, where the source is 3-hourly. Superseded by the enumeration rewrite above, which removes the rule statement rather than patching it.

**Raised by one, taken:**
- **Validation cron left ~30 s of margin** over the update's poll deadline (fire + deadline − 30 s), so "the run being validated has always stopped writing" was not guaranteed. Moved validation to `55 5,11,17,23` (10 minutes clear) and tightened `CheckCurrentData` to 11 h, the tightest deadline that still absorbs one missed cycle.
- **`_with_window_comment` was a third near-identical copy.** Extracted `prepend_comment` into `common/config_models.py` and migrated all three call sites (both GFS virtual configs and `noaa-hrrr-analysis-virtual`). All three templates regenerate byte-identically.
- **No end-to-end check of the `kg m-2` → `m` ScaleOffset.** Added `snow_water_equivalent_surface` to the integration snapshot: 0.143832 m on the central Greenland ice sheet, where a missing divide-by-1000 would read as 143 m of snow.
- **Comment staleness and repetition.** Dropped hardcoded array/lead counts where they were incidental, de-duplicated the publication timings, and restated the 18 h window comment to say what the window is *for* (self-healing missed fires) rather than repeating the publication times.
- **A test docstring overstated its claim** ("turns every one of pgrb2's 743 messages into a ref" is true only by cancellation at leads 1–6). Reworded.

**Declined:**
- *Add a clause to all 44 window comments explaining that lead 0 is NaN.* The comment text is a settled cross-suite standard, and a zero-length window at lead 0 already follows from the sentence; the HRRR forecast virtual sets no precedent for it either.

## Checks

`ruff format`, `ruff check`, `ty check` clean. `pytest tests/noaa/gfs tests/common tests/datasets_test.py tests/noaa/hrrr` — 2020 passed. Whole fast suite — 2368 passed. Template regenerated and committed; `generate-manual-workflows` run.

## Owner review follow-up

The accumulation comments now name the window and stop there, in both the forecast and analysis virtual configs; the sentence about subtracting an earlier value to get the exact total between two steps is gone. `test_the_differencing_sentences_are_true_for_the_pairs_they_claim` was deleted rather than reworked: its whole premise was that the bucket carries a qualified differencing sentence and the running total an unqualified one, and with neither sentence in the metadata there is no claim left for it to check. The window-length and window-start facts it leaned on stay covered by `test_the_bucket_comment_describes_every_window_length_the_source_produces`.

The analysis virtual's wording rides its already-queued metadata refresh, so the two datasets describe equivalent variables identically.

### Follow-up required after merge

This PR changes convective precipitation variable attributes in the shared GFS virtual template config, which `noaa-gfs-analysis-virtual` also consumes. After this merges and deploys, that store needs a metadata-only backfill (Manual: Backfill, `operation=overwrite-metadata`, ref `main`) to pick up the new attributes, verified afterwards by diffing every array's attributes on the store against the template.


src/reformatters/common/config_models.py
```diff
@@ -263,6 +263,12 @@ def stores_hour_0_values(self) -> bool:
         return self.has_hour_0_values() and not self.internal_attrs.deaccumulate_to_rate
 
 
+def prepend_comment[V: DataVar[Any]](var: V, sentence: str) -> V:
```
aldenks: has this shown its need to be in a repo wide common code? unless there are other uses planned imminently after this PR, revert the addition here and the factoring out of _with_window_comment in the template config py

src/reformatters/noaa/gfs/forecast_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+                },
+                dim="init_time",
+            ),
+        )
+    )
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        # This forecast uses leads 0-384, which publish ~init+3h32m to ~init+5h34m.
+        # Fire a few minutes before the earliest and poll until the last one lands. The
+        # deadline bounds waiting on a file that never publishes.
```
aldenks:
```suggestion
        # A full forecast publishes ~init+3h32m to ~init+5h34m.
        # Fire a few minutes before the earliest and poll until the last one lands. The
        # deadline bounds waiting on a file that never publishes.
```
change to remove ambiguity of if 384 is the full forecast or not

src/reformatters/noaa/gfs/forecast_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{self.dataset_id}-update",
+            schedule="29 3,9,15,21 * * *",
+            pod_active_deadline=timedelta(hours=2, minutes=30),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="4",
```
aldenks: 3.5 cpu to fit on 4 cpu node

src/reformatters/noaa/gfs/forecast_virtual/template_config.py
```diff
@@ -0,0 +1,226 @@
...
+
+_BUCKET_EXTENT = (
+    "the preceding 1-6 hours of forecast lead time, since the last lead time divisible "
+    "by 6 before this step."
+)
+_BUCKET_COMMENTS = {
+    "accum": (
+        f"Accumulated over {_BUCKET_EXTENT} Subtracting the value at an earlier lead "
+        "time with the same window start gives the exact total between those two lead "
+        "times."
```
aldenks:
```suggestion
        f"Accumulated over {_BUCKET_EXTENT}."
```
our users are smart

src/reformatters/noaa/gfs/forecast_virtual/template_config.py
```diff
@@ -0,0 +1,226 @@
...
+    ),
+    "avg": f"Averaged over {_BUCKET_EXTENT}",
+    "max": f"Maximum value over {_BUCKET_EXTENT}",
+    "min": f"Minimum value over {_BUCKET_EXTENT}",
+}
+_RUN_TOTAL_COMMENT = (
+    "Accumulated from the forecast initialization time to this step, so the window "
+    "lengthens with lead time and never resets. Subtracting the value at an earlier "
+    "step gives the exact total between those two steps."
```
aldenks:
```suggestion
    "Accumulated from the forecast initialization time to this step."
```

PR #1000, opened 2026-09-02
Add UCSB CHC CHIRPS analysis final and preliminary datasets

### LINKED ISSUE #1006: CHIRPS `final` and `preliminary` materialized datasets
UCSB CHC CHRIPS global precipitation datasets. Two materialized analysis datasets for their preliminary and final products. Both at daily step, chunked for time series analysis. CHRIPS v3 data.

- `ucsb-chc-chirps-analysis-final` — the final product (`final/rnl`), 1981-01-01 onward.
- `ucsb-chc-chirps-analysis-preliminary` — the preliminary product (`prelim/sat`), 2025-01-01 onward.

aldenks:
## What

Adds two materialized Icechunk datasets for CHIRPS v3.0 daily precipitation:

- `ucsb-chc-chirps-analysis-final`: `final/rnl`, starting 1981-01-01.
- `ucsb-chc-chirps-analysis-preliminary`: `prelim/sat`, starting 2025-01-01.

The display names are `UCSB CHC CHIRPS analysis, final` and `UCSB CHC CHIRPS analysis, preliminary`. The products share implementation but never substitute for one another. Final uses ERA5 daily disaggregation; preliminary uses IMERG and the station observations available in near real time. Source exploration: #1006.

`precipitation_surface` is the average rate over the 24 hours beginning at its daily time coordinate, converted from mm/day to `kg m-2 s-1`. The source's -9999 sentinel becomes NaN, including oceans where CHIRPS makes no estimate. Variable names and physical-quantity attributes follow IMERG/MRMS. CHIRPS retains eight mantissa bits under the precipitation-rate convention; those siblings currently retain seven. Both templates use the concise product descriptions, precipitation comment, and attribution requested in review.

## Layout and measurement

Chunks are `(365, 60, 60)`; shards remain `(365, 1200, 1800)`: 600 chunks per shard and eight spatial shards.

Remeasured 365 actual daily `final/rnl` files for 2025, using 300×300 Amazon, Congo, and Sahara windows. The pipeline is float32 source-fill masking, rate conversion, eight-bit mantissa rounding, then Blosc Zstd level 3 with byte shuffle. The table uses complete 365-day chunks, decimal MB, and the wet Amazon/Congo samples:

| Spatial chunk | Raw MB | Wet samples | Compressed min / median / max MB |
| --- | ---: | ---: | ---: |
| 50×50 | 3.650 | 72 | 0.748 / 1.117 / 1.219 |
| **60×60** | **5.256** | **50** | **1.105 / 1.563 / 1.728** |
| 75×75 | 8.213 | 32 | 1.778 / 2.512 / 2.693 |
| 100×100 | 14.600 | 18 | 3.229 / 4.445 / 4.788 |

The proposed 0.05 compression ratio is not supported over wet land; measured sizes support 60×60 for the requested 1–2.5 MB range. Sahara's 60×60 median is 0.288 MB; ocean chunks are smaller still. These measurements do not establish compression for preliminary or other source years. A shard is 3.1536 GB raw; about 0.94 GB compressed is a wet-land estimate at a 0.3 ratio, not a measured whole-shard size.

Owner checks matched a full-year chunk byte-for-byte through the actual Zarr codec, verified its round trip, and matched retained tiles against fresh January, July, and December source downloads. A global one-day selection still decodes 25.2288 GB because the time chunks span 365 days; its logical requested float32 payload is 69.12 MB.

## Updates and validation

Updates replay the shard containing the newest stored day and request subsequent days through now. The shared base trims the template to the newest successfully read day. Interior missing days remain NaN and do not block advancement. There is no accepted-date list or CHIRPS-specific trimming override.

A failed reread of an already stored day can replace that day's values with NaN when a later day succeeds. A subsequent replay can refill it; after the shard leaves the update window, recovery needs a targeted backfill. Icechunk rejects updates whose proposed time dimension retracts and leaves main unchanged in that case. The newest-three-day NaN check does not detect older interior gaps.

The preliminary archive currently lacks nine source days, 2025-02-26–28 and 2025-03-26–31. Historical source 404s log exceptions; those dates are ordinary missing data, not special cases in code. Publication-related not-found logging uses `expected_unavailable_window`: 60 days for final, 10 for preliminary, also used by freshness validation. Other materialized datasets keep the 48-hour default.

Final updates run at 00:00 UTC on every third day of the month (`0 0 */3 * *`); validation runs at 01:00 UTC on those same dates (`0 1 */3 * *`), exactly one update active deadline later. Keeping both fires on the same date preserves that pairing across month boundaries. Preliminary updates run daily at 19:00 UTC with validation at 20:00 UTC; sampled winter source files arrived after the previous 17:00 UTC update time. Both products retain a 60-minute update deadline and a 30-minute validation deadline. NaN validation samples 40 points over the newest three days, excludes points that are always NaN in that window, and applies a 5% threshold to the remainder.

Update pods request 3.5 CPUs, 50G memory, 25.5G shared memory, and 20G ephemeral storage; one worker downloads at most eight files concurrently. The global 365-day float32 processing buffer alone is 25.2288 GB. The 50G request is provisional. **Tuning reminder: after the first full-shard run, measure peak memory and tune the request back down.** All update and validation crons remain suspended pending backfill and the full-shard memory/runtime check.

## Review decisions and verification

Earlier review pass: `refmt-chirps-pr1000-claude` posted six comments (four initial, two follow-up); `refmt-chirps-pr1000-codex` posted none and assessed all 21 existing author threads. Both supported the shared gap behavior. Removed the inaccurate final-publication timing comment and updated the description's safety, layout, precision, resource, and test claims.

The shared cron parser now supports stepped days of month, using one bounded field parser for hours and days. Its month-boundary tests include a skipped month end. The reviewers disagreed about whether this extension was needed: retained it because the existing registered-dataset schedule test parses every cron, while correcting the claim that CHIRPS materialized updates themselves call the parser. Claude withdrew the objection after checking that test.

Deliberately retained eight mantissa bits rather than matching the siblings’ seven; retained 60/10-day freshness windows because the observed batches still meet those deadlines under the selected schedules; and kept the three-day NaN check rather than adding product/date-specific historical-gap policy. Broader validation could detect older gaps, but persistent source gaps would then keep failing validation. Oversized positive cron steps remain valid, matching the scheduler’s semantics. The coordinator and second-look settled these decisions with no unresolved disagreement. The memory-tuning reminder was also passed to the supervisor at the author’s request.

Owner verification passed:

- `uv run ruff format`, `uv run ruff check --fix`, and `uv run ty check`.
- `uv run prek run generate-manual-workflows --all-files`.
- `uv run pytest -n 2 -m "not slow"`: 2,356 passed, including the shared template/CF checks and eight update/validation schedule-pairing cases.
- `uv run pytest -n 0 tests/ucsb_chc/chirps -m slow`: all 10 affected slow tests passed.

The schedule regression covers 31-day, short-month, and leap-year boundaries. An independent check confirmed that the old final schedule fails the pairing relationship at leap-February and April boundaries while the new schedule passes.

CI runs the full suite; local slow testing is limited to CHIRPS and any other touched slow tests.


aldenks:
Applied the synthesized CHIRPS review fixes.

Changed:

- Added explicitly accepted preliminary source gaps for 2025-02-26–28 and 2025-03-26–31. Updates still stop at unexplained gaps, try accepted dates on every replay, and end on a successfully read day. Added replay, late-arrival, trailing-gap, and Icechunk failed-publication regression coverage.
- Made expected missing-file logging configurable, preserving the shared 48-hour default. CHIRPS uses the existing 60-day final and 10-day preliminary freshness delays, and treats accepted source-gap 404s as expected.
- Corrected the preliminary description to include station observations and identified the final ERA5-disaggregated `rnl` series; regenerated both templates.
- Fixed the shared mutable raster mock and replaced finiteness-only checks with precipitation-value assertions. Tightened comments and documented gap recovery.

Deliberately unchanged:

- Product selection (`final/rnl`, `prelim/sat`), IDs, starts, chunk/shard layout, rounding, resources, schedules, and validator thresholds. Proposed schedule/resource changes lacked evidence of a failure; full-shard runtime and memory remain checks for the first deployment.
- No extra retry wrapper: the shared obstore client already configures 16 retries with a five-minute retry budget.
- No age-based acceptance of gaps or scan starting only at the newest stored day: those can hide failed downloads or allow old good data to be overwritten. New permanent gaps require an explicit reviewed update to the accepted-date set.
- No alias relocation, timestamp rewrite, or speculative multi-variable generalization. The dataset currently has one variable; revisit completeness across variables if that changes.

Accepted gaps must be removed from the set if CHC restores the files; a slow source test flags restoration. A restored date left in the set could become NaN again after a failed replay. Restored dates outside the active shard require a targeted backfill. Record the historical nine-day gap in the validation report's review notes during validation.

Three independent reviews (Fable correctness, Astra correctness, Astra style) produced 10, 2, and 5 findings. Fable synthesized them into seven change groups: 10 original findings accepted, 2 deferred, 5 rejected, with duplicates consolidated. All disagreements were settled. Follow-up review found no new Astra findings and three low-priority Fable items: two documentation/readability corrections applied, one already covered by the downloader unit test.

Validation: repository-wide ruff format/check and ty passed; manual workflow generation passed; full fast suite passed (2,346 tests, `-n 2`, including all 423 template/CF cases); all 11 affected CHIRPS slow tests passed serially (`-n 0`). The full unrelated slow suite is left to CI.

Pushed commit: [40dfe2a6](https://github.com/dynamical-org/reformatters/commit/40dfe2a6aa5b6a4a727fca6df9864266ed2e7d72).

docs/ops_card.md
```diff
@@ -21,4 +21,5 @@ Accessible via manually triggered github actions. Follow link and click "run wor
...
\ No newline at end of file
+- **Update fails**: Look at issues and logs. Failed jobs usually require a code change to fix (e.g. structural change to data at the source). If it appears a code change is needed, make a PR, merge it, wait for the deploy action to complete, then re-run the update and validation workflows. If it appears transient, re-run the update job followed by validation.
+- **CHIRPS source day permanently missing** (`CheckCurrentData` fails and the update's newest published day stops advancing while later days exist at the source): Confirm the day returns 404 at the source, add it to that product's `known_missing_days`, merge and deploy, then re-run update and validation. Remove the day from `known_missing_days` once the source publishes it. If updates no longer replay the day's shard, fill it with an `overwrite-chunks` backfill targeting that shard. Record historical dates in the validation report's Review notes, not variable metadata.
```
aldenks: Doesn't belong here, remove

src/reformatters/common/materialized_region_job.py
```diff
@@ -49,6 +50,8 @@ class MaterializedRegionJob(
     # will be pipelined within a region job.
     max_vars_per_download_group: ClassVar[int | None] = None
 
+    expected_missing_window: ClassVar[timedelta] = timedelta(hours=48)
```
aldenks:
```suggestion
    expected_unavailable_window: ClassVar[timedelta] = timedelta(hours=48)
```

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/precipitation_surface/zarr.json
```diff
@@ -0,0 +1,85 @@
...
+        "index_location": "end"
+      }
+    }
+  ],
+  "attributes": {
+    "long_name": "Precipitation rate",
+    "short_name": "prate",
+    "standard_name": "precipitation_flux",
+    "units": "kg m-2 s-1",
+    "comment": "Average precipitation rate over the 24 hours starting at the time coordinate. Units equivalent to mm/s. NaN where CHIRPS makes no estimate, which is the ocean and marginal seas; large lakes carry values.",
```
aldenks:
```suggestion
    "comment": "Average precipitation rate over the 24 hours starting at the time coordinate. Units equivalent to mm/s. NaN over oceans.",
```

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/zarr.json
```diff
@@ -0,0 +1,324 @@
+{
+  "attributes": {
+    "dataset_id": "ucsb-chc-chirps-analysis-final",
+    "dataset_version": "0.1.0",
+    "name": "UCSB CHC CHIRPS analysis final",
+    "description": "Daily precipitation from the Climate Hazards Center Infrared Precipitation with Stations (CHIRPS) version 3.0 final product (rnl variant), which incorporates quality controlled station observations and splits each pentad total into days using ERA5.",
```
aldenks:
```suggestion
    "description": "Daily precipitation from the Climate Hazards Center Infrared Precipitation with Stations (CHIRPS) version 3.0 final product.",
```

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/zarr.json
```diff
@@ -0,0 +1,324 @@
...
+    "attribution": "UCSB Climate Hazards Center CHIRPS version 3.0 data processed by dynamical.org from the Climate Hazards Center data archive. Cite Funk, C., Peterson, P., Harrison, L. et al. The Climate Hazards Center Infrared Precipitation with Stations, Version 3. Sci Data 13, 718 (2026), data https://doi.org/10.15780/G2JQ0P.",
```
aldenks:
```suggestion
    "attribution": "UCSB Climate Hazards Center CHIRPS version 3.0 data processed by dynamical.org from the Climate Hazards Center data archive. Funk, C., Peterson, P., Harrison, L. et al. The Climate Hazards Center Infrared Precipitation with Stations, Version 3. Sci Data 13, 718 (2026), data https://doi.org/10.15780/G2JQ0P.",
```

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/precipitation_surface/zarr.json
```diff
@@ -0,0 +1,85 @@
...
+        "index_location": "end"
+      }
+    }
+  ],
+  "attributes": {
+    "long_name": "Precipitation rate",
+    "short_name": "prate",
+    "standard_name": "precipitation_flux",
+    "units": "kg m-2 s-1",
+    "comment": "Average precipitation rate over the 24 hours starting at the time coordinate. Units equivalent to mm/s. NaN where CHIRPS makes no estimate, which is the ocean and marginal seas; large lakes carry values.",
```
aldenks:
```suggestion
    "comment": "Average precipitation rate over the 24 hours starting at the time coordinate. Units equivalent to mm/s. NaN over oceans.",
```
make this equal to the other

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/zarr.json
```diff
@@ -0,0 +1,324 @@
+{
+  "attributes": {
+    "dataset_id": "ucsb-chc-chirps-analysis-preliminary",
+    "dataset_version": "0.1.0",
+    "name": "UCSB CHC CHIRPS analysis preliminary",
+    "description": "Daily precipitation from the Climate Hazards Center Infrared Precipitation with Stations (CHIRPS) version 3.0 preliminary product, a lower latency estimate that blends satellite precipitation with the station observations available in near real time and splits each pentad total into days using IMERG, so it differs from the final (rnl) product in its station inputs and daily distribution.",
```
aldenks:
```suggestion
    "description": "Daily precipitation from the Climate Hazards Center Infrared Precipitation with Stations (CHIRPS) version 3.0 preliminary product.",
```

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/zarr.json
```diff
@@ -0,0 +1,324 @@
...
+    "attribution": "UCSB Climate Hazards Center CHIRPS version 3.0 data processed by dynamical.org from the Climate Hazards Center data archive. Cite Funk, C., Peterson, P., Harrison, L. et al. The Climate Hazards Center Infrared Precipitation with Stations, Version 3. Sci Data 13, 718 (2026), data https://doi.org/10.15780/G2JQ0P.",
```
aldenks:
```suggestion
    "attribution": "UCSB Climate Hazards Center CHIRPS version 3.0 data processed by dynamical.org from the Climate Hazards Center data archive. Funk, C., Peterson, P., Harrison, L. et al. The Climate Hazards Center Infrared Precipitation with Stations, Version 3. Sci Data 13, 718 (2026), data https://doi.org/10.15780/G2JQ0P.",
```

src/reformatters/ucsb_chc/chirps/analysis_final/dynamical_dataset.py
```diff
@@ -0,0 +1,23 @@
...
+class UcsbChcChirpsAnalysisFinalDataset(UcsbChcChirpsAnalysisMaterializedDataset):
+    template_config: UcsbChcChirpsAnalysisFinalTemplateConfig = (
+        UcsbChcChirpsAnalysisFinalTemplateConfig()
+    )
+    region_job_class: type[UcsbChcChirpsAnalysisFinalRegionJob] = (
+        UcsbChcChirpsAnalysisFinalRegionJob
+    )
+
+    update_schedule: ClassVar[str] = "0 23 * * *"
+    validate_schedule: ClassVar[str] = "0 0 * * *"
```
aldenks: every 3rd day

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def validators(self) -> Sequence[validation.Validator]:
+        return (
+            validation.CheckCurrentData(
+                max_delay=self.region_job_class.expected_missing_window
+            ),
+            validation.CheckRecentNans(
+                # About 72% of grid cells are structurally NaN over ocean and marginal
+                # seas.
+                max_nan_fraction=0.05,
```
aldenks: how does this comment make sense w/ this max nan fraction?

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{self.dataset_id}-update",
+            schedule=self.update_schedule,
+            pod_active_deadline=timedelta(minutes=60),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="7",
+            memory="38G",
```
aldenks: lets see if we can squeak it into 31G   so it'll fit on a 32G node

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+    validate_schedule: ClassVar[str]
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{self.dataset_id}-update",
+            schedule=self.update_schedule,
+            pod_active_deadline=timedelta(minutes=60),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="7",
```
aldenks:
```suggestion
            cpu="3.5",
```
we could go faster but this is not a low latency dataset

src/reformatters/ucsb_chc/chirps/region_job.py
```diff
@@ -0,0 +1,169 @@
...
+    ) -> ArrayFloat32:
+        assert coord.downloaded_path is not None, "File must be downloaded first"
+        with rasterio.open(coord.downloaded_path) as reader:
+            data: ArrayFloat32 = reader.read(1, out_dtype=np.float32)
+
+        assert data.shape == (GRID_LAT_SIZE, GRID_LON_SIZE), (
+            f"unexpected source grid shape {data.shape} in {coord.get_url()}"
+        )
+        mask_source_fill_value_inplace(data, data_var.internal_attrs)
+        data *= np.float32(MM_PER_DAY_TO_KG_M2_S)
```
aldenks: nest this within an if data_var.name == ...

src/reformatters/ucsb_chc/chirps/region_job.py
```diff
@@ -0,0 +1,169 @@
...
+        }
+        times = self.template_ds.coords[self.append_dim].values
+        stop = self.region.start
+        for i in range(self.region.start, len(times)):
+            time = pd.Timestamp(times[i])
+            if time in read_times:
+                stop = i + 1
+            elif time not in self.known_missing_days:
+                break
+        return self.template_ds.isel({self.append_dim: slice(None, stop)})
```
aldenks: why is this and the whole known_missing_days set needed? the default behavior should do what we want: attempt to download all files, log an exception if one is missing, and on an update only extend the time dim through successfully processed timestamps. We don't care if we log.exception about a couple known missing files during the backfill, that's good fyi for us and silencing it is not worth keeping around a whole code mechanism.

src/reformatters/ucsb_chc/chirps/region_job.py
```diff
@@ -0,0 +1,169 @@
...
+    product: ChirpsProduct
+
+    download_parallelism: int = 8
+    # Days the source never published are skipped by the contiguity scan but still downloaded.
+    known_missing_days: ClassVar[frozenset[pd.Timestamp]] = frozenset()
+
+    def is_expected_missing(self, coord: UcsbChcChirpsAnalysisSourceFileCoord) -> bool:
+        return (
+            super().is_expected_missing(coord) or coord.time in self.known_missing_days
+        )
```
aldenks: cut i think (see update template w/ results comment)

src/reformatters/ucsb_chc/chirps/region_job.py
```diff
@@ -0,0 +1,169 @@
...
+        reformat_job_name: str,
+    ) -> tuple[
+        Sequence[
+            RegionJob[DataVar[BaseInternalAttrs], UcsbChcChirpsAnalysisSourceFileCoord]
+        ],
+        xr.DataTree,
+    ]:
+        existing_ds = xr.open_zarr(primary_store, chunks=None)
+        # Include the newest stored day so its shard is selected even when no later
+        # day is available.
```
aldenks: cut comment

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,235 @@
...
+        "total into days using ERA5."
+    ),
+    "preliminary": (
+        "Daily precipitation from the Climate Hazards Center Infrared Precipitation "
+        "with Stations (CHIRPS) version 3.0 preliminary product, a lower latency "
+        "estimate that blends satellite precipitation with the station observations "
+        "available in near real time and splits each pentad total into days using "
+        "IMERG, so it differs from the final (rnl) product in its station inputs and "
+        "daily distribution."
+    ),
```
aldenks, line 49: update these per my comments in the template jsons

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,235 @@
...
+    @property
+    def dataset_attributes(self) -> DatasetAttributes:
+        return DatasetAttributes(
+            dataset_id=f"ucsb-chc-chirps-analysis-{self.product}",
+            dataset_version="0.1.0",
+            name=f"UCSB CHC CHIRPS analysis {self.product}",
+            description=_DESCRIPTIONS[self.product],
+            attribution=(
+                "UCSB Climate Hazards Center CHIRPS version 3.0 data processed by "
+                "dynamical.org from the Climate Hazards Center data archive. Cite "
```
aldenks:
```suggestion
                "dynamical.org from the Climate Hazards Center data archive. "
```

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,235 @@
...
+                    short_name="prate",
+                    long_name="Precipitation rate",
+                    standard_name="precipitation_flux",
+                    units="kg m-2 s-1",
+                    step_type="avg",
+                    comment=(
+                        "Average precipitation rate over the 24 hours starting at "
+                        "the time coordinate. Units equivalent to mm/s. NaN where "
+                        "CHIRPS makes no estimate, which is the ocean and marginal "
+                        "seas; large lakes carry values."
```
aldenks:
```suggestion
                        "the time coordinate. Units equivalent to mm/s. NaN over oceans."
```

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,235 @@
...
+# CHIRPS v3.0 quasi-global 0.05 degree grid; pixel centers, rows north -> south.
+GRID_LAT_SIZE = 2400
+GRID_LON_SIZE = 7200
+_LAT_NORTH = 59.975
+_LAT_SOUTH = -59.975
+_LON_WEST = -179.975
+_LON_EAST = 179.975
+
+# mm/day -> kg m-2 s-1 (= mm/s): 1 mm/day of water is 1 kg m-2 per 86400 s.
+MM_PER_DAY_TO_KG_M2_S = 1.0 / 86400.0
```
aldenks: expand 86400.0 to a multiplication that is self explanatory (24 * 60 * 60)

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,235 @@
...
+        var_chunks: dict[Dim, int] = {
+            "time": 365,
+            "latitude": 50,
+            "longitude": 50,
+        }
+        var_shards: dict[Dim, int] = {
+            "time": var_chunks["time"],
+            "latitude": var_chunks["latitude"] * 24,
+            "longitude": var_chunks["longitude"] * 36,
+        }
```
aldenks: the 0.3 compression ratio used when calculating these is quite suspect. we usually get ~0.02 on hourly precipitation data -- daily will definitely have fewer 0s, but we get ~0.15 - ~0.25 even on the most dense and hard to compress variables (winds) so daily precip I'd expect is ~0.05 or so. recheck the expected compression ratio.

then redo these spatial chunk sizes. I imagine we can go much larger spatially while keeping compressed chunk sizes ~1-2.5MB

keep 365 days for time chunk/shard size

comment in this code should note estimated chunk and shard sizes uncompressed and compressed

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Low · **File:** `src/reformatters/ucsb_chc/chirps/analysis_preliminary/dynamical_dataset.py:22-23`

**Preliminary update cron fires one hour after CHC's observed publication time.**

`update_schedule = "0 17 * * *"`, but every preliminary file I checked landed at ~16:01 UTC (`Last-Modified` from HEAD requests, 2026-09-15):

| file | published (UTC) | lag from the day's timestamp |
|---|---|---|
| prelim 2026.08.26 (pentad 26–31 Aug) | Wed 02 Sep 16:01:10 | 7.7 days |
| prelim 2026.08.31 | Wed 02 Sep 16:01:16 | 2.7 days |
| prelim 2026.09.10 (pentad 6–10 Sep) | Sat 12 Sep 16:01:47 | 2.7 days |

So the daily update catches a new pentad with ~59 minutes to spare. Any slip at CHC past 17:00 delays ingestion by a full day. That does not break `CheckCurrentData` (the first day of a 6‑day pentad is due at 10 days and publishes at ~7.7, so one missed day still passes), but it eats most of the 2.3‑day slack the 10‑day window leaves, and `validate` at 18:00 would then only ever see a store one pentad behind.

**Suggested change:** move the preliminary update later, e.g. `"0 19 * * *"` with validate at `"0 20 * * *"` (keeps validate = update start + 60‑minute deadline). The PR description's "two days after the pentad ends" is right; the schedule just needs to sit comfortably after the ~16:00 UTC drop.

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Nit · **File:** `src/reformatters/ucsb_chc/chirps/analysis_final/region_job.py:11-13`

**"Monthly batches arrive in the third week of the following month" does not match the observed publication dates.**

`Last-Modified` from HEAD requests on 2026-09-15:

| final file | published (UTC) | lag from day 1 of that month |
|---|---|---|
| rnl 2026.07.01 | Fri 14 Aug 21:33 | 44 days |
| rnl 2026.08.01 | Fri 11 Sep 22:04 | 41 days |
| rnl 2026.08.31 | Fri 11 Sep 22:04 | 11 days |

Both drops were in the second week (the 11th and 14th), not the third. The 60‑day `expected_missing_window` / `max_delay` still has ~16 days of margin, and stays comfortable if the final update moves to every third day as requested elsewhere on this PR.

**Suggested change:** either drop the sentence (the value speaks for itself) or state the observed fact: the whole month is published in one batch roughly 40–45 days after its first day.

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Low · **File:** `src/reformatters/ucsb_chc/chirps/region_job.py:117-137` and the PR description ("Operational updates" section)

**Risk note on dropping the custom trim in favour of the base `update_template_with_results` (the change requested in the `known_missing_days` thread).** I agree with the request; this is the one behaviour that changes and that the PR description currently promises the opposite of.

The base default trims to the newest *successfully read* day. `assert_no_append_dim_retraction` only compares append-dim lengths. So in a replayed trailing shard:

- A day whose re-download fails **after** the newest success (or the newest stored day itself) → template shorter than the store → loud `ValueError`, main untouched. Same as today.
- A day whose re-download fails **before** a later success → the shard is written with NaN at that day, over previously-good stored values, and the run commits. Nothing alerts: `CheckRecentNans(append_dim_window=3)` sees only the newest three positions.

Within a year the second case self-heals on the next daily replay, so it is a one-day blip. At the year boundary it does not: the run that first extends into January is the last one that replays the previous year's shard, and a 404/5xx that survives obstore's 16 retries in that run leaves a permanent NaN day with no signal. `nasa-imerg` already lives with exactly this (it uses the base default), so this is the repo convention, not a CHIRPS defect. It is also the case the PR's `test_update_rejects_a_failed_reread_without_changing_the_icechunk_store` currently pins and would need to be re-scoped to "trailing failure only".

**Suggested change:** accept it as convention, and:
1. Rewrite the description's "template extends only through the contiguous successfully read prefix… unexplained gaps stop advancement… a failed reread… Icechunk finalization rejects the update" to what the default actually guarantees: the store extends through the newest read day; a failure at or after the newest stored day fails the update; an earlier failure is written as NaN and refilled on the next replay of that shard.
2. Optionally, make the year-boundary case visible instead of silent: `CheckRecentNans(append_dim_shards=1, ...)` checks every position of the trailing shard and reads only `sampled_points` chunks in one pass, so the cost is 40 chunk reads. The trade-off is that a future permanent CHC gap fails validation until the year rolls over, which is arguably the alert you want.

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Low · **File:** `src/reformatters/ucsb_chc/chirps/template_config.py:232` vs PR description ("with metadata matching the existing `precipitation_surface` in `noaa-mrms` and `nasa-imerg`")

**`keep_mantissa_bits=8` does not match either sibling; both use 7.**

- `nasa/imerg/template_config.py:236`: `keep_mantissa_bits=7`
- `noaa/mrms/conus_analysis_hourly/template_config.py:209`: `default_keep_mantissa_bits = 7`, used for `precipitation_surface`

AGENTS.md gives two rules that disagree here: the table says a precipitation flux/rate gets 8, and the same paragraph says to match an existing equivalent variable rather than re-derive. Whichever you pick, the description's "matching" claim is only true of the attrs, not the rounding, and the new datasets will be the odd ones out on precision.

**Suggested change:** either set 7 to match the two published `precipitation_surface` variables (this is a new store, so no compatibility cost), or keep 8 and say in the description that CHIRPS deliberately keeps one more bit than IMERG/MRMS. One line either way.

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Low (scope) · **File:** uncommitted `src/reformatters/common/kubernetes.py` (`_parse_schedule` / `previous_fire_time`) and `tests/common/kubernetes_test.py`

**The day-of-month parser extension is not required for `"0 23 */3 * *"` to run.** Kubernetes accepts the schedule as-is; the only thing in this repo that parses a schedule string is `CronJob.previous_fire_time`, and it has exactly two callers:

- `DynamicalDataset._operational_update_jobs` (`dynamical_dataset.py:632`), only when the region job's `operational_update_jobs` declares `job_fire_time` (today: WeatherNext2 only).
- `DynamicalDataset._virtual_poll_deadline` (`dynamical_dataset.py:804`), virtual datasets only.

`UcsbChcChirpsAnalysisFinalRegionJob` is materialized and does not declare `job_fire_time`, so the parser is never reached for it at runtime, and no test sweep (`tests/datasets_test.py`, `tests/common/dynamical_dataset_test.py`) calls `previous_fire_time` on registered datasets. I also could not find a doc that constrains the schedule shape.

**Suggested change:** drop the `kubernetes.py` / `kubernetes_test.py` change from this PR and keep the schedule string alone. If you would rather keep the generalization so `previous_fire_time` cannot assert on this schedule if a fire-time consumer is added later, that is a reasonable choice, but then it is a deliberate shared-code change, not a prerequisite, and the PR description should say so.

For the record, if kept, the logic is correct: I probed month boundaries directly (`0 23 */3 * *`): now 2026-10-01T22:00 → 2026-09-28 23:00 (day 30 correctly skipped, day 31 absent), 2026-10-01T23:00 → 2026-10-01 23:00, 2027-03-01T22:59 → 2027-02-28 23:00, 2026-09-01T22:00 → 2026-08-31 23:00 (consecutive-day fire, matching cron's `1,4,…,31` expansion of `*/3`). The hourly walk-back is bounded at 744 steps for `*/31`.

aldenks:
**Reviewer:** refmt-chirps-pr1000-claude · **Severity:** Nit · **File:** uncommitted `src/reformatters/common/kubernetes.py` (`_parse_days_of_month`) — only applies if the parser extension is kept

**`_parse_days_of_month` duplicates `_parse_hours`' range/step logic.** One field parser with the field's bounds, e.g. `_parse_field(field, first, last)` used as `_parse_field(hours, 0, 23)` and `_parse_field(days_of_month, 1, 31)`, removes the new function and keeps one way of expanding a cron field; the regex still restricts days-of-month to `*` / `*/N`, so it does not widen what is accepted.

Two validation edges, both pre-existing for hours and now inherited by the new field, worth a single `assert` if you touch this: `*/0` raises `ValueError: range() arg 3 must not be zero` instead of the parser's `AssertionError` (and is not in `test_previous_fire_time_rejects_unsupported_schedule`), and `*/N` with `N > 31` silently collapses to day 1 (`"0 23 */40 * *"` → fires on the 1st only), which matches cron but hides a typo.

docs/ops_card.md
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed the CHIRPS-specific ops-card entry together with the accepted-date mechanism.

src/reformatters/common/materialized_region_job.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Renamed to `expected_unavailable_window` in the shared base, CHIRPS subclasses, validator configuration, tests, and implementation guide. The shared default remains 48 hours.

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/precipitation_surface/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Applied the requested precipitation comment in the shared configuration and regenerated both templates.

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Applied this exact final-product description and regenerated the template.

src/reformatters/ucsb_chc/chirps/analysis_final/templates/latest.zarr/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed “Cite ” from the shared attribution and regenerated both templates.

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/precipitation_surface/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Applied the same requested precipitation comment to both products through their shared configuration and regenerated the templates.

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Applied this exact preliminary-product description and regenerated the template.

src/reformatters/ucsb_chc/chirps/analysis_preliminary/templates/latest.zarr/zarr.json
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed “Cite ” from the shared attribution and regenerated both templates.

src/reformatters/ucsb_chc/chirps/analysis_final/dynamical_dataset.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Changed final updates to `0 23 */3 * *` (23:00 UTC on every third day of the month). Validation remains daily at 00:00 UTC, so it follows those updates rather than running before them on the same restricted date.

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Replaced the ocean-fraction comment with the relevant denominator: points that are always NaN in the sampled window are excluded. The 5% threshold applies to the remaining points; 40 samples help find land. The existing test covers ocean masking and a missing land day.

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Set the update memory request to `50G` per your latest instruction, retaining `25.5G` shared memory. Added a tuning reminder in the PR description: measure peak memory on the first full-shard run, then tune the request back down. The crons remain suspended.

src/reformatters/ucsb_chc/chirps/dynamical_dataset.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Reduced the update CPU request to `3.5`. The 60-minute deadline remains; the description calls for measuring full-shard runtime before enabling the suspended jobs.

src/reformatters/ucsb_chc/chirps/region_job.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Scoped the unit conversion to `data_var.name == "precipitation_surface"`; source-fill masking remains generic.

src/reformatters/ucsb_chc/chirps/region_job.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed `known_missing_days` and the custom trim. Updates now use the shared base behavior: advance through the newest successfully read day, leaving interior failures as NaN. Tests cover gap recovery, an older stored value becoming NaN after a failed reread, and atomic rejection when the proposed time dimension retracts. The description now states these guarantees and the need for targeted backfill once a damaged shard leaves the replay window.

src/reformatters/ucsb_chc/chirps/region_job.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed the CHIRPS override along with the accepted-date set. The shared logging logic retains the publication-lag window; old missing files log exceptions.

src/reformatters/ucsb_chc/chirps/region_job.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed the comment; selection still includes the newest stored day so its shard is replayed.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Updated both descriptions to the concise wording requested on the generated templates, then regenerated those templates.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Removed “Cite ” from the shared attribution.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Replaced the variable comment with the exact requested wording and regenerated both templates.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Changed the conversion constant to `1.0 / (24 * 60 * 60)`.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d). Remeasured all 365 daily final/rnl files for 2025 through the production conversion, eight-bit rounding, and Blosc codec. Chose `(365, 60, 60)` chunks: 50 wet Amazon/Congo samples compressed to 1.105–1.728 MB (median 1.563 MB); 75×75 reached 2.693 MB and 100×100 reached 4.788 MB. The wet-land ratio remains around 0.3, so 0.05 would undersize the estimate there. Kept `(365, 1200, 1800)` shards and documented raw/compressed chunk and estimated shard sizes. The description has the measured comparison and explicitly limits it to final/rnl 2025. Owner verified actual Zarr encoded bytes and round trips, plus fresh source matches in January, July, and December.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673238118), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Moved preliminary updates to 19:00 UTC and validation to 20:00 UTC. Additional source checks found winter files published after 17:00 UTC (for example, 2025-11-05 was published on November 7 at 17:07 UTC), strengthening the case for the change. The description now states that observation; the 10-day freshness window remains.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673238257), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Removed the inaccurate final-publication comment, and also removed the preliminary comment that stated a fixed two-day lag. The publication windows remain 60 and 10 days; the observed batches still meet those deadlines under the selected schedules.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673238374), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Removed the custom trim and accepted-date set as requested. Updated the description, shared processing documentation, and Icechunk tests to cover both interior failed rereads becoming NaN and rejected extent retraction. The precise guard is proposed end < stored end: even failure of the newest stored day can commit when a later new day succeeds. Kept the newest-three-day NaN check rather than adding historical-gap policy; a wider check could detect older failures, but persistent source gaps would then keep failing validation. The description explicitly states that older gaps can require targeted backfill once their shard leaves the replay window.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673238493), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Kept eight mantissa bits under the repository's precipitation-rate convention. Corrected the description to say the names and physical-quantity attributes follow IMERG/MRMS, while explicitly noting that those siblings retain seven bits. Their published encodings are unchanged.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673396625), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Kept the parser extension because `tests/common/deploy_test.py::test_registered_dataset_schedules_are_parseable` calls `previous_fire_time` for every registered cron and fails on the new `*/3` schedule without it. Your materialized-runtime distinction is correct; the description now explains the actual requirement. Your follow-up withdrew this objection after checking that test.

aldenks:
Resolution of [review comment](https://github.com/dynamical-org/reformatters/pull/1000#issuecomment-5673396747), in [b6ec76ef](https://github.com/dynamical-org/reformatters/commit/b6ec76ef4d76877dbdf4fc8846592153a95d929d):

Consolidated hours and days into one bounded `_parse_schedule_field` helper and added the skipped-month-end regression (October 1 before the scheduled fire resolves to September 28). Kept oversized positive steps valid, matching cron semantics; zero steps already fail. Codex confirmed that behavior against the upstream parser, and your follow-up withdrew the proposed upper bound.

src/reformatters/common/materialized_region_job.py
```diff
@@ -305,15 +308,14 @@ def _call_download_file(coord: SOURCE_FILE_COORD) -> SOURCE_FILE_COORD:
             except Exception as e:
                 updated_coord = replace(coord, status=SourceFileStatus.DownloadFailed)
 
-                # For recent files, we expect some files to not exist yet, just log the path
-                # else, log exception so it is caught by error reporting but doesn't stop processing
+                # Recently unavailable files log quietly; other failures reach error reporting.
```
aldenks:
```suggestion
                # Recent unavailable files log quietly; older file failures reach error reporting.
```

src/reformatters/ucsb_chc/chirps/analysis_final/dynamical_dataset.py
```diff
@@ -0,0 +1,23 @@
...
+class UcsbChcChirpsAnalysisFinalDataset(UcsbChcChirpsAnalysisMaterializedDataset):
+    template_config: UcsbChcChirpsAnalysisFinalTemplateConfig = (
+        UcsbChcChirpsAnalysisFinalTemplateConfig()
+    )
+    region_job_class: type[UcsbChcChirpsAnalysisFinalRegionJob] = (
+        UcsbChcChirpsAnalysisFinalRegionJob
+    )
+
+    update_schedule: ClassVar[str] = "0 23 */3 * *"
+    validate_schedule: ClassVar[str] = "0 0 * * *"
```
aldenks: Validate schedule should be update schedule+update pod active deadline

src/reformatters/ucsb_chc/chirps/template_config.py
```diff
@@ -0,0 +1,228 @@
...
+            "longitude": np.linspace(_LON_WEST, _LON_EAST, GRID_LON_SIZE),
+        }
+
+    @computed_field
+    @property
+    def dataset_attributes(self) -> DatasetAttributes:
+        return DatasetAttributes(
+            dataset_id=f"ucsb-chc-chirps-analysis-{self.product}",
+            dataset_version="0.1.0",
+            name=f"UCSB CHC CHIRPS analysis {self.product}",
```
aldenks:
```suggestion
            name=f"UCSB CHC CHIRPS analysis, {self.product}",
```

src/reformatters/common/materialized_region_job.py
aldenks: Addressed in [1006266c](https://github.com/dynamical-org/reformatters/commit/1006266c3993e7acdba4218a768de6b79839e2cc). Applied the logging comment exactly as suggested. Logging behavior is unchanged: recent not-found responses log quietly, while older missing files and other failures still reach error reporting.

src/reformatters/ucsb_chc/chirps/analysis_final/dynamical_dataset.py
aldenks: Addressed in [1006266c](https://github.com/dynamical-org/reformatters/commit/1006266c3993e7acdba4218a768de6b79839e2cc). Final updates now run at 00:00 UTC (`0 0 */3 * *`) and validation at 01:00 UTC on the same dates (`0 1 */3 * *`), exactly the update pod’s 60-minute active deadline later. Moving both fires onto the same date avoids month-end errors from shifting a stepped day-of-month schedule across midnight. Added regression coverage for both products across short-month, leap-year, and other month boundaries.

src/reformatters/ucsb_chc/chirps/template_config.py
aldenks: Addressed in [1006266c](https://github.com/dynamical-org/reformatters/commit/1006266c3993e7acdba4218a768de6b79839e2cc). Applied `UCSB CHC CHIRPS analysis, final` and `UCSB CHC CHIRPS analysis, preliminary` in the shared template configuration, updated the existing name assertions, and regenerated both checked-in templates.

PR #1017, opened 2026-09-06
Add the NOAA GEFS 16 day 0.5 degree virtual forecast dataset

aldenks:
Adds `noaa-gefs-forecast-16-day-0-5-degree-virtual`: every message the 0.5 degree
`pgrb2a` and `pgrb2b` files publish, for all 31 ensemble members, over the 105 lead
times from 0 to 384 hours (3 hourly through 240, 6 hourly after). Stacked on #993.

This is the first GEFS virtual dataset built from the a and b products, on the 0.5
degree grid, and with vertical groups, so most of the work extends the shared
`virtual_template_config.py` rather than only subclassing it.

## Variables — 268

| group | arrays | levels |
| --- | --- | --- |
| root | 247 | — |
| `pressure_level` | 12 | 31 |
| `model_level` | 6 | 4 (hybrid) |
| `height_above_mean_sea_level` | 3 | 8 |

229 of the 270 candidate arrays matched an existing variable in
`noaa-gfs-analysis-virtual` on `(grib element, idx level)` and take its name and
metadata verbatim.

The a and b products **partition** the message set: at every lead sampled the two
indexes have zero messages in common, and neither carries a whole vertical level set
on its own. A variable therefore names the *set* of files its messages live in
(`NoaaGefsVirtualInternalAttrs.source_file_types`), and file presence is probed
through a root variable, whose chunk needs no level pick.

## Manifest splitting

Splits: root 8, `pressure_level` 4, `model_level` 4,
`height_above_mean_sea_level` 5 inits.

| group | split | refs per array manifest | bytes at ~16.4 B/ref |
| --- | --- | --- | --- |
| root | 8 | 26,040 (25,792 for a windowed variable) | 413 KiB |
| `pressure_level` | 4 | 403,620 | 6.3 MiB |
| `model_level` | 4 | 52,080 | 834 KiB |
| `height_above_mean_sea_level` | 5 | 130,200 | 2.0 MiB |

**Refs per commit: 11.2M** (11,224,356) counting each variable's actually published
levels; 12.0M (11,966,496) if every declared level were filled. Under the ~12M the operational HRRR forecast 48
hour store sustains at p50 2.8 s. The smallest manifest is a root array at
8 × 105 × 31 = 26,040 refs, 26× icechunk's 1,000-ref `min_num_chunks` floor, so every
manifest gets location compression. The 16.4 B/ref figure is carried from HRRR-virtual
and should be re-measured with `repo.list_manifest_files` once the backfill has
written data.

One consequence worth a look before backfilling: at 8,665 inits (2020-10-01 to
2026-09-06) the full archive holds **M = 311,953 manifests**, against 20,596 for the
10 day virtual store and 9,512 for HRRR forecast 48 hour. Refs per commit is the
sizing metric these splits were chosen against and they meet it exactly; the manifest
count is the other axis, and icechunk's per-commit `manifest_info` scan is linear in M.

## Operations

Cron jobs ship `suspend=True`.

Publish lags measured over the eight cycles of 2026-09-03 and 2026-09-04 (all 6,510
a and b files of each cycle, from S3 `last_modified`):

- first file: init + 3h46m to 3h56m
- last member's f384: init + 6h31m to **7h11m** (the 18z cycles are the slow ones)

So the update fires at `45 3,9,15,21` (init + 3h45m, just before the burst) with a
4 hour `pod_active_deadline`, polling to init + 7h45m — 34 minutes past the slowest
observed end, and finishing two hours before the next cycle's fire. Validation runs at
`55 7,13,19,1`, the update's fire plus its deadline plus 10 minutes.

No `tick_interval` override (the loop self-paces) and no completeness gate in
`discover_available`: a forecast publishes refs as files arrive, and a partly filled
newest init is what readers want. The leading-edge gate stays an analysis convention.

## Source facts established here

- Inventory is **stable across every era**: the a file is 85 messages (71 at f000) and
  the b file 505 (473 at f000) at every date from 2020-10-01 to 2026-09-01. The single
  change is the cloud mixing ratio element, spelled `CLWMR` through 2025-12-18 and
  `CLMR` from 2025-12-19; both are matched. No variable needs an `available_from`.
- `SNOWC` and `SNOHF` appear at f000 with a degenerate `0-0 day ave fcst` window and
  are declined there, like `TMAX`/`TMIN` in the 10 day store.
- Three instantaneous fields (`PRES` at convective cloud bottom and top, `TCDC` in the
  convective cloud layer) are simply absent at f000 and carry `hour_0=False`.
- `ICEC` decodes to exactly {0, 1} in five samples spanning 2020–2026, both seasons,
  control and perturbed members, leads 0–384. It is a binary mask here, so it carries
  HRRR's `flag_values`/`flag_meanings` rather than GFS's area-fraction wording.
- `ICSEV` codes are GRIB2 code table 4.207, whose figures are **not** in severity
  order: `none light moderate severe trace heavy`.

## Metadata equivalence sweep

Run by `tests/common/datasets_cf_compliance_test.py`, which compares every variable,
short name, long name and coordinate across all datasets. What it caught:

- **`SNOWC` is a percentage in GEFS and a fraction everywhere else.** HRRR and AIFS
  publish `surface_snow_area_fraction` in `1`, which is also its CF canonical unit.
  Ours now applies a percent-to-fraction filter and is named
  `average_snow_area_fraction_surface`, matching HRRR's `snow_area_fraction_surface`
  bar the window prefix. Verified against data: the Greenland ice sheet reads 1.0.
- **GEFS spells low/medium/high cloud cover `TCDC` at a cloud-layer level; GFS uses
  `LCDC`/`MCDC`/`HCDC`.** Same quantity, so ours takes GFS's names
  (`average_low_cloud_cover` etc.) and short names `lcc`/`mcc`/`hcc`.
- `model_level` already existed in the HRRR virtual stores as
  `Hybrid model level number` / `model_level_number` / `int64`; ours matches exactly
  rather than inventing a second spelling.
- `NCPCP` gets ECMWF's canonical `ncpcp` / `Large scale precipitation (non-convective)`
  and CF's `stratiform_precipitation_amount` (CF has no `large_scale_*` name).
- `BRTMP` stays in kelvin: `toa_brightness_temperature` is canonically K, unlike the
  air temperatures we convert to Celsius.
- `ACPCP` follows the GFS catalog's current `cp` / `Convective precipitation`, which it
  moved to while this branch was in flight.

A second sweep keyed on `standard_name` rather than variable name — the keying that
would have caught `volumetric_soil_moisture_0_10cm` against AIFS's
`volumetric_soil_moisture_layer_1` — reports 68 standard names here and finds nothing
further. Every name this dataset introduces is a level or quantity no other dataset
serves: the isentropic families (`temperature_320k`, `wind_u_320k`,
`potential_vorticity_310k`, ...), `pressure_mean_sea_level` (a third MSL product,
distinct from both reductions in the data, not a renaming of either),
`cloud_cover_475mb`, `large_scale_precipitation_surface`,
`brightness_temperature_top_of_atmosphere`, and the 30-0 mb layer fields, whose
siblings at that layer already use the same suffix. `average_snow_area_fraction_surface`
differs from HRRR's `snow_area_fraction_surface` only by the window prefix the catalog
already uses for averaged variants.

The step-type divergences added to `CROSS_DATASET_CONSISTENCY_EXCEPTIONS` are the
already-accepted GEFS-vs-HRRR ones (GEFS publishes window averages where HRRR
publishes instantaneous values), plus `tcc`, where `TCDC` names both the column total
and a layer fraction within one dataset — the same exception `noaa-gfs-analysis-virtual`
already carries.

## Settled decisions, with two notes

Surface geopotential height is dropped, as decided. Worth recording for later: it is
not actually lead-0-only across the pair — the a file carries `HGT:surface` at f000 and
the b file carries it at every lead from f003, so the two together span the whole lead
axis and it could be added as an ordinary variable. Adding a variable is a
non-breaking change, so nothing here forecloses it.

The `pressure_level` dimension is the settled 31 levels. The source publishes exactly
one isobaric message outside them, ozone at 125 mb, which is not served; every other
isobaric message lands on one of the 31. Variables that skip levels within the 31
(relative humidity, absolute vorticity and cloud mixing ratio above 10 hPa, ozone below
400 hPa, the two icing fields outside 800–300 hPa) say so in their `comment`.

`keep_mantissa_bits` is `"no-rounding"` for every variable, as in every other virtual
store: the chunks are the source bytes and are never rewritten, so the float32 rounding
convention has nothing to act on.

## Tests

- `tests/.../template_config_test.py` — structure, group membership, the a/b level
  partition, window wording, flag codes and the lead grid's step up at 240 hours.
- `tests/.../region_job_test.py` — coord generation, both-products requests, lead 0,
  the root-variable probe, and 18 real archived indexes (both products, both element
  spellings, leads 0/6/120/240/246/384, control and perturbed) asserting that every
  message a file offers is taken exactly once and nothing requested is absent.
- `tests/.../dynamical_dataset_test.py` — a real local backfill and operational update
  with snapshot values for every variable it processes, at six lead times and two
  members, plus the cron, validator and manifest-split arithmetic.

`uv run pytest -m "not slow" -n 4` passes (2,649). Every module holding slow tests in
`tests/noaa/gefs/` and `tests/common/virtual*` passes, run one path per invocation.
`ruff format`, `ruff check` and `ty check` are clean.

---

Carrying #991's review over. Both bounds were round numbers; both are now derived from the
publication lags this PR measured.

### `CheckCurrentData` — `max_delay` 12h → **7h50m**

| | |
|---|---|
| `pod_active_deadline` | **4h** |
| worst-case moment init *T* becomes present | init + 7h45m |
| validate cron fires | init + **7h55m** |

- **Lower bound, > 7h45m.** The binding instant is not the validate fire — at the fire a healthy
  store already holds init *T*. It is the last instant at which *T* is legitimately still absent:
  the end of its own update run. Just before init+7h45m the newest position present is *T*−6h, so
  anything ≤ 7h45m false-alarms on a healthy store.
- **Upper bound, ≤ 7h55m.** For the validation run that *follows* a missed cycle to catch it, *T*
  must already be due when that run fires. Beyond 7h55m the miss goes unreported for six more hours.
- **Chosen 7h50m** — 5 min above the lower bound, 5 min below the hard bound.

`test_validators` drives the real `CheckCurrentData` and pins that **the first** stalled cycle
alerts (was: the second).

### `CheckVirtualManifestCompleteness` — `(0.95, 1.0)` → **`(1.0,)`**

In files, since that is what the check counts:

| | |
|---|---|
| expected source files per init | 105 leads × 31 members × 2 products = **6,510** |
| last member's f384 lands | init + **7h11m** |
| poll window closes | init + 7h45m |
| validate fires | init + **7h55m** |
| `0.95` silently tolerated | 6510 − ⌈0.95 × 6510⌉ = **325 missing files** |

The whole burst closes 44 minutes before validation fires, and the update polls through it. Nothing
is left for a leading tier to excuse: every init the store reached must be whole.

## Cron job naming

The dataset id plus `-validate` is 53 characters against kubernetes' 52. The cron names now drop
`-degree` and **keep** the dashed resolution, rather than borrowing the provider's `p` spelling:

| | before | after | len |
|---|---|---|---|
| update | `noaa-gefs-forecast-16-day-0p5-virtual-update` | `noaa-gefs-forecast-16-day-0-5-virtual-update` | 44 |
| validate | `noaa-gefs-forecast-16-day-0p5-virtual-validate` | `noaa-gefs-forecast-16-day-0-5-virtual-validate` | **46** |

Checked in code — `CronJobName` is `max_length=52` and
`test_cron_job_names_fit_the_kubernetes_limit` asserts it. The manual-workflow choices were
regenerated by the `generate-manual-workflows` step.

URLs, and `FILE_RESOLUTIONS`.

## Metadata

- **Dataset description** cut to the one sentence the published materialized siblings already use,
  byte-for-byte. **Manifest split comment** reduced to a single orienting line. The accumulation
  window comment follows the shared catalog trimmed in #993.
- **`haines_index_surface` now declares `flag_values` / `flag_meanings`** (separate commit). This
  was a pre-existing gap in this PR, not a consequence of the changes above: the variable carries
  `flag_values=(2, 3, 4, 5, 6)` and `flag_meanings="very_low_potential very_low_potential
  low_potential moderate_potential high_potential"` on `noaa-gfs-analysis-virtual` and
  `noaa-gfs-forecast-virtual` (both already on `main`), and this catalog copied the GFS `comment`
  verbatim while dropping the two flag attrs, so
  `datasets_cf_compliance_test::test_metadata_consistency_across_datasets` failed at this PR's tip.
  Matched verbatim rather than reworded, per AGENTS.md. The repeated `very_low_potential` is
  correct — the Haines index runs 2–6 and 2–3 are both very low potential — and gives one
  blank-separated label per flag value as CF §3.5 requires. The fix lands in the shared 0.5°
  catalog, so it carries up to #1019 on rebase (confirmed in that PR's regenerated template).

- **`total_ozone_atmosphere` and the two gravity wave stresses now follow the GFS catalog**
  (separate commit), which `main` corrected in #1021 while this PR was open. Total ozone is
  published in metres under
  `standard_name="equivalent_thickness_at_stp_of_atmosphere_ozone_content"`, decoded from the
  source's Dobson units by a `ScaleOffset` filter of scale 1e5 (1 DU = 1e-5 m; the filter decodes
  as `encoded / scale`). `eastward_` / `northward_gravity_wave_surface_stress` drop CF's
  `atmosphere_*ward_stress_due_to_gravity_wave_drag`, whose sign convention the NCEP physics
  inverts, and state that convention in their comments instead — declaring a CF name we invert is
  worse than declaring none. Both names were already in `main`'s
  `ALLOWED_MISSING_STANDARD_NAME`, so no allowlist edit was needed here.

  **Only the sign sentence is copied from GFS.** Its window prefix is analysis wording; these are
  forecasts, so the window half comes from our own `window_comments` as it does for every other
  windowed variable in this catalog. `total_ozone_atmosphere` carries no window prefix and so is
  byte-identical to GFS. Verified on the regenerated templates, not assumed.

## Tests

`file_refs` reads the GRIB header at the index's last offset to reject a stale index (#1021), so the
`fake_download` closures that stubbed only the index download reached the real S3 object and failed
in CI. They now go through `tests/noaa/grib_index_fixtures.stub_grib_source_file_reads`, the helper
#1021 established, which stubs both reads — no new stub machinery, and none was needed: the helper
already took `data_file_size` for the single-entry case that gives it no next offset to derive a
length from.


src/reformatters/noaa/gefs/forecast_16_day_0_5_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,89 @@
...
+        cron_job_name_prefix = self.dataset_id.replace("-0-5-degree", "-0-5")
+        # A run publishes ~init+3h46m through ~init+7h11m.
+        # Fire just before the first files become available and stop 30m after expected completion.
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{cron_job_name_prefix}-update",
+            schedule="45 3,9,15,21 * * *",
+            pod_active_deadline=timedelta(hours=4),
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="4",
```
aldenks: 3.5 to fit on 4 cpu node

PR #1018, opened 2026-09-06
Correct GFS virtual variable metadata from the validation passes

aldenks:
Metadata-only corrections to the shared GFS virtual variable catalog, from two
validation passes of `noaa-gfs-analysis-virtual` and the independent reviews of both.
No stored value changes and no change to which source message any chunk references, so
neither store needs re-ingestion. `virtual_template_config.py` is shared, so most of
these land in both `noaa-gfs-analysis-virtual` and `noaa-gfs-forecast-virtual`; the last
one is analysis-only.

## The fixes

1. **Convective cloud pressure pair** — `pressure_convective_cloud_bottom` and
   `pressure_convective_cloud_top` now say NaN means the source does not provide a
   convective cloud base/top pressure at that cell, which can happen even where
   `convective_cloud_cover` is nonzero. They previously shared the eight layer-cloud
   variables' "near zero cover" wording, which is false for this pair; the eight are
   unchanged.
2. **`haines_index_surface`** — declares `flag_values=(2,3,4,5,6)` and the matching
   NOAA labels (2 and 3 are both "very low potential"), making the enumerated ordinal
   self-describing.
3. **`soil_type_surface`** — drops `standard_name="soil_type"`. CF defines that name as
   a flag variable, and GFS's interpolated fractional class numbers are neither integer
   codes nor flag values. No flags added, because the values genuinely are not codes.
4. **`albedo_surface`** — says the value is exactly 0 wherever the averaging window
   received no sunlight and that those zeros must be excluded from a time-mean albedo.
   `fill_value` stays NaN.
5. **The eight soil-moisture variables** — say the near-1 values over permanent land ice
   are placeholders rather than soil moisture measurements, and end with
   `Mask values >= 0.9.`
6. **`ventilation_rate_planetary_boundary_layer`** — says source precision can be as
   coarse as 1,000 m2 s-1, so a decoded 0 is a small value rounded down rather than
   missing data.
7. **The two UV-B fluxes** — replace the 280-315 nm band, which the source does not
   support, with the source's actual band, 263-345 nm. See Source verification below.
8. **The five instantaneous variables the source does not publish at lead 0**
   (`instantaneous_precipitation_convective_surface`,
   `potential_evaporation_rate_surface`, `pressure_convective_cloud_bottom`,
   `pressure_convective_cloud_top`, `convective_cloud_cover`) — the analysis says their
   00/06/12/18 UTC values come from the previous forecast cycle. This is the documented
   shortest-published-lead contract rather than a defect, but it is not obvious, and it
   is why instantaneous convective precipitation can exceed the total precipitation rate
   at those hours.
9. **`storm_relative_helicity_3000_0m`** — says the values use a right-moving storm
   motion in both hemispheres, so southern hemisphere values are positive-mean rather
   than mirrored. See Source verification below.

## Deliberately not included

Changes that would alter stored values, change which source message a chunk references,
or that need the dataset owner's ruling:

- **`sunshine_duration_surface` lead selection** — selecting leads 1-6 would change
  which source message the synoptic-hour chunks reference. Owner ruling plus
  re-ingestion.
- **`total_ozone_atmosphere` units** — moving from `DU` to a CF-recognizable scaled
  length changes decoded values.
- **Previous-cycle instantaneous fields, coverage vs coherence** — fix 8 documents the
  current contract; whether the analysis should instead prefer same-cycle coherence at
  the synoptic hours is the owner's contract decision.
- **Deep-soil precision change** (`liquid_volumetric_soil_moisture_100_200cm`,
  2022-02-28 to 2022-07-03) — a bounded source-data fact for the validation report's
  Review notes, not static template metadata.
- **Stale-index guard fix** — pending the owner's ruling.

## Source verification

Fixes 7 and 9 were held back in the first pass of this PR because no reviewer had
consulted the authoritative source. Both have since been settled by reading NOAA UPP and
the forecast-model radiation code, and the two fixes are now included here. Full findings
with citations are in `.upp_source_check.md` in the `gfs-virtuals` worktree
(`/home/alden/code/dynamical/worktrees/gfs-virtuals/.upp_source_check.md`).

**Version provenance.** NOAA `global-workflow` `release/gfs.v16.3.34` pins UPP via
[`Externals.cfg` L24-29](https://github.com/NOAA-EMC/global-workflow/blob/0ae15e62ef9ba55eca80889e0edea415ce889f69/Externals.cfg#L24-L29)
to tag `upp_v8.3.0`, which dereferences to commit `c5f3053`. The forecast-model pin
resolves through `ufs-weather-model` `GFS.v16.3.26` to `fv3atm` `9a9a701` and NCAR
`ccpp-physics` `01ed01f`.

**UV-B band: 263-345 nm, not 280-315 nm.** UPP relays the diagnostic rather than
computing it: [`INITPOST_NETCDF.f` L2827-2836](https://github.com/NOAA-EMC/UPP/blob/fe4f02b41fd789e414fb4bf23ad27c58ef058ee4/sorc/ncep_post.fd/INITPOST_NETCDF.f#L2827-L2836)
reads `duvb_ave`/`cduvb_ave`, which
[`GFS_diagnostics.F90` L243-268](https://github.com/NOAA-EMC/fv3atm/blob/9a9a701333305131fa7fd20ed32b8c3cf8570399/gfsphysics/GFS_layer/GFS_diagnostics.F90#L243-L268)
maps to radiation slots 21 and 22, filled from `scmpsw%uvbfc`/`uvbf0` in
[`GFS_radiation_driver.F90` L2073-2076](https://github.com/NOAA-EMC/fv3atm/blob/9a9a701333305131fa7fd20ed32b8c3cf8570399/gfsphysics/GFS_layer/GFS_radiation_driver.F90#L2073-L2076).
RRTMG defines UV-B as shortwave band 27
([`radsw_main.f` L345-346](https://github.com/NCAR/ccpp-physics/blob/01ed01fb0b3112e96eb619e0339d88fb0201982f/physics/radsw_main.f#L345-L346),
selected at [L3037-3076](https://github.com/NCAR/ccpp-physics/blob/01ed01fb0b3112e96eb619e0339d88fb0201982f/physics/radsw_main.f#L3037-L3076)).
With `NBLOW=16`, band 27 is table element 12, whose bounds in
[`radsw_param.f` L117-180](https://github.com/NCAR/ccpp-physics/blob/01ed01fb0b3112e96eb619e0339d88fb0201982f/physics/radsw_param.f#L117-L180)
are 29,000-38,000 cm-1 — that is 10^7/38000 = 263.16 nm to 10^7/29000 = 344.83 nm,
written as 263-345 nm in the comments.

**Helicity: right mover in both hemispheres.** At `upp_v8.3.0`,
[`MISCLN.f` L168-184](https://github.com/NOAA-EMC/UPP/blob/c5f3053c9b22ac18287f14f9e255b45994cab6a3/sorc/ncep_post.fd/MISCLN.f#L168-L184)
sets `DEPTH(1)=3000.0` and emits `HELI(:,:,1)` for this variable, and
[`CALHEL.f` L329-345](https://github.com/NOAA-EMC/UPP/blob/c5f3053c9b22ac18287f14f9e255b45994cab6a3/sorc/ncep_post.fd/CALHEL.f#L329-L345)
applies the Bunkers 7.5 m s-1 right-moving deviation unconditionally. That file contains
no reference to latitude at all, so there is no hemisphere branch — which is why the
field is positive-mean in both hemispheres. The previously contested claim is therefore
correct for this archive.

**To watch at a future GFS upgrade, not a problem now.** UPP
[`12ab90c`](https://github.com/NOAA-EMC/UPP/commit/12ab90c72627d11644734af919f900edc2083932)
("Account for left-mover in SH in SRH", 2022-10-10) added a hemisphere switch that
selects the left-moving solution south of the equator; it is in current `develop`
([`CALHEL.f` L314-322](https://github.com/NOAA-EMC/UPP/blob/fe4f02b41fd789e414fb4bf23ad27c58ef058ee4/sorc/ncep_post.fd/CALHEL.f#L314-L322))
but not in the UPP version that produced any data in this archive. The convention is a
property of the source version rather than a law of nature. If a GFS upgrade brings that
UPP line into the archive, southern-hemisphere helicity flips sign and the boundary
belongs in the validation report's Review notes; the comment deliberately states only
what is true of the data, with no version numbers that would go stale.

## Checks

`ruff format --check`, `ruff check`, `ty check`, `pytest -m "not slow" -n 4` (2395
passed), and the four slow modules under `tests/noaa/gfs/{analysis,forecast}_virtual/`
all pass. Both templates regenerated in each commit.


src/reformatters/noaa/gfs/analysis_virtual/template_config.py
```diff
@@ -34,6 +34,10 @@
     "max": f"Maximum value over {_WINDOW_EXTENT}",
     "min": f"Minimum value over {_WINDOW_EXTENT}",
 }
+_PREVIOUS_CYCLE_COMMENT = (
+    "At 00, 06, 12 and 18 UTC this value comes from the previous forecast cycle, "
+    "because the source does not publish this variable at lead time 0."
+)
```
aldenks: Cut this on all vars. 

We will have a general docs note on `### Construction` about how hour 0 vs not vars are handled (eg for GFS analysis materialized the docs say "GFS starts a new model run every 6 hours and dynamical.org has created this analysis by concatenating the first 6 hours of each forecast along the time dimension."

aldenks (commented):
After this change, approved to automerge

PR #1019, opened 2026-09-06
Add the NOAA GEFS 35 day 0.5 degree virtual forecast dataset

aldenks:
Stacked on #1017 (`gefs-virtual-16-day`). Review that one first; this PR's diff against it is the new dataset plus the refactor that lets both share one catalog.

`noaa-gefs-forecast-35-day-0-5-degree-virtual` — every 0.5 degree pgrb2a and pgrb2b message of all 31 members, out to the 840 hour lead only the 00z cycle reaches. 24 hourly inits from `GEFS_VIRTUAL_ARCHIVE_START`, 181 lead times (3 hourly to 240h, 6 hourly to 840h).

It deliberately overlaps the 16 day dataset. Virtual refs are cheap and a reader wanting the long leads should not have to stitch two stores together.

## Shape

| | count |
|---|---|
| data variables | 268 |
| root | 247 |
| `pressure_level` (31 levels) | 12 |
| `model_level` (4 levels) | 6 |
| `height_above_mean_sea_level` (8 levels) | 3 |
| lead times | 181 |
| ensemble members | 31 |
| source files per init | 11,222 (181 × 31 × 2 products) |

## Manifest splitting

Splits are 8 inits at the root, 4 for the two small vertical groups and 2 for `pressure_level`. Measured against the real catalog at 2,166 inits (append dim start to today), with 5,611 refs per root array per init (181 × 31):

| array class | arrays | split | refs/manifest | manifest bytes @16.4 B/ref | refs/commit | manifests |
|---|---|---|---|---|---|---|
| root | 247 | 8 | 44,888 | 0.70 MiB | 11,087,336 | 66,937 |
| `pressure_level` | 12 | 2 | 347,882 | 5.44 MiB | 4,174,584 | 12,996 |
| `model_level` | 6 | 4 | 89,776 | 1.40 MiB | 538,656 | 3,252 |
| `height_above_mean_sea_level` | 3 | 4 | 179,552 | 2.81 MiB | 538,656 | 1,626 |
| **total** | **268** | | | | **16,339,232 (16.3M)** | **84,811** |

- **Manifest bytes** are inside the reader budgets: 0.70 MiB for a single-level variable (budget ≤ 3 MiB), 5.44 MiB for the largest vertical group (budget 5–8 MiB). `pressure_level` stays at 2 for exactly this reason — it is already the largest reader object, and doubling it would put it over.
- **Every manifest is far above icechunk's 1,000 ref `min_num_chunks` floor** — the smallest is now the root's 44,888, twice what it was.
- **The trade this coarsening makes, stated plainly.** Against the 4/2/2/2 alternative: **M falls 1.85× (156,617 → 84,811)** and refs/commit **rises 1.6× (10.3M → 16.3M)**. That is the hyperbola, not a free win — it chooses which end to pay. Root carries it: at split 8 the root is 79% of M.
- **Why that end.** icechunk 2.1.2 (which we deploy; 2.2.0 does not fix it) re-scans the whole manifest list once per manifest ref per commit, so commit cost carries a term quadratic in M. On the fitted coefficient — **derived, not measured on this dataset** — of 1.115e-9·M² seconds, the scan term falls from **~27 s to ~8 s per commit**. Refs per commit at 16.3M is above the 12.1M that operational HRRR sustains at a p50 of 2.8 s, so the ref term grows while the scan term shrinks; the net is favourable only because the scan term is the larger of the two here.
- **What would move these numbers.** The 16.4 B/ref is measured on our own manifests and is a packing fact — re-measure it if ref layout changes. The scan coefficient is the open item; if it moves, every scan figure above moves with it, while the refs/commit and MiB columns do not.

## Measured publication lags behind the crons

Five consecutive 00z cycles (2026-08-31 through 09-04) listed from `noaa-gefs-pds`, all 11,222 a+b files each:

| | measured |
|---|---|
| f000 lands | init+3h46m |
| lead times ≤ 384h complete | init+6h35m to 6h43m |
| 840h extension starts | init+6h35m |
| 840h extension complete | init+27h07m to **28h05m** |
| published by init+9h45m | 68.85% (all 5 cycles) |

A cycle publishes in two stages and the second outlasts the init interval, so the schedule is one fire per cycle that leads the first stage, with the trickle swept in a batch the next day:

- **update `45 3 * * *`, deadline 6h.** 03:45 leads f000 by a minute; the deadline covers the ≤384h stage's end with three hours to spare and reaches past init+28h05m of the cycle a day older, where that cycle's extension finishes.
- **validate `55 9 * * *`, deadline 30m** — the update's fire plus its deadline plus 10 minutes.
- **`operational_update_window` 72h** — three fires, so two consecutive missed runs still self-heal. Even with none missed the window has to reach a second init, because a cycle's own extension publishes past the next fire.
- Both crons ship **suspended**.

The tradeoff worth a second opinion: the >384h leads land in the store up to ~24h after the source publishes them. Following the trickle live instead would need a near-24h poll, and with no `tick_interval` override that turns ~250 commits/day into ~660, each rewriting 10.3M refs. The materialized 35 day dataset already ingests its extension a cycle late, so this is not a regression. Fire schedule is the only batching lever we have.

## Validators

- `CheckCurrentData(max_delay=9h50m)` — five minutes inside the validation offset, so a cycle is due as soon as the validation following it fires even if the pod starts early. Inits are a day apart, so tolerating one would leave a whole day unreported; the test pins that it alerts on the first stalled cycle. Derivation below.
- `CheckVirtualManifestCompleteness(min_present_fraction=(0.57, 1.0))` — the newest init holds only its 105 of 181 lead times when validation fires (6510 of 11222 files = 0.580). The slack absorbs the whole missing extension but no longer a whole ensemble member; the test pins both bounds. Derivation below.
- `CheckVirtualDecodeHealth()` — every variable appears in the ≤384h leads, so `positions=1` covers them all.

No completeness gate in `discover_available` — that is an analysis convention; a forecast publishes refs as files arrive and a partial newest position is intentional.

## Verification

- **22 real archived `.idx` fixtures** drive the real coord generation and ref building: both products, both cloud-mixing-ratio spellings (CLWMR/CLMR), lead 0, the 3 and 6 hour windows, the 240h step up, the 384h lead where every other cycle stops, the **390h lead where only 00z continues**, and the **840h end** — at the 2020-10-01 archive start and at a recent cycle, control and perturbed members. Every message the file offers is taken exactly once.
- **Integration test** backfills 2024-06-01 locally and runs a real operational update to 2024-06-02, with snapshot values for every variable kind at 8 lead times (including 390h and 840h), two members, and the ocean/ice-sheet/below-ground NaN cases. The overlapping leads (0, 3, 6, 240, 246, 384h) return **byte-identical values to the 16 day dataset's checked-in snapshots**, confirming the two stores agree where they overlap.
- The materialized `noaa-gefs-forecast-35-day` template regenerates **byte-identical** — `GEFS_EXTENSION_MAX` only names the 840h it already used. That dataset is published, so this was checked by recursive byte diff, not just `git status`.
- Full fast suite (2,795 passed) plus every module holding slow tests in `tests/noaa/gefs` and `tests/common/virtual*`, one per invocation.

## Metadata equivalence sweep

Keyed on `standard_name` across all 30 implemented datasets. The catalog is identical to the 16 day dataset's (verified field by field), so these are inherited, not new. Three clusters where this dataset carries a value no other dataset does — all three are distinct source messages that CF gives one name, so **no change made**:

| standard_name | ours | elsewhere |
|---|---|---|
| `air_pressure_at_mean_sea_level` | `pres` "Pressure" (PRES:mean sea level, b file) alongside `prmsl` and `mslet` | only `prmsl` and `mslet` |
| `temperature_difference_between_ambient_air_and_air_lifted_adiabatically` | `pli` "Parcel lifted index (to 500 hPa)" (PLI:30-0 mb above ground) alongside `4lftx` | only `4lftx` |
| `toa_brightness_temperature` | `btmp` "Brightness temperature" (BRTMP:top of atmosphere) | HRRR's `sbt113`/`sbt114`/`sbt123`/`sbt124` simulated channel temperatures |

All three confirmed against real index messages: PRES, PRMSL and MSLET are three separate messages at `mean sea level`; LFTX, PLI and 4LFTX are three separate lifted-index messages, and LFTX correctly takes the `..._from_the_surface` standard_name here exactly as in GFS and HRRR.

The repo's own `test_metadata_consistency_across_datasets` needed the same 15 `CROSS_DATASET_CONSISTENCY_EXCEPTIONS` entries the 16 day dataset carries (the `tcc` dual meaning, and GEFS publishing the categorical flags, cloud cover and heat/radiation fluxes as period averages where other models publish them instantaneously).

## Also in this PR

Two attribute corrections to the 16 day dataset, both unpublished and unbackfilled:

- `forecast_resolution` was `"Forecast step 3 hourly"`, wrong for its own 246–384h leads. Now derived from the lead axis: `"Forecast step 0-240 hours: 3 hourly, 246-384 hours: 6 hourly"`, matching the materialized 35 day dataset's wording.
- Its description said the a and b files *end* at 384 hours. This dataset proves otherwise; reworded to "the 384 hour lead time every cycle reaches".

---

Carrying #991's review over. This dataset's bounds were already derived rather than round, so both
moves here are small tightenings.

### `CheckCurrentData` — `max_delay` 9h55m → **9h50m**

| | |
|---|---|
| `pod_active_deadline` | **6h** |
| worst-case moment init *T* becomes present | init + 9h45m |
| validate cron fires | init + **9h55m** |

- **Lower bound, > 9h45m.** The last instant at which *T* is legitimately still absent is the end
  of its own update run, not the validate fire.
- **Upper bound, ≤ 9h55m.** For the validation run following a missed cycle to catch it, *T* must
  be due when that run fires. Inits are a day apart, so overshooting costs a **full day** of
  silence, not six hours.
- **Chosen 9h50m.** 9h55m sat exactly *on* the upper bound: `now - max_delay` landed precisely on
  *T*, so a pod starting one second early would floor to *T*−24h and a missed cycle would go
  unreported that run. 9h50m buys 5 minutes of early-start tolerance while staying 5 minutes above
  the lower bound.

`test_validators` drives the real check and still pins that the first stalled cycle alerts.

### `CheckVirtualManifestCompleteness` — `(0.55, 1.0)` → **`(0.57, 1.0)`**

This is the dataset the validator's own docstring names as *the* example for a leading tier
("GEFS 35-day's slow long lead times"), so the newest init is legitimately partial here. In files:

| | |
|---|---|
| expected source files per init | 181 leads × 31 members × 2 products = **11,222** |
| present at the validate fire — leads ≤ 384h | 105 leads × 62 = **6,510** (= 0.5801) |
| legitimately absent — the 840h extension | 76 leads × 62 = **4,712 files** |
| one ensemble member's share of the first stage | 105 leads × 2 products = **210 files** |

The extension is what is genuinely unpublished: leads ≤ 384h land between ~init+3h46m and
~init+6h43m, then the 840h extension arrives in bursts until ~init+28h05m and is swept by the next
day's fire. So the tier has to sit at or below 0.5801 — but only just.

| tier | requires present | tolerated missing | slack beyond the legitimate 4,712 |
|---|---|---|---|
| `0.55` (before) | ≥ 6,173 | 5,049 | 337 files — **more than one whole member (210)** |
| `0.57` (after) | ≥ 6,397 | 4,825 | **113 files — under one member** |

That is the whole point of the change: at `0.55` an ensemble member entirely absent from the
compact first stage **passed** while its init was newest. At `0.57` it fails. The legitimate
4,712-file extension absence is still tolerated with room. The test now pins
`allowed_missing < extension_files + per_member_first_stage` where it previously pinned the
opposite direction.

Counts confirmed by running the dataset's own `generate_source_file_coords` over one init:
11,222 coords, 6,510 at leads ≤ 384h.

Note the second tier stays `1.0` and holds: init *T*−24h's extension completes at *T*+4h05m and is
swept by the fire at *T*+3h45m, which runs to *T*+9h45m — so by the validate fire every older init
is whole.

## Cron job naming

The dataset id plus `-validate` is 53 characters against kubernetes' 52. The cron names now drop
`-degree` and **keep** the dashed resolution, rather than borrowing the provider's `p` spelling:

| | before | after | len |
|---|---|---|---|
| update | `noaa-gefs-forecast-35-day-0p5-virtual-update` | `noaa-gefs-forecast-35-day-0-5-virtual-update` | 44 |
| validate | `noaa-gefs-forecast-35-day-0p5-virtual-validate` | `noaa-gefs-forecast-35-day-0-5-virtual-validate` | **46** |

Checked in code — `CronJobName` is `max_length=52` and
`test_cron_job_names_fit_the_kubernetes_limit` asserts it.

URLs, and `FILE_RESOLUTIONS`.

## Metadata

**Dataset description** cut to the one sentence the published materialized siblings already use,
byte-for-byte. **Manifest split comment** reduced to a single orienting line. The accumulation
window comment follows the shared catalog trimmed in #993 (including the copy this PR hoisted into
`tests/noaa/gefs/virtual_template_config_test.py`).

`haines_index_surface` picks up the `flag_values` / `flag_meanings` fix made in #1017's shared 0.5°
catalog — confirmed present in this PR's regenerated template, not assumed.

`total_ozone_atmosphere` and the two gravity wave stresses pick up the GFS-matching metadata made in
#1017's shared 0.5° catalog — metres under
`standard_name="equivalent_thickness_at_stp_of_atmosphere_ozone_content"` with a `ScaleOffset`
filter, and no CF standard name on the stresses whose sign convention NCEP inverts. Confirmed
present in this PR's regenerated template, and confirmed that the stress comments keep **our**
forecast window wording with only the sign sentence taken from GFS.

## Tests

`file_refs` reads the GRIB header at the index's last offset to reject a stale index (#1021), so the
`fake_download` closures that stubbed only the index download reached the real S3 object. They now
go through `tests/noaa/grib_index_fixtures.stub_grib_source_file_reads`, the helper #1021
established, which stubs both reads.


huanglinfei091-cmd:
Hi @aldenks — I noticed the `Code Quality (arm64)` job failed at `Run Pytest` in [this GitHub Actions run](https://github.com/dynamical-org/reformatters/actions/runs/34073560564) for this PR.

I’m testing [RunBack](https://github.com/huanglinfei091-cmd/runback), an early alpha CLI that reconstructs a failed public Actions run locally with `act`. For this run, the useful question is whether the same pytest test and failure message are found, or whether reconstruction stops with an honest blocked result.

```bash
runback doctor
runback https://github.com/dynamical-org/reformatters/actions/runs/34073560564
```

The Linux amd64 binary and checksum are in the [v0.1.0-alpha release](https://github.com/huanglinfei091-cmd/runback/releases/tag/v0.1.0-alpha). If you try it, the verdict (`SAME_FAILURE`, `DIFFERENT_FAILURE`, or `REPLAY_BLOCKED`), Stage/Cause, and any install friction would be useful. Please do not post tokens or secrets. I’m asking for candid alpha evidence rather than a star or promotion; a failed reproduction is useful too.

src/reformatters/noaa/gefs/forecast_35_day_0_5_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def operational_kubernetes_resources(self, image_tag: str) -> Sequence[CronJob]:
+        # The dataset id plus "-validate" exceeds the 52 character cron job name limit.
+        cron_job_name_prefix = self.dataset_id.replace("-0-5-degree", "-0-5")
+        # A run publishes ~init+3h46m through ~init+28h05m.
+        # Fire just before the first files become available and stop 30m after expected completion.
+        operational_update_cron_job = ReformatCronJob(
+            name=f"{cron_job_name_prefix}-update",
+            schedule="45 3 * * *",
+            pod_active_deadline=timedelta(hours=6),
```
aldenks, line 60: Clarify the two bursts GEFS extended publish schedule: f000 - f3xx publishes ~init+XX through ~init+YY, f368-f840... publishes ...; 6h deadline covers first burst, 2nd burst is ingested asap on next days fire

src/reformatters/noaa/gefs/forecast_35_day_0_5_degree_virtual/dynamical_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+            image=image_tag,
+            dataset_id=self.dataset_id,
+            cpu="4",
+            # A fire opens by ingesting the whole extension of the previous cycle,
+            # the largest single batch of refs it holds.
+            memory="16G",
```
aldenks: cpu 3.5 and mem 15G (to schedule on a 4 core 16G node)

PR #1024, opened 2026-09-07
Document GFS PV surface hemispheric gaps

aldenks:
## Summary

- Extend the existing NaN comment on all 48 GFS virtual PV-surface variables to describe the hemispheric structure: positive PV surfaces are largely absent in the southern hemisphere, and negative surfaces in the northern.
- Regenerate the shared catalog metadata into both the analysis and forecast virtual templates.
- This is metadata-only, rides an operational metadata refresh, and requires no re-backfill or re-ingest.

Two large backfills are currently running for these datasets. Opening this PR is safe, but this metadata change must not be deployed under those active backfills.

Neither GFS virtual dataset is currently in the production STAC catalog, so this additive attribute correction has no production backward-compatibility obligation.

## Verification

- Confirmed 48 variables: six quantities on each of the eight surfaces (`0p5`, `1`, `1p5`, `2` PVU and their negative counterparts).
- `uv run ruff format`
- `uv run ruff check --fix`
- `uv run ty check --output-format concise`
- `uv run pytest -m "not slow" -n 4` (2,624 passed)
- `uv run pytest tests/common/common_template_config_subclasses_test.py` (234 passed)
- `uv run pytest tests/common/datasets_cf_compliance_test.py` (225 passed)
- All nine GFS test modules containing slow tests, one module per invocation (237 passed total).

### DIFF
--- a/src/reformatters/noaa/gfs/analysis_virtual/templates/latest.zarr/geopotential_height_0p5pvu/zarr.json
+++ b/src/reformatters/noaa/gfs/analysis_virtual/templates/latest.zarr/geopotential_height_0p5pvu/zarr.json
@@ -37,7 +37,7 @@
     "short_name": "gh",
     "standard_name": "geopotential_height",
     "units": "m",
-    "comment": "NaN where this potential vorticity surface does not exist in the column.",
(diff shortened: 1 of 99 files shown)


aldenks:
We don't need to explain normal details of source data, closing

PR #1026, opened 2026-09-07
Add a cleanup script for stale backfill job artifacts

aldenks:
Adds `src/scripts/cleanup_job_artifacts.py`: lists the `_job_*` staging branches and `_internal/<job>/` coordination prefixes a killed or crashed backfill leaves on a dataset's icechunk store, checks whether the kubernetes job is still active (fails closed on any API error other than an explicit 404), prints a plan, and with `--apply` deletes only stale artifacts, refusing `main`, any non-`_job_` branch, and anything belonging to an active job. `--job` matches full job names only, since successive runs of one backfill share prefix and timestamp. Dry run by default; the ops card notes that deleting a branch only unpins manifests and `icechunk_utils.py garbage-collect` frees the bytes.

Motivation: the analysis and forecast GFS virtual stores each carry a staging branch from a killed run (`..._muv9`, `..._153c`) that pins storage until removed.

Implemented on codex (gpt-5.6-sol), reviewed by Claude with refusal-path tests added and mutation-checked.


aldenks:
Not merging, +400 LoC for a little utility. Can be resurrected if needed

PR #1029, opened 2026-09-07
Distinguish rejected virtual source files from ref-building failures

aldenks:
**Design choice / missing-data trade:** discarding a bad source file lets the run publish, but it leaves fill values at that coord. The coord leaves `pending` and is never retried **within that run**. Because no representative ref is written, a later run automatically re-offers it; the gap self-heals only if a later run still covers the coord after the publisher replaces the object with a valid one. Otherwise it is a permanent gap. Until then, the evidence is one per-file WARNING, one per-worker ERROR aggregate grouped by reason, each ingest batch's skipped count, manifest-completeness validation for recent operational coords, and availability plots. Historical gaps have no automatic detector on `main` today.

This PR takes that quiet branch only for source states proven from one published object's bytes or metadata. Missing representative coverage and GEFS all-variable completeness remain fatal because either can be our catalog/configuration bug, and no design should make our bug quiet as a side effect.

## Why

A 391-pod GFS backfill could not finalize after one worker hit a truncated `gfs.20241020/18` pgrb2b f249 object: 71 MB instead of about 238 MB, with a matching short index. `_assert_probe_chunk_covered` raised; because the old wrapper deliberately re-raised `AssertionError` while discarding every other exception, that one source object killed the worker and prevented its index from producing a results file. The finalizer then polled forever for every worker result, so the run published nothing.

The old boundary also had the opposite failure mode: arbitrary `ValueError`, `KeyError`, and transient download/storage errors inside `file_refs` became silent gaps. The lasting fix is therefore not “catch this assert”; it is one explicit source-rejection type and a complete classification of the reachable assertions and escaping raises.

This PR delivers a narrower invariant: **A BACKFILL SOURCE-REJECTION BURST MUST NOT STRAND THE FINALIZER.** Only `SystemicSourceRejectionsError` gets publish-then-report. `_assert_probe_chunk_covered` — the incident's original trigger — the GEFS completeness assertion, OOM, and any untyped raise still kill a worker; after retries that index writes no results file, while `wait_for_workers` polls every 10 seconds with no timeout. [PR #1035](https://github.com/dynamical-org/reformatters/pull/1035) bounds those coordination waits to the Job's active deadline. For the exception class handled here, a backfill commits every good ref, records the worker error, writes the normal results file in a multi-worker run, lets the last worker finalize and publish, and only then makes the Kubernetes Job non-zero. Publishing 723 good files and screaming is strictly better than publishing none and hanging because 21 were bad.

## Recommended boundary

`SourceFileRejectedError`, raised by `file_refs`, is the typed per-file discard protocol. `_file_refs_or_skip` catches only that type. It does not catch `AssertionError`, broad `Exception`, or a returned exception object. The thread result wrapper only transports a raised rejection back to the loop; it is not a second subclass API.

The current code has two rejection protocols. `SourceFileRejectedError` is the general typed boundary; HRRR's `_TRUNCATED_SOURCE_FILES` is a separate discovery-time denylist for two known damaged objects, so they bypass the typed boundary, counters, and aggregate. PR #1031 retains that denylist because one object lacks the representative probe and cannot safely reach its detector. This PR does not churn the list while that composition is reviewed.

| Location / condition | Classification | Consequence before | Consequence in this PR | Datasets |
|---|---|---|---|---|
| Common broad `file_refs` wrapper | Mixed and unsafe | Discarded almost every exception, including our bugs and transient failures; assertions escaped | Catch only typed source rejection; all other failures stay fatal/retriable | All virtual datasets |
| Bare empty `file_refs` result | Our implementation/configuration invariant | Supported silent skip | Assertion names the source URL and is fatal | All virtual datasets |
| Representative probe not covered | Ambiguous: incomplete source or wrong representative selection | Assertion, fatal | Remains assertion, fatal | All virtual datasets; operationally important for GFS/GEFS/HRRR |
| GEFS requested variable has no ref | Ambiguous: incomplete source or wrong `available_from`/catalog era | Assertion, fatal | Remains assertion, fatal; never discards the file's other valid variables | Both GEFS virtual datasets |
| NOAA empty/malformed index | Untrusted source state | Empty could silently skip; malformed behavior depended on exception type | Typed rejection | GFS, GEFS, HRRR virtuals |
| NOAA non-increasing, negative, or out-of-object index range | Untrusted source state | Some discarded; negative offsets could escape or produce invalid refs | Typed rejection before any ranged read | GFS, GEFS, HRRR virtuals |
| NOAA last offset lacks a complete GRIB2 section 0, declares length <16, or does not fit | Untrusted source state | Recognized stale indexes skipped; impossible short lengths could pass | Typed rejection | GFS, GEFS, HRRR virtuals |
| AIFS empty/malformed/schema-invalid index | Untrusted source state | Discarded by the old broad wrapper; became fatal in the first PR version | Typed rejection | AIFS virtual |
| AIFS boolean/non-integer offsets or lengths on any row; invalid levels that could select a requested field; non-positive, negative, or out-of-object ranges | Untrusted source state | Pandas could coerce booleans into integers and build valid-looking wrong refs; eager level parsing could reject unrelated rows | Decode raw JSON scalars before pandas coercion; validate every byte range eagerly; convert levels only for candidate lookup rows | AIFS virtual |
| WeatherNext2 present-but-invalid size or MD5 for a named native object | Untrusted source metadata | Some values asserted, escaped untyped, or were truncated into valid-looking refs | Strict size/MD5 validation reaches `file_refs` and raises typed rejection | Both WeatherNext2 virtual datasets |
| WeatherNext2 required listing field absent | Our API model is wrong | Severity varied by field | Fatal consistently | Both WeatherNext2 virtual datasets |
| Template/chunk geometry, representative/catalog selection, coord generation, registered containers, append extent, worker batching, and source-layout assertions | Our invariants | Mixed: assertions were fatal, but some non-assert failures inside `file_refs` were swallowed | Fatal | Their respective virtual datasets |
| Discovery, download, ranged-read, Icechunk commit, and object-store exceptions | Infrastructure/transient state | Discovery/commit were fatal; errors inside `file_refs` were discarded | Fatal so orchestration can retry | Their respective virtual datasets |

An available NOAA file whose index contains no requested variable remains fatal. For a filtered new-variable backfill on GFS or HRRR forecast, which do not encode every variable's source-era boundary, the operator must use `--filter-start` at the first source date that carries the variable. The backfill guide now says so; silently treating an unknown era mismatch as an absent file would hide catalog mistakes.

## Backfill reporting threshold

The backfill-only rule is:

```text
one rejection is always allowed;
from the second onward, report systemic failure if
rejected / initially_planned >= 10% or rejected > 20
```

The denominator is `len(remaining)` at the start of the worker loop, after the manifest-presence filter, not attempts so far. It includes files that have not published yet. The counts are worker-wide across the union of assigned jobs and across WeatherNext2's manifest-sized backfill groups. The aggregate counts reason strings rather than an exception class that is identical for every rejection.

Crossing the threshold no longer interrupts the write loop. The current batch and every later manifest group yield their good refs. `process_worker_jobs` raises the reported condition only after those commits; in a multi-worker run, `_process_region_jobs` defers it and writes `errors/worker-N.txt` before `results/worker-N.json`. That order is pinned because the results count gates the finalizer's error read. The last worker collects every result, finalizes the temporary branch onto `main`, logs the aggregate failure, and exits with dedicated code 42. A reported failure retains its temp branch and coordination files, so a cleanup error or interruption after publication but before exit 42 can retry against intact state. The Job's first `podFailurePolicy` rule maps code 42 to `FailIndex`, which marks the completion index failed immediately without consuming `backoffLimitPerIndex` or creating a replacement pod. This rule must remain before `Ignore` on `DisruptionTarget`: `Ignore` creates a replacement pod, so reversed rules could replace an exit-42 pod disrupted on a spot node. Other failures retain the normal per-index retry policy.

On a GFS forecast backfill the measured 2,925,247 coords across 391 workers are about 7,481 planned per worker, so the absolute ceiling reports on rejection 21. On a 744-file worker, 21 rejected files still allow the other 723 files' refs to publish. Before this change, the same 21 typed rejections would have published 723 and logged errors; the previous breaker version instead discarded all 723 and deterministically stranded finalization.

Every current dataset inherits the same `1 / 10% / 20` backfill policy. The class variables remain dataset-overridable, but the measured population evidence calls for a processing-mode distinction rather than per-dataset tuning: backfills have enough observations for a rate/ceiling breaker; updates do not.

## Operational updates are exempt

Operational update mode never evaluates the rate/ceiling breaker. Its population is the post-manifest-filter active window, which is deliberately small. Two unchanged bad objects can therefore exceed 10% and block every fire until they age out, even though the worker-wide initial-plan denominator is calculated correctly.

Measured from each dataset's checked-in operational job shape:

| dataset | window | cron | cold-window planned | reports at | steady-state planned | reports at |
|---|---:|---:|---:|---:|---:|---:|
| **GEFS analysis 0.25°** | 18h | 3h | **9** | **2** | ~2 | **2** |
| **GFS analysis** | 12h | 6h | 28 | 3 | ~14 | **2** |
| **HRRR analysis** | 12h | 1h | 48 | 5 | ~4 | **2** |
| HRRR forecast 18h | 6h | 1h | 342 | 21 | ~57 | 6 |
| HRRR forecast 48h | 14h | 6h | 294 | 21 | ~126 | 13 |
| GFS forecast | 18h | 6h | 1254 | 21 | ~418 | 21 |
| GEFS forecast | 18h | 6h | 7533 | 21 | ~2511 | 21 |
| AIFS forecast | 20h | 6h | 183 | 19 | ~55 | 6 |

GEFS analysis 0.25° would report on two rejections even in a cold nine-coordinate window; GFS and HRRR analysis would report on two in steady state. Exemption avoids turning two permanent source defects into a 12–18 hour publication wedge. Nothing becomes silent: per-file WARNINGs, the reason-counted worker ERROR aggregate, skipped counts, and completeness validation all remain active in update mode.

## Evidence for the backfill constants

- GFS forecast: 335 rejected files out of 2,925,247 across 327 pods, roughly one per pod overall.
- GFS analysis: 8 rejected files, all in one pod of about 744 files, or 1.1%.

The measured clustered worst case, 1.1%, remains tolerated. Ten percent is about a 9x margin above it; the absolute ceiling prevents that fractional margin from permitting hundreds of discards on a large worker. The one-file floor handles tiny backfill workers, the rate catches an all/nearly-all small backfill worker, and the ceiling catches a substantial burst in a large worker. Ambiguous catalog symptoms stay fatal from the first file.

## Interaction with #1027 and #1031

#1027 moves the GFS pgrb2b probe from about 90.8% into a normal index to about 12%. With that shallower probe, the incident's 30%-sized file passes the probe, passes the last-offset header check, and is accepted with 77 of 308 refs. There is then no rejection, ERROR, or skipped count, and the present probe ref prevents automatic re-offer on later runs.

External evidence from the recovery-run watcher establishes that NOAA published a matching short index: the final index offset plus that GRIB message's declared length equals the object's byte count exactly. The index and object are mutually consistent; this is an internally valid file that is only 30% of the normal product. No intra-file offset, range, or header check can distinguish it from an intentionally smaller product.

So #1027 stops that object from killing a worker, while #1029 makes proven per-object damage use an explicit typed boundary and prevents its reporting threshold from stranding finalization. Neither detects a matching short data/index pair. #1031 adds the missing cross-file size/index-line baseline while retaining HRRR's hand-maintained `_TRUNCATED_SOURCE_FILES` path for the object that cannot safely reach the detector. To make the two PRs compose, this PR supplies the generic observation hook and passes it every accepted `_FileRefsResult`, including zero-ref files, before enforcing the empty-ref assertion; a regression test fails if that order is reversed.

## Alternatives considered

1. **Implemented:** typed discard only for unambiguous per-object source state; keep ambiguous/internal failures fatal; report backfill bursts only after publishing all good refs; exempt operational updates.
2. Keep every source anomaly fatal. This maximizes completeness but repeats the demonstrated failure mode in which one publisher defect strands a distributed run.
3. Discard every typed or ambiguous symptom without a backfill threshold. This maximizes publication progress but can quietly erase regions for our representative/catalog bug.
4. Tune separate update thresholds per dataset. The eight measured populations show that this adds knobs without fixing the underlying mismatch: update windows are too small for a population-rate breaker, while their existing completeness checks already provide the right signal.

This remains one reviewable PR because every code change enforces the same failure boundary. The AIFS and WeatherNext2 exact-integer fixes stayed here because each was a small strict-parse-then-typed-reject correction; neither required a separate parser design.

## Verification

The final verification invocations ran after the machine-wide pytest/render/memory gates, with at most one pytest process and `-n 4`; each reviewed module ran in its own invocation. No render was run.

- Common virtual region job: 85 passed.
- Common parallel coordination / dynamical dataset: 23 / 30 passed.
- NOAA shared virtual region job: 30 passed.
- AIFS virtual region job: 30 passed.
- WeatherNext2 native virtual region job: 19 passed.
- GEFS analysis / forecast virtual region jobs: 48 / 32 passed.
- HRRR analysis / 18-hour / 48-hour virtual region jobs: 16 / 2 / 6 passed.
- Kubernetes job manifest: 32 passed.
- Repository-wide `ruff format`, `ruff check --fix`, and `ty check --output-format concise`: passed.


aldenks:
Closing: not worth its code.

This introduced a typed rejection boundary so an unambiguously bad source object is discarded and counted rather than killing a worker, with a rate breaker to keep a systemic outage loud. Four review rounds established that it delivers publish-then-report for exactly one exception class, while every other worker death — a different assert, an OOM, an eviction — still behaves as before. 1,309 added lines for that, against a source-damage base rate of one anomalous file in 2,925,247 ingested.

The part worth keeping is the smallest part of the diff and is now #1037: AIFS parsed its index through pandas, whose dtype inference runs across rows, so a boolean offset or length was silently coerced to an integer and produced a reference that looked fine and pointed at the wrong bytes. That was reproduced on the production path, and it is the only defect here that produced wrong data rather than a loud failure.

Note for anyone tempted to lift more from this branch: its level parsing was made strict too, and a later review measured the result — a float level rejected the whole file while quoted variants were dropped with no warning, no count and no detector, where main matched all three. #1037 keeps the byte-range strictness and preserves main's level behaviour.

The remaining findings live in the coordination-dir reports. Branch retained.

PR #1031, opened 2026-09-07
Report consistently truncated NOAA virtual sources

aldenks:
## Summary

- detect NOAA GRIB data/index pairs that are consistently truncated relative to like-for-like files at other init times
- report both a file-level ERROR and a per-run ERROR count without rejecting the source; usable refs continue through the normal write path
- include files whose parsed indexes produce zero refs, so the strongest truncations are not silently omitted

## Why this must be cross-file

The motivating `gfs.20241020/18/atmos/gfs.t18z.pgrb2b.0p25.f249` object and index are mutually plausible but incomplete. The last indexed message fits within the 71,229,440-byte object; an additional unindexed GRIB message begins after it and is cut off before its declared end. Healthy early HRRR files also demonstrate that an index may legitimately omit complete trailing messages, so neither indexed-message coverage nor exact equality with EOF can distinguish this incident generically.

Independent downstream evidence agrees with the truncation diagnosis:

- the object is about 30% of a healthy peer
- the index has 112 rows versus 349 for the peer
- the product contributes 77 owned messages where it normally contributes 308
- two recovery workers over identical 4,180-file sets differed by exactly 231 refs (`308 - 77`)
- the published read side has pgrb2b levels 425–875 hPa absent while 375 hPa and above are present, matching the short index's final `HGT:375 mb` line

That is why the detector compares sibling files instead of adding another intra-file integrity rule.

## Rule and behavior

Files form a cohort when every source-coordinate field except `init_time`, `data_vars`, and mutable status/download fields matches. In practice this holds model/dataset, product, domain, member, and forecast lead fixed and compares only other initialization times. A file is reported only when **both**:

- object bytes are below 50% of the median of its peers
- parsed index line count is below 50% of the median of its peers

The known GFS positive is 71,229,440 / 227,056,502 bytes = **31.37%** and 112 / 349 index rows = **32.09%** against its like-for-like peer. Files whose parsed index yields no usable refs still enter the comparison; the real `hrrr.t12z.wrfsfcf01` control is 17,301,504 bytes / 23 rows against roughly 85 MB / 102 rows and yields zero refs.

Detection is observational and never discards or rejects a file. Any usable refs remain eligible for the commit, while neutral “detected” wording avoids claiming that a later probe, write, or commit succeeded. The detailed and aggregate ERROR logs surface through Sentry.

## Backfill-only scope

This is a backfill/re-backfill detector, not an operational detector. An operational fire calls `filter_already_present` first and normally retains exactly one new `init_time`. Every like-for-like cohort consequently has one member, so its peer list is empty and the detector makes no claim. An ops-capable follow-up needs a durable per-cohort baseline from outside the filtered run; that is intentionally out of scope here.

The rule also cannot report a truncation that retains at least half of either bytes or rows, or a cohort whose peers share the same truncation.

## HRRR denylist remains load-bearing

Both denylisted 2016 `wrfnat` objects flag against real peers, but they cannot be admitted safely:

- 10Z is 223,477 bytes / 3 rows. Its only surviving messages are `CLMR`, `CICE`, and `SNMR` at one hybrid level; the representative `model_level/temperature` chunk is absent, so removing the denylist makes the ingestion probe assertion kill the worker.
- 12Z is 9,162,752 bytes / 14 rows. The detector identifies it, but it remains denylisted because admitting it would permanently mark a roughly 2%-complete file as ingested.

The denylist therefore stays in both this PR and #1029. Removing it later is a data-quality change, not cleanup: 12Z would change from excluded and uniformly NaN to permanently ingested at roughly 2% completeness once its probe chunk is present.

## GEFS limitation

GEFS truncations cannot currently reach the observation hook. `NoaaGefsVirtualRegionJob._check_refs_complete` asserts inside `file_refs` first, so an incomplete GEFS file kills the worker before detector metrics are evaluated. Changing that hook belongs with the typed rejection work, not this PR. The detector is therefore exercised end-to-end for GFS/HRRR behavior only; GEFS appears below solely in compatibility regression tests.

## Threshold evidence

An independent archive sweep covered 2,223 healthy like-for-like comparisons across GFS, HRRR, and GEFS and deliberately straddled HRRR v2/v3/v4, GEFS MSLET and b22, GFS v16.2/v16.3, and earlier row-count eras. It produced **zero flags** at the 50%-AND rule.

The closest healthy adjacent-init ratios were **80.66% for bytes** and **87.06% for rows**, leaving about **30 percentage points of byte clearance** and **37 points of row clearance**. Across eras the floors were 55.96% and 60%. Row-count invariance is a GFS-only property: healthy HRRR transitions reach 87%, so tightening a row arm above roughly 85% would create version-boundary false positives.

The peer axis is what excludes the earlier failed design. A same-cycle across-lead comparison falsely reports healthy 2016 HRRR `wrfsfc` f04/f05 at about 23% in both measures; holding lead fixed across init times avoids that failure. The regression test now uses distinct init times plus a healthy same-lead peer, so it fails if lead is removed from the cohort key.

## Composition with #1029 — merge requirement

#1029 and #1031 conceptually compose, but their current tips conflict in the shared write loop, NOAA implementation, and tests. Merge #1029 first, then resolve #1031 on top without retaining #1031's older broad exception catch. At #1029 SHA `74e6aed7`, the non-rejected zero-ref observation requirement is satisfied: `_file_refs_or_skip` returns a non-rejected `_FileRefsResult(refs=[])`; `_accepted_ref_batch` passes that result to `source_file_anomalies` first, logs and accumulates the returned anomalies, and only then applies the fatal empty-ref assert. #1031 requires no further implementation from #1029.

Conflict resolution must preserve that order along with `_FileRefsResult`, its rejection accounting/breaker, and the typed-only catch. Every non-rejected result—including empty `result.refs`—must reach `source_file_anomalies` before zero-ref write-batch filtering and the assert. Rejected results must stay outside the observation hook so `SourceFileRejectedError` remains the sole quiet rejection boundary and its accounting stays unchanged. The detector remains observational: it introduces no rejection status, sentinel, or second catch path. Retain the HRRR discovery-time denylist described above.

## Verification

- `uv run --no-sync ruff format`
- `uv run --no-sync ruff check --fix`
- `uv run --no-sync ty check --output-format concise`
- `tests/common/virtual_region_job_test.py` — 78 passed
- `tests/noaa/noaa_virtual_region_job_test.py` — 30 passed
- `tests/noaa/gfs/virtual_region_job_test.py` — 93 passed
- the zero-ref wiring regression fails under an in-process mutation that passes only ref-bearing files to the observation hook
- the same slow virtual write-loop modules cited in review, one path per pytest invocation with `-n 4` — 64 passed, 1 skipped across common, HRRR, GFS, GEFS, and AIFS; the GEFS results establish shared-loop compatibility, not detector reachability


aldenks:
Closing: not worth its code.

This detects a source file truncated together with its byte index — a shape every intra-file check is structurally blind to, because the index and the object are mutually consistent. It works: it flags the known case at 31% of its peers' size and produced zero false positives across 864 like-for-like comparisons spanning GFS, HRRR and GEFS.

But it is 520 added lines to detect something with a measured base rate of one occurrence in 2,925,247 files ingested, and it is backfill-only: an operational fire ingests a single init_time after `filter_already_present`, so every cohort has one member and the peer list is always empty. It would not have reported the next occurrence.

The durable findings are kept in the coordination-dir reports rather than the code: the peer-comparison rule and its 864-comparison sweep, the measured clearance (~30 percentage points on object size, ~37 on index rows, against healthy minima of 90.75% and 100%), the fact that healthy row counts are invariant only for GFS and vary to 87% adjacent and 60% across eras for HRRR, and the two HRRR objects that would flag if the discovery denylist did not exclude them first.

Branch retained.

PR #1035, opened 2026-09-07
Bound parallel worker waits by the pod deadline

aldenks:
## Summary

- derive one absolute coordination deadline from each Kubernetes Job pod active deadline, reserving its already-configured termination grace period
- time out result coordination with every missing worker index in the error, without finalizing or publishing partial data
- apply the same deadline to worker-0 setup polling in a separate commit
- require the injected deadline values for every multi-worker run rather than silently falling back to an unbounded wait

## Gap closed

#1029 makes one exception class publish-then-report, but _assert_probe_chunk_covered — the original incident trigger — the GEFS completeness assert, OOM, and any untyped raise still kill a worker; after retries that index writes no results file while wait_for_workers polls every 10 seconds forever. #1027 and #1029 fixed triggers. This PR fixes the mechanism.

## Deadline composition and safety

The Job model already requires pod_active_deadline (with a six-hour default), and every Job, CronJob, and ReformatCronJob that can reach multi-worker coordination inherits it. Backfill Jobs copy that value from the reformat resource. Local backfills are single-worker and do not coordinate. The manifest now also names the existing 30-second terminationGracePeriodSeconds setting, validates that the active deadline exceeds it, and injects both values into every worker.

At process startup the worker derives one monotonic deadline as:

    process start + pod active deadline - pod termination grace period

Processing, worker-0 setup polling, and final result polling all share that one deadline. The timeout cannot cause a failure that would not have happened anyway, since it fires strictly inside the window Kubernetes would have killed the pod in regardless. It does not shorten a healthy run; it converts a silent SIGKILL into a diagnosable Python error naming the missing indexes while preserving the configured grace window for logging and monitoring.

On result expiry, the coordinator lists results/worker-N.json and raises TimeoutError naming every missing index. It does not call finalize, reset Icechunk main, publish Zarr metadata, or clear the temporary branch or coordination files. collect_results delegates to the same wait, so materialized updates get identical behavior. Setup expiry similarly names worker index 0 and occurs before any waiting worker processes data.

## Verification

- uv run pytest -m "not slow" -n 4: 2632 passed
- tests/common/dynamical_dataset_test.py: 30 passed
- tests/common/failure_injection_test.py: 2 passed
- tests/common/parallel_writes_test.py: 22 passed
- tests/common/virtual_multi_group_test.py: 17 passed
- tests/common/virtual_region_job_test.py: 77 passed
- uv run ruff format
- uv run ruff check --fix
- uv run ty check --output-format concise

Mutation proof: in an isolated /tmp source copy, deleting the result-wait expiry branch made TestWaitForWorkers::test_never_reporting_worker_times_out_with_missing_indexes fail immediately with "wait continued past deadline". The shared worktree was never mutated for this proof.

### DIFF
--- a/docs/parallel_processing.md
+++ b/docs/parallel_processing.md
@@ -73,7 +73,7 @@ For `workers_total > 1`, workers coordinate via files in an object store directo
 
 ### Setup signal
 
-Worker 0 writes `setup/ready.json` after completing setup (creating branches, writing metadata). Workers 1+ poll for this file before proceeding.
+Worker 0 writes `setup/ready.json` after completing setup (creating branches, writing metadata). Workers 1+ poll for this file before proceeding. This poll shares the same deadline as processing and result coordination; if worker 0 never reports readiness, each waiting worker raises an error naming index 0 while the live store remains untouched.
 
 ### Results
 
@@ -105,6 +105,8 @@ On restart, worker 0 retries setup:
 - `setup/ready.json` is written (or overwritten) when setup completes
 
 Workers 1+ that were polling for setup will proceed once the file appears.
+
+If worker 0 never reaches readiness before exhausting its retries, those workers stop polling at the shared coordination deadline and report that index 0 did not complete setup.
 
 ### Last worker dies during finalization
 
--- a/docs/parallel_processing.md
+++ b/docs/parallel_processing.md
@@ -79,6 +79,8 @@ Worker 0 writes `setup/ready.json` after completing setup (creating branches, wr
 
 Each worker writes `results/worker-{N}.json` containing its `process_results` dict. The last worker (by index) polls until all result files are present, then aggregates them. For updates, the aggregated results drive `update_template_with_results` to trim the template based on what was actually processed.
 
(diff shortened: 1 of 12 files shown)


aldenks:
Closing: a second mechanism enforcing a bound Kubernetes already enforces.

This bounds `wait_for_workers` and the setup poll in code, deriving the deadline from the pod's own `activeDeadlineSeconds` minus its termination grace, and on expiry raises naming the indexes that never reported.

The bound it adds is the one Kubernetes already applies authoritatively: `activeDeadlineSeconds` terminates the pod regardless. 332 added lines buy a tidier final message and nothing else — and now that #1036 logs the missing worker indexes on every poll, that message arrives every ten seconds from the first minute instead of once at the end. An in-code deadline enforcing the same bound is a second way to do one thing.

Two facts found while building it are worth more than the change and are kept in the coordination-dir reports. The CronJob controller's `batch.kubernetes.io/cronjob-scheduled-timestamp` annotation is set on Jobs but **not** on Pods, so `fieldRef` cannot read it — verified as 109 of 117 Jobs and 0 of 18 Pods on the live cluster. And a Job's name suffix is the scheduled fire in Unix minutes, matching that annotation exactly on all 107 cron-owned Jobs checked, which makes it a usable shared reference if one is ever needed. Separately: `_virtual_poll_deadline` derives its anchor from the current clock, so a pod replacing an evicted one near a fire boundary can compute a later deadline than its predecessor — the thing its docstring says it prevents. Harmless while those updates are single-worker, but real.

Branch retained; the unpushed work is preserved in a named stash recorded in the handoff report.

PR #1036, opened 2026-09-07
Log missing parallel worker indexes

aldenks:
## Summary

- list the exact worker indexes that have not reported results yet
- cap each log entry at the first 10 missing indexes
- keep the result poll as an object-store listing without reading result contents

## Why this is separate

This diagnostic was pulled out of #1035 so it can merge without waiting on the job-wide timeout redesign. During the recent incident, the finalizer logged only `Waiting for all workers to complete...` every 10 seconds for eight hours even though the identity of the one missing index was available from the same object-store listing. This change makes that stall actionable on the first poll.

## Follow-up: centralize the worker result filename format

The `worker-{i}.json` format is duplicated: `dynamical_dataset.py` builds it while `parallel_coordination.py` parses it with a separately maintained regex. Both new tests also hand-write the literal in their fakes instead of exercising the producer, so a rename can break the wait loop without failing a test. A shared `worker_result_file_key(i)` used by the writer and regex closes the gap in about four lines, and `dynamical_dataset.py` already imports `parallel_coordination`, so it adds no dependency. This is deliberately deferred because #1036 should merge today.

## Merge order

#1036 lands first. The `gfs-worker-wait-timeout` branch for #1035 already contains this exact hunk plus a deadline parameter, so #1035 must rebase onto #1036 before it lands; whichever branch lands second otherwise conflicts in `wait_for_workers`.

## Verification

- `uv run --no-sync pytest -n 4 tests/common/parallel_coordination_test.py tests/common/storage_test.py` (95 passed)
- `uv run --no-sync ruff format` on changed files
- `uv run --no-sync ruff check` on changed files
- `uv run --no-sync ty check --output-format concise` on changed files

### DIFF
--- a/src/reformatters/common/parallel_coordination.py
+++ b/src/reformatters/common/parallel_coordination.py
@@ -4,6 +4,7 @@ See docs/parallel_processing.md for the overall design.
 """
 
 import json
+import re
 import time
 from collections.abc import Collection, Mapping, Sequence
 from pathlib import Path
@@ -21,6 +22,8 @@ from reformatters.common.storage import StoreFactory
 from reformatters.common.zarr import copy_zarr_metadata
 
 log = get_logger(__name__)
+
+_RESULT_FILE_PATTERN = re.compile(r"worker-(?P<worker_index>\d+)\.json")
 
 _WORKER_RESULTS_ADAPTER: TypeAdapter[dict[str, list[SourceFileResult]]] = TypeAdapter(
     dict[str, list[SourceFileResult]]
@@ -130,11 +133,22 @@ def wait_for_workers(
     # Rely on kubernetes pod_active_deadline for timeout.
     if workers_total <= 1:
         return
-    while (
-        store_factory.count_coordination_files(reformat_job_name, "results")
-        < workers_total
-    ):
-        log.info("Waiting for all workers to complete...")
+    while True:
+        result_files = store_factory.list_coordination_files(
+            reformat_job_name, "results"
+        )
+        reported_workers = {
+            int(match["worker_index"])
+            for result_file in result_files
+            if (match := _RESULT_FILE_PATTERN.fullmatch(result_file)) is not None
+        }
+        missing_workers = sorted(set(range(workers_total)) - reported_workers)
+        if not missing_workers:
+            return
+        log.info(
+            f"Waiting for {len(missing_workers)} of {workers_total} workers to "
+            f"complete; missing worker indexes: {missing_workers[:10]}"
+        )
         time.sleep(10)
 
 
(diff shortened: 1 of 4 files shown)


src/reformatters/common/parallel_coordination.py
```diff
@@ -125,16 +128,26 @@ def parallel_setup(
...
+            int(match["worker_index"])
+            for result_file in result_files
+            if (match := _RESULT_FILE_PATTERN.fullmatch(result_file)) is not None
+        }
+        missing_workers = sorted(set(range(workers_total)) - reported_workers)
+        if not missing_workers:
+            return
+        log.info(
+            f"Waiting for {len(missing_workers)} of {workers_total} workers to "
+            f"complete; missing worker indexes: {missing_workers[:10]}"
```
aldenks:
```suggestion
            f"complete; missing worker indexes: {missing_workers[:100]}"
```

aldenks:
Validation for 5070e29c, run on a local merge with main (post-#1040/#1045) so the pinned toolchain matches CI — a branch predating #1044 syncs the *old* `ty`/`icechunk` pins, and `uv sync` cannot fix that because the lockfile itself is the stale thing.

| step | result |
|---|---|
| `ruff check` | clean |
| `ruff format --check` | 504 files already formatted |
| `ty check` (0.0.78) | all checks passed |
| `pytest -m "not slow" -n 4` | 2820 passed |
| `parallel_writes_test.py` (slow) | 21 passed |
| `dynamical_dataset_test.py` (slow) | 30 passed |

The two slow modules are the ones in the blast radius: `parallel_writes_test.py` is the e2e module for `wait_for_workers`, and `dynamical_dataset.py` is its real caller. CI runs the full suite with no marker filter, so the remaining slow modules are covered there.

Notes on the test itself, since the cap moved from 10 to 50:

- The old `assert "10, 11" not in caplog.text` had become the exact negation of the assertion above it. It is replaced rather than deleted — 60 workers against the 50 cap, asserting the count reports all 60 missing *and* the list stops at 50. Both together are what pin the cap; deleting the line would have left nothing pinning it, so a later change to 500 or to no cap would pass silently.
- Verified the test can fail: cap 500 fails, cap removed fails, cap 50 passes.
- It matches on `caplog.messages` rather than `caplog.text` because the `parallel_coordination.py:147` prefix contains digits that collide with the numbers under test, which would make the assertion pass or fail for reasons unrelated to the cap.
- Swept all of `src/` and `tests/` for anything else pinned to the old cap: nothing. The other tens are unrelated — the 10-second poll interval and an icechunk `retry(max_attempts=10)`.

PR #1037, opened 2026-09-07
Parse the AIFS index without pandas dtype inference

aldenks:
An AIFS index parsed through pandas can infer byte-range column dtypes across rows, coercing a boolean offset or length to an integer and producing a clean-looking reference to the wrong bytes; this was reproduced through the production `file_refs` path. Parse each JSON-lines row directly to preserve raw scalar types, strictly validate byte ranges, and preserve exact-integral level matching.


aldenks:
this is purely theoretical hardening against ECMWF one day publishing a mix of integer and non integer offset and length values in their indexes, highly unlikely and if they did our operational validators would catch it

PR #1048, opened 2026-09-08
NOMADS-only HRRR 18-hour virtual product with a moving 3-day window

mrshll:
> **2026-09-21 — being split, per Marsh.** The mirror now ships first as #1095. The product, rebuilt as one commit stacked on that branch (6212bc5c merged with current main, template regenerated, full suite 3531 passed), is on `feat/hrrr-nomads-product` at 2c288581. This PR's branch is unchanged at 6212bc5c and everything below still describes it, mirror included. Pending Marsh's choice: replace this PR's head with 2c288581 and rebase it onto `feat/hrrr-nomads-mirror`, or open a fresh PR from `feat/hrrr-nomads-product` and close this one. The old tip is kept as `backup/1048-pre-split-6212bc5c`.

# NOMADS-only HRRR 18-hour virtual product with a moving 3-day window: `noaa-hrrr-forecast-18-hour-virtual-fast`

Implements dynamical-org/meta#202 under the decision of 2026-09-17: `noaa-hrrr-forecast-18-hour-virtual-fast` reads only the rolling NOMADS mirror in Cloudflare R2 and keeps a moving 3-day `init_time` window. `noaa-hrrr-forecast-18-hour-virtual` is untouched and stays the only NODD product, with the history. Here `-virtual-fast` means NOMADS-sourced; on the closed #740/#821 it meant a reduced variable set, which this is not.

**A moving start date is a deliberate property of this product.** The first `init_time` advances every hour and older forecasts are removed. The dataset description, `time_domain` and the `init_time`/`valid_time` statistics say so, and `docs/virtual_datasets.md` documents moving windows generically (it names no data product; this product's specifics live in its code, the ops card and this description); the STAC item must too (step 3 below).

## Needs Marsh

Already done by Marsh: the bucket, the secret, lifecycle rules (4-day expiry as of 2026-09-17, 1-day multipart abort), the custom domain with public access, default caching left on.

1. ~~Expiry margin.~~ **Done 2026-09-17:** Marsh replaced the bucket's `expire-3d` rule with `expire-4d` (verified with `wrangler r2 bucket lifecycle list`). A 72 h expiry under a 72 h window was a negative margin: init *I* leaves the store at about *I* + 73 h 50 m, while its f00 object, uploaded at about *I* + 50 m, would have expired at *I* + 72 h 50 m. With 4 days the window rides out about 23 h of update outage. Still to check once the mirror cron runs: `x-amz-expiration` on a freshly copied object is about 4 days out.
2. **Create the store** after merge with one `create-new-store` backfill (Manual: Backfill) whose `append_dim_end` is the update cron's most recent fire time (`…HH:50`); with the default end, a run between :00 and :49 creates a store one hour ahead of the template and the next validation fails its alignment assertion until the :50 fire realigns it. It is the only backfill this product supports. It is mirror-fed, fills whatever the mirror still holds, persists the single container, and takes minutes. There is no NODD backfill. No store exists at `s3://dynamical-noaa-hrrr/noaa-hrrr-forecast-18-hour-virtual-fast/v0.1.0.icechunk` today (checked anonymously 2026-09-17), so nothing migrates. Keep the update cron from firing into the backfill (see the unsuspend/catch-up hazard).
3. **dynamical-stac**: a staging `CatalogItem` with `virtual_chunk_container_prefixes=("https://noaa-hrrr-nomads-mirror.r2.dynamical.org/",)` and a description that states the moving 3-day window, that long-lived readers reopen `main`, and that `noaa-hrrr-forecast-18-hour-virtual` holds the history.
4. **Snapshot expiry and GC for this repo.** Dropping refs from `main` does not reclaim old manifests, and every snapshot older than the mirror's expiry references deleted files. Not in this PR.
5. **Merge = deploy** of the mirror cron and the fast update/validate crons. Then measure per init and file type: NOMADS mtime → mirror → committed ref → public decode, against the regular product, before calling it faster; and time the hourly drop on the filled store (it rewrites every retained ref's manifest entry; tested here only on small stores). The wxopticon follow-on is not started here.

The "reopen main / repoint" contract question of the previous design is gone: nothing is repointed.

## Design

Three single-pod crons.

- **Mirror** `noaa-hrrr-nomads-mirror-gribs` (`noaa/hrrr/nomads_mirror.py`, unchanged in this redesign): copies each init's sfc/prs/nat f00–f18 GRIBs and NOAA's `.idx` into R2 under NODD's keys, a GRIB only once it parses whole and an index only once it agrees with the mirrored GRIB, through R2's signed S3 API.
- **Update** (`forecast_18_hour_virtual_fast/region_job.py`): lists and reads the mirror through the signed store (listing, index download, GRIB header reads); refs point at the public domain, the dataset's only virtual chunk container. A file whose index lacks a message the template expects is skipped and re-offered, never ingested in part. Each fire sweeps every position of its template once for missing files, so gaps the mirror can still fill heal after an outage, then polls only the most recent 6 h, so a file that never arrives neither costs listings nor holds the run open until the deadline.
- **Validate**: `CheckCurrentData`, `CheckVirtualManifestCompleteness` and `CheckVirtualDecodeHealth` over the most recent 6 h (so a file NOMADS never published alerts for hours, not three days), plus the new **`CheckMirrorWindow`**: for one instant variable per file type, the oldest chunk the window holds decodes over the public domain (positions with no ref read as fill without a fetch, so the walk costs one chunk per file type). A window that stopped moving is caught before any check runs: the validation job asserts the store's labels equal the template of the most recent update fire, so a missed drop fails validation on that assertion (ops card).

### The moving window

- **Template**: `append_dim_coordinates(end)` starts at `floor(end, 1h) − 72 h` (never before `append_dim_start`). A fire at H:50 gets 73 labels, H − 72 h through H. Update and validation both build their template from the update cron's previous fire time, so they agree for the whole hour once that fire's drop has run.
- **Drop** (`VirtualRegionJob.drop_before_template_start`, opt-in by `drops_before_template_start`): at the start of each update fire, before `refresh_metadata` and before anything probes the manifest, one commit of its own removes the positions before the template's first label: `Session.shift_array` moves every stored data array's refs by −k along `init_time` (manifest-only), coordinate arrays are moved with raw zarr reads and writes, every append-dim array shrinks. That restores "store position *i* is template position *i*", so `refresh_metadata`, `read_ingested_through`, `_committed_append_dim_size`, `sync_dims_to` and `_emit_refs` are unchanged. A store that is not an earlier window of the template (ahead of it, or different labels) raises; an empty store is left alone; dropping everything and regrowing leaves no old refs behind. Single repository only (asserted).
- **Why its own commit** rather than one commit with the appends: each state is consistent for readers, expired inits leave at once instead of waiting minutes for the first new file, and the write loop stays as it is.
- **Writer fencing**: icechunk's `ConflictDetector` rebase silently replays a stale session's positional `set_virtual_refs` over another writer's shift (reproduced locally in 2 of 3 cases). So every update write of a moving-window job, the metadata refresh and each ref batch, asserts alignment on its own writable session and commits without rebase, one attempt (a failed pod is retried by the job's backoff and resumes from the manifest); a pod replaced by the next fire fails instead of writing refs or coordinates at the wrong labels. Backfill workers keep rebasing: they share a temp branch nothing drops from. Validation asserts alignment before any manifest probe.

### Checked first

- All 176 data arrays are regular-grid with one `init_time` position per chunk (142 4-D, 34 5-D). The `init_time`-dependent coordinates (`init_time`, `valid_time`, `expected_forecast_length`, in each of 3 groups) are single-chunk, so they are rewritten, not shifted.
- `shift_array` gives identical results with no manifest split, split 4 (shifts crossing split boundaries) and split 1500. The dataset keeps the 18-hour product's explicit per-group split map because a repo policy test requires one; the window never reaches any split, so each array has one manifest.
- Shift, shrink, regrow and `set_virtual_refs` in one session commit cleanly; a plain commit from a pre-shift session raises `ConflictError`.

## Diffs outside the dataset, each justified

- `common/virtual_region_job.py`: `drops_before_template_start`, `drop_before_template_start`, `assert_aligned_to_template`, the fenced commit in `process_virtual`. Off by default; no other dataset changes behaviour.
- `common/storage.py`: `commit_if_icechunk(..., rebase=True)`; `rebase=False` commits once with no rebase.
- `common/dynamical_dataset.py`: the drop runs before `refresh_metadata` in the virtual update; the validation job asserts alignment.
- `common/template_utils.py`: `refresh_store_metadata(..., aligned_size=)`, the same fence for the metadata refresh.
- `noaa/hrrr/nomads_mirror.py` (new) and `__main__.py`: the mirror and its registration; the dataset's registry entry.
- `common/download.py`: `httpx_get_text` and `signed_s3_store`.
- `noaa/noaa_virtual_region_job.py`: `download_index(coord)` and `read_data_bytes(coord, start, end)` hooks around the two source reads in `file_refs`, so a coord can be read from another store.
- `noaa/hrrr/virtual_region_job.py`: a `source_file_coord_class` ClassVar. `noaa/hrrr/region_job.py`: `NoaaHrrrSourceFileCoord.relative_path()`, the mirror key.
- `CLAUDE.md`: the reader recipe hands `HttpAccess` to `https://` containers.
- `.github/workflows`: regenerated choices.

Removed relative to the previous push: `SourceBucket` and coord routing, NODD-first `discover_available`, pending-repoint records and their re-offer, `CheckMirrorRefsRepointed`, the NODD container, `StoreFactory.delete_coordination_file`, the ops card's manual repoint repair.

### 2026-09-08 — first design (superseded)
Two processes, wgrib2-built indexes, repoint twins with markers and tag-filtered expiry.

### 2026-09-11 — reworked per Alden's review (superseded in part)
Three processes, NOMADS' own indexes, no tagging, 3-day age-out, as an in-place change to the existing 18-hour dataset on an AWS mirror bucket.

### 2026-09-16 — R2 mirror, separate fast product (superseded in part)
Mirror moved to R2 behind a signed store; separate `-virtual-fast` dataset; NODD-first routing with pending-repoint records. The mirror, the signed reads, the `file_refs` hooks and the partial-file guard survive; the routing and repoint machinery does not.

### 2026-09-17 — NOMADS-only, moving window
Redesigned per Marsh: mirror-only, 3-day window, NODD product keeps the history. Check-first findings above. Second-look (Codex) round 4 accepted the separate drop commit and required writer fencing, `k < 0` as an error, store-driven array iteration, validation alignment, and a full-window sweep per fire; round 5 rejected an HTTP HEAD as a stand-in for decoding the oldest init, so `CheckMirrorWindow` decodes. Its points deferred to Marsh: snapshot expiry/GC, etag checksums on refs, a drop benchmark on a filled store. Its concern that a partial GRIB/index pair that agrees could stay in the mirror forever stands as a residual risk of having no NODD fallback: the update skips such a file every fire and the completeness check alerts for 6 h.

### 2026-09-17 — dual-vendor review
Claude Opus and Codex both found the drop itself correct (label arithmetic, datetime units, empty store, groups, axis handling, no change for other datasets) and disagreed on the metadata refresh: Opus judged its rebase safe, Codex showed a stale writer's refresh could write an old window's coordinates over a newer drop. Took Codex's side and fenced the refresh (test: a stale refresh fails with "another writer moved the window"). Also fixed: the update region skipped the template's oldest position (both); `CheckMirrorWindow`'s lag and length branches could never run because validation's alignment assertion fires first, so they are gone and the docs name the assertion as the symptom (Opus); the hourly scan of about 2.3 million ref locations is replaced by a bounded decode walk (Opus); a never-arriving old file held every fire open to its deadline, fixed by sweeping once and then polling a recent subset (Opus); backfill end time guidance (Opus); 74 h → 73 h in the docs; two NODD-fallback leftovers in `nomads_mirror.py` (both). Not changed: the single commit attempt under the fence; the job's pod backoff retries and resumes.

Tests: full suite 3500 passed, 13 skipped (`uv run pytest`); `ruff format`, `ruff check`, `ty check` clean. New: moving-window tests in `tests/common/virtual_region_job_test.py` (drop then append, repeat is a no-op, drop everything then regrow, store ahead of template raises, stale writer refuses, commit does not rebase over a drop, stale metadata refresh refuses, validation alignment), a multi-group drop in `tests/common/virtual_multi_group_test.py`, and the fast dataset's tests including an end-to-end update over the real GRIB fixture that ingests, moves the window, and decodes at the new position.

### 2026-09-17 — docs made product-neutral
Per Marsh: removed the "NOMADS mirror" and "R2 setup" sections from `docs/virtual_datasets.md` and stripped the product's details from "Moving window", which now describes the mechanism, fencing, reader contract and expiry-margin rule for any dataset.

### Retro
Re-establishing "store position i is template position i" at the start of each fire, rather than teaching the write loop about an offset, kept the redesign to one primitive plus a fence and left every other base-class path alone. The expensive lesson was icechunk's rebase: a bare commit from a stale session conflicts, but the `ConflictDetector` rebase the repo uses everywhere replays positional refs silently, which only a probe showed; the refresh path needed the same fence and only the second reviewer saw it. The validator went through three shapes (HEAD, location scan, decode walk) because I wrote it before deciding that the alignment assertion, not a named check, owns "the window stopped moving".


aldenks (commented):
Design:

3 processes, each their own single-pod and cronjob returned from operational kubernetes resources:
- Nomads Mirror: polls nomads, copies files (gribs + indexes) to our mirror bucket
- Update: virtual update job. standard with one tweak described below.
- Validate: standard virtual validate 

Our virtual update job can we pretty darn standard, I don't think we need any of the common code changes. Instead, our virtual hrrr 18h subclass can override filter_already_present to keep offering coords unless they are in the _final_ bucket. [Ideally discover_available would then check the store and not re-yield refs to the mirror bucket that the store already has, but only if thats not too complicated. Worst case we re-set the same refs again -- that's idempotent.] 

Draft PR comments:
- Overall comment: current PR diff touches so many common files. After rework, double check each of those diffs is really worth having at all, most are not. This should be primarily a single dataset PR.
- Rebase on latest retry code on main (if we need a shorter max retry time, just set max attempts)

To the Open questions
1. Two options I like equally, pick whichever implementation is clearest and simplest:
     a. `dynamical-noaa-nomads-mirror` for bucket name, and then i would use the exact, full nomads path copied over. e.g. if `https://nomads.ncep.noaa.gov/pub/data/nccf/com/hrrr/prod/hrrr.20260911/conus/hrrr.t00z.wrfnatf00.grib2.idx`  then the key in the bucket is `pub/data/nccf/com/hrrr/prod/hrrr.20260911/conus/hrrr.t00z.wrfnatf00.grib2.idx`. 
    b. copy the NODD S3 bucket structure. In this case I would make a _per model_ bucket so we are only changing the bucket and the key is equal when we repoint. aka the ref that will ultimately point to `s3://noaa-hrrr-bdp-pds/hrrr.20260911/conus/hrrr.t00z.wrfnatf01.grib2.idx` initially points to `s3://dynamical-noaa-hrrr-nomads-mirror/hrrr.20260911/conus/hrrr.t00z.wrfnatf01.grib2.idx`

2. I'd run the normal bucket setup script, then go override to add explicit age out after 3 days. This also enables access logs which is good

3. Access with IAM in existing secret will already work

4. Tagging doesn't seem worth the complexity, let it age out and our validators find a broken ref

5. Proposed policy is good, deploy stac first. to MM's point, no this does need to be a new container, icechunk needs one per source bucket.

6. We don't need wgrib2, lets revert all that. The indexes on nomads can be trusted, lets copy them too

7. We should be polling _every second_ otherwise there's no way to get to our < 6s total end to end latency target (that's from "file appears on nomads" to "ref commited in virtual" so needs to include the polling nomands, copy to mirror, polling mirror in virtual update job and commit in virtual update job). That plus the grib + the index will make the rate limit much tigher but should be ok. usually 2 or fewer files per minute + 60 polls is still well under 100/min. The real nomads rate limit is 120/min, but if you hit it you are banned for at least 5 mins so we want a little buffer. could probably go up to 110/min

8. No, include all 3 file types srf, prs, nat. Include in rate limit check. Bandwith is not going to be an issue.

mrshll:
Thanks Alden — reworked to your design in the latest push (three single-pod crons: mirror, standard update with overrides on the 18-hour subclass, standard validate; rebased on main).

**Draft PR comments.** Common-code changes are gone except two small ones, each justified in the PR body: `httpx_get_text` in `download.py` (a directory listing needs the same limiter/retries as a download and `httpx_download_to_disk` cannot write a directory URL to disk) and `persist_virtual_config()` before each update's refresh (so the new container reaches the repo config; only backfill setup persisted it). Retries are main's exponential backoff, unchanged; no shorter budget turned out to be needed. The `filter_already_present` override keeps offering a present file while the mirror still lists it (so the update rewrites it from NODD on the next fire), and `discover_available` only ever resupplies such a file from NODD, so refs already in the store are never re-yielded to the mirror; the rewrite is idempotent when they already point at NODD. The override is on the update job only; validation keeps the plain manifest probe.

1. **1b**: per-model bucket `dynamical-noaa-hrrr-nomads-mirror`, NODD key structure. Simplest because the repoint is then a pure prefix swap and one key function (`coord.relative_path()`) serves both buckets.
2. Normal bucket script, then the 3-day expiry override; exact commands are in the PR body and `docs/virtual_datasets.md`. Not run: waiting on Marsh.
3. Yes — the mirror writes with the existing secret's principal (needs Put/List on the new bucket); reads are anonymous.
4. Tagging, markers and the repoint hooks are gone; files age out after 3 days and `CheckVirtualDecodeHealth` finds a broken ref.
5. STAC first, then this PR; the mirror is a second virtual chunk container on the 18-hour dataset only.
6. All wgrib2 work reverted; the mirror copies NOMADS' `.idx` as published (GRIB first, index only once the GRIB is in the mirror).
7. Polling once a second: one NOMADS directory listing per poll covers all files, so ~60 listings plus a few copies per minute, all through the shared 100/min limiter. Note the listing is ~360 KB per request (~20 MB/min of listing traffic).
8. sfc, prs and nat, f00–f18, all under the same limiter.

Also fixed: `test_backfill_local_and_operational_update` no longer touches the live bucket (it injects a local mirror store).

docs/virtual_datasets.md
mrshll, line 1: urgh

PR #1049, opened 2026-09-10
Back off exponentially between retries

aldenks:
## What

Follow-up on [REFORMATTERS-JW](https://dynamical.sentry.io/issues/7714868749/): `PutObject` returning `InternalError` while copying metadata in the `nasa-smap-level3-36km-v9` update.

We *are* retrying that exception — `sync_to_store` wraps the write in `retry(..., max_attempts=6)` and `OSError` is retryable by default — but the retry window was far too short to be worth anything against a provider 5xx:

- `retry` slept `attempt * ~1s + 0.1`, so six attempts spanned ~10s of sleep total.
- By the time one of our attempts fails, s3fs/botocore has already exhausted its own 4 retries (`reached max retries: 4` in the event). Our layer's only job is to wait meaningfully longer, and ~10s isn't that.

## Changes

- One shared backoff schedule, `exponential_backoff_time(attempt) -> float`: 0.2s, then 1s doubling to a 16s cap, each jittered down to as little as half its value. Callers do the sleeping. `_exists_many` and `_httpx_get_with_retry` retry a subset of their work so they can't call `retry` directly, but they no longer carry their own backoff formulas.
- `sync_to_store` goes to 10 attempts: ~40–80s of sleep across attempts (plus each attempt's own client-level retries) instead of ~10s. It's the last-mile write that publishes an update, so a minute of retrying is cheap.

## Caveat on the linked issue

This does not claim to fix that particular run. The Sep 9 failures were six pod attempts of one k8s job (`backoffLimitPerIndex: 5`) failing at 06:02, 06:05, 06:08, 06:11, 06:18 and 06:23 — the same write failed across 21 minutes, longer than any sane in-process backoff. The 18:00 run succeeded. Sep 6 looks like the case this does help: two failures three minutes apart, then a later pod attempt succeeded.

Two things worth noting separately, not changed here:

- The store is Cloudflare R2 (`upstream-gridded-zarrs`), not S3, and `InternalError` on `PutObject` is a documented R2 failure mode; the 21-minute span looks like an R2 episode rather than a code fault.
- A `_copy_metadata_files` failure part-way through leaves a zarr v3 store with some arrays' `zarr.json` extended and others not. There is no crash-safe ordering for non-atomic metadata — that's what icechunk solves — and this dataset lives in `contrib/` so it isn't in the published catalogs, but it's why the retry window on this specific write matters.

## Testing

`uv run ruff format && uv run ruff check && uv run ty check` clean; `uv run pytest -m "not slow"` (2822 passed) and the slow tests under `tests/common/` (124 passed). Retry tests record sleeps instead of taking them, and pin the backoff schedule and its jitter band.

https://claude.ai/code/session_012z3FKHrDZbqJPDSG1gnRdq

### DIFF
--- a/src/reformatters/common/retry.py
+++ b/src/reformatters/common/retry.py
@@ -3,13 +3,22 @@ from collections.abc import Callable
 
 import numpy as np
 
+_INIT_BACKOFF_SECONDS = 1.0
+_MAX_BACKOFF_SECONDS = 16.0
+
+
+def sleep_before_retry(attempt: int) -> None:
+    """Sleep after the given 0-indexed failed attempt, backing off exponentially with jitter."""
+    backoff = min(_INIT_BACKOFF_SECONDS * 2**attempt, _MAX_BACKOFF_SECONDS)
+    time.sleep(backoff * np.random.default_rng().uniform(0.8, 1.2))
+
 
 def retry[T](
     func: Callable[[], T],
     max_attempts: int = 6,
     retryable_exceptions: tuple[type[Exception], ...] = (Exception,),
 ) -> T:
-    """Simple retry utility that sleeps for a short time between attempts."""
+    """Retry utility that backs off exponentially between attempts."""
     last_exception = None
     for attempt in range(max_attempts):
         try:
@@ -17,7 +26,6 @@ def retry[T](
         except retryable_exceptions as e:
             last_exception = e
             if attempt < max_attempts - 1:  # sleep unless we're out of attempts
-                rng = np.random.default_rng()
-                time.sleep(attempt * rng.uniform(0.8, 1.2) + 0.1)
+                sleep_before_retry(attempt)
 
     raise last_exception or AssertionError("unreachable")
--- a/src/reformatters/common/retry.py
+++ b/src/reformatters/common/retry.py
@@ -3,14 +3,14 @@ from collections.abc import Callable
 
 import numpy as np
 
-_INIT_BACKOFF_SECONDS = 1.0
 _MAX_BACKOFF_SECONDS = 16.0
 
 
-def sleep_before_retry(attempt: int) -> None:
-    """Sleep after the given 0-indexed failed attempt, backing off exponentially with jitter."""
-    backoff = min(_INIT_BACKOFF_SECONDS * 2**attempt, _MAX_BACKOFF_SECONDS)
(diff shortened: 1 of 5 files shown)


src/reformatters/common/retry.py
```diff
@@ -3,21 +3,29 @@
...
 import numpy as np
 
+_INIT_BACKOFF_SECONDS = 1.0
+_MAX_BACKOFF_SECONDS = 16.0
+
+
+def sleep_before_retry(attempt: int) -> None:
+    """Sleep after the given 0-indexed failed attempt, backing off exponentially with jitter."""
+    backoff = min(_INIT_BACKOFF_SECONDS * 2**attempt, _MAX_BACKOFF_SECONDS)
+    time.sleep(backoff * np.random.default_rng().uniform(0.8, 1.2))
```
aldenks: this function should return the float seconds to sleep, not actually sleep (callers do that).  `def exponential_backoff_time(attempt: int) -> float`

equation: sleep time = uniform(0.5 max delay, max delay):
 - if attempt == 0: max delay = 0.2s
 - else max delay = min(2^(attempt - 1), 16)

aka 0.2 1 2 4 8 16 16 16 ... with random jitter from 50% to 100% of those values

src/reformatters/common/zarr.py
```diff
@@ -225,7 +225,8 @@ def sync_to_store(store: Store, key: str, data: bytes) -> None:
             ),
             timeout=90,  # In seconds. Timeout needs to be long enough to upload a large shard.
         ),
-        max_attempts=6,
+        # Transient object store 5xx bursts can outlast the default retry window.
```
aldenks: cut comment

src/reformatters/common/retry.py
aldenks: Done in 70c15a1 — `exponential_backoff_time(attempt: int) -> float` with exactly that equation, and the three callers (`retry`, `_exists_many`, `_httpx_get_with_retry`) sleep on the returned value. Test now asserts the 0.2/1/2/4/8/16/16/16 schedule and that jitter spans the 50–100% band.

---

src/reformatters/common/zarr.py
aldenks: Cut in 70c15a1.

---

PR #1054, opened 2026-09-10
Retry terminal Sentry cron check-ins

aldenks:
The 06:20 UTC HRRR analysis virtual validation completed successfully after #1050 was deployed, but Sentry again retained only its `in_progress` check-in. A longer flush wait cannot recover an HTTP send that the Sentry SDK has already dropped; the SDK does not retry failed envelopes. This matches getsentry/sentry-python#5423.

Send each terminal `ok` or `error` update twice with the same deterministic check-in ID, flushing for up to 15 seconds after each attempt. The repeated update is idempotent in Sentry and gives a transient failed send another chance without creating another logical run.

Validation:
- `uv run pytest tests/common/monitoring_test.py`
- `uv run ruff format`
- `uv run ruff check --fix`
- `uv run ty check --output-format concise`

### DIFF
--- a/src/reformatters/common/monitoring.py
+++ b/src/reformatters/common/monitoring.py
@@ -46,21 +46,26 @@ def monitor_cron(
     monitor_slug = os.getenv("CRON_JOB_NAME") or cron_job.name
 
     def capture_checkin(status: Literal["ok", "in_progress", "error"]) -> None:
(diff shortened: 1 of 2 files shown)


aldenks:
a sentry incident prompted this. seems not needed now and would result in lots of warnings about double delivered events.

PR #1073, opened 2026-09-17
AIFS Single virtual: snow area fraction has values over water at forecast hour 0

mrshll:
Corrects the `comment` on `snow_area_fraction_surface` in `ecmwf-aifs-single-forecast-virtual`. It said "Applies over land only; NaN over water", which is not true at forecast hour 0.

New text: "Fraction (0-1) of the grid box covered by snow. Applies over land only; NaN over water, except at forecast hour 0, which has values over water: 0 over open ocean and up to 1 in polar seas."

## Evidence

Read from the live v0.2.0 store on 2026-09-17, using the hour-0 land-sea mask to pick water cells. At hour 6 the field is NaN on 99.5% of water cells. At hour 0 it has no NaN cells anywhere:

| Forecast | Water cells equal to 0 | Equal to 1 | Between | Where the 1s are |
|---|---:|---:|---:|---|
| 2026-05-13T00 | 82.0% | 9.6% | 8.4% | 76.5°S to 90°N |
| 2026-06-01T00 | 81.9% | 9.9% | 8.3% | 76.5°S to 90°N |
| 2026-09-01T00 | 85.4% | 3.8% | 10.8% | 78.2°S to 59.5°S |
| 2026-09-17T00 | 82.8% | 4.1% | 13.1% | 78.2°S to 58.8°S |

Water between 40°S and 40°N is exactly 0 in all four. A chunk of this archive is ECMWF's own GRIB message, so this is how ECMWF publishes the hour-0 field, not something ingestion introduced. 2026-05-13T00 is the first forecast that carries the variable, so the behaviour holds for its whole record, which is why it belongs in the `comment` and gets no validation-report review note.

The wording differs from the one first proposed ("1.0 over water"), because most water cells at hour 0 are 0, not 1.

## Scope and compatibility

One attribute on one variable, plus the regenerated template. No data, encoding or fill value changes. The dataset is in the staging catalog only, so the production compatibility policy does not apply yet; this should land before dynamical-org/dynamical-stac#108 promotes it.

## After merge

The live store keeps the old comment until its metadata is refreshed: Manual: Backfill with `dataset_id` `ecmwf-aifs-single-forecast-virtual` and operation `overwrite-metadata`, which rewrites metadata in place and launches no workers. Then regenerate the STAC for #108, since the collection's variable table is read from the store, and re-render the validation report draft so its variable section shows the new comment.

## Checks

`uv run ruff format`, `ruff check` and `ty check` are clean. `tests/common/common_template_config_subclasses_test.py`, `tests/common/datasets_cf_compliance_test.py` and `tests/ecmwf/aifs_single`: 625 passed. I did not run the full suite for a one-attribute change; CI does.


mrshll:
Applied to the live store: [Manual: Backfill run 35253444645](https://github.com/dynamical-org/reformatters/actions/runs/35253444645) (`overwrite-metadata`) committed "Refresh metadata from template" at 17:42:22Z, after the 17:20Z update job had completed. Read back from `main`: the new comment is present, both virtual chunk containers persist, the extent is intact at 3,599 init times, and a filled 2025-01-24T00 cell and the newest init both decode with 0% NaN.

PR #1074, opened 2026-09-17
Rename the ECMWF 46-day CAPE variable to name its surface level

aldenks:
Renames the ECMWF 46-day daily CAPE variable so its suffix stops claiming a column integral.

```
average_convective_available_potential_energy_atmosphere
  ->  average_convective_available_potential_energy_surface
```

`_atmosphere` is used in this catalog for whole-column integrals — `precipitable_water_atmosphere`,
`total_cloud_cover_atmosphere`, `cloud_work_function_atmosphere` — **but not exclusively**: DWD's
`dwd-icon-eu-forecast-5-day` carries `convective_available_potential_energy_atmosphere` on ICON's
`cape_con`, which is no more a column integral than ECMWF's. So the suffix is inconsistent across the
catalog, and an earlier revision of this description claimed it was reserved for integrals, which is
wrong. The argument for this rename does not rest on that claim.
ECMWF's CAPE is a single-parcel energy carried on a GRIB record whose level is
`SFC "Ground or water surface"`: per ECMWF's parameter database (param 59) it considers "parcels of
air departing at different model levels below the 350 hPa level… The maximum CAPE produced by the
different parcels is the value retained." `_surface` matches the source record and lines the variable
up with the NOAA and ECCC CAPE variables that already use it.

## ⚠️ This breaks backwards compatibility on a dataset in the production catalog

`ecmwf-ifs-ens-forecast-46-day-daily-1-5-degree` **is** in `stac.dynamical.org`. Alden's ruling:
the dataset has only just entered the production STAC, is unannounced, and we should break
compatibility now rather than later.

**Merging this without the migration below stops the daily update.** The drift guard
(`template_utils.structural_mismatches`) walks every variable in the *store* and requires each to be
present in the template; additions to the template are free, removals are not. After deploy the live
store still holds `_atmosphere`, which the new template does not, so worker 0 raises and the 09:00Z
update stops advancing the archive.

## Migration — one metadata commit, no data movement

Live store today: snapshot `JQJ53ZKXY8JYM0QCN6B0`, **258 init times** 2026-01-01 → 2026-09-15,
`dataset_version` 0.2.0, 35 data variables. The CAPE array is `(258, 47, 101, 121, 240) float32` —
142 GB uncompressed.

**Icechunk can rename the array in place.** `repo.rearrange_session("main").move(old, new)` followed
by a commit is metadata-only: the node id is preserved, the chunk data is untouched, and attributes
come with it. Verified on a local icechunk repo — data byte-identical after the move, attrs
preserved, sibling arrays untouched.

So the migration is **one small commit**, not a 142 GB rewrite. No backfill, no NaN window, no
validate-cron alerts.

*(An earlier revision of this description proposed delete-then-re-backfill, on the grounds that no
cheap rename existed. That was wrong: I had tested `zarr.Group.move`, which does raise
`NotImplementedError` on an icechunk store, and concluded from it that no rename existed anywhere —
without checking icechunk's own session API, where `rearrange_session` lives. Testing one mechanism
is not testing the capability.)*

## Operational sequence at merge

The daily update fires **once a day at 09:00Z** (`ecmwf-ifs-ens-46-day-daily-update`).

The drift guard compares the store against the template in one direction: every variable **in the
store** must be in the template. So there is a window between deploying the new template and renaming
the array in which the two disagree, whichever order you pick — the store holds a name the template
does not, or the reverse. **The window is minutes and the next fire is up to 23 hours away**, so do
both in one sitting:

**1. Merge and deploy**, and confirm the cronjob image tag equals the merged sha.

**2. Immediately rename the array in the live store.** Shape, using the registry instance because that
is where a dataset carries its storage config:

```python
import zarr
from reformatters.__main__ import DYNAMICAL_DATASETS

DATASET_ID = "ecmwf-ifs-ens-forecast-46-day-daily-1-5-degree"
OLD = "/average_convective_available_potential_energy_atmosphere"
NEW = "/average_convective_available_potential_energy_surface"

dataset = next(d for d in DYNAMICAL_DATASETS if d.dataset_id == DATASET_ID)
# print the store first and confirm it is the v0.2.0 production store
repo = ...  # the icechunk Repository behind dataset.store_factory
session = repo.rearrange_session("main")
session.move(OLD, NEW)
print(session.commit(f"Rename {OLD} to {NEW}"))
```

**This snippet has not been executed against production and cannot be from here**: the prod store
factory reads credentials from a Kubernetes secret and the read-only kubeconfig available to me gets
401. The `rearrange_session(...).move(...)` mechanism itself is verified on a real icechunk repo; the
wrapper around it is not. Print and confirm the store URL before committing.

**If an update fires inside the window** it halts on the drift error and loses nothing — complete the
rename and the next fire recovers.

**No backfill is needed.** Historical values move with the array, so there is no gap to fill and the
validate cron has nothing to complain about.

## Verification

- fast suite **3050 passed**; every `mark.slow` module under `tests/ecmwf` run one gated invocation
  each at `-n 3`, all passing; `ruff format`, `ruff check`, `ty` clean
- template regenerated; the stale `..._atmosphere` array directory is removed and **no reference to
  the old name survives anywhere in the repo**
- the deletion mechanism and the `Group.move` non-availability were both tested against a real local
  icechunk repository, not reasoned about

The NOAA CAPE parcel comments are in #1072 and unaffected by this PR.

## Follow-ups this PR does not do

1. **Regenerate dynamical-stac.** The live STAC collection's `cube:variables` still names
   `_atmosphere`; the catalog is the source of truth for dynamical.org.
2. **Re-publish the validation report.** The published report for this dataset also names
   `_atmosphere`.
3. **DWD ICON-EU's `convective_available_potential_energy_atmosphere`** has the same misleading
   suffix on a convection-scheme diagnostic. Separate decision; its parcel is unverified.

*(A previous revision told you to cap backfill parallelism at 8 because `s3://ecmwf-forecasts` limits
concurrent readers. That cap does not apply to this dataset, which reads the source.coop archive — its
sibling 6-hourly dataset backfilled at parallelism 50-60. The point is moot now that no backfill is
needed, but the claim was wrong and is withdrawn rather than left standing.)*


mrshll:
Merged as 54ffccf8 and the migration is done. Recording what the runbook got right, and the one thing it missed, for whoever runs the next rename of this shape.

**Executed**

- Merge 54ffccf8; `Deploy Operational Updates` green; both `ecmwf-ifs-ens-46-day-daily-{update,validate}` cronjobs now on image tag `54ffccf860951cc17ba992b8ab04ba7f062d3cf1`.
- Rename commit `63RY2TWEYHNE398JX440`, parent `JQJ53ZKXY8JYM0QCN6B0`. `repo.diff` on that pair reports exactly one entry, the move, and nothing else.
- Array metadata is unchanged beyond the name: the raw `zarr.json` for `_atmosphere` before and `_surface` after compare equal as parsed JSON — shape, dtype, chunks, shards, fill value, codecs, dimension names, attrs. Sampled cells at three init times, including the newest, read identical values. The other 43 arrays are untouched.
- The real drift guard — `structural_mismatches(template, store_factory.open_primary_datatree(), "init_time")` with the deployed template against the live store — returns no mismatches, so the 09:00Z fire is clear.

**The runbook missed consolidated metadata.** The live store's root `zarr.json` carries an inline `consolidated_metadata` block, because `copy_zarr_metadata` copies the template's root `zarr.json` into the store as raw bytes. `rearrange_session().move()` does not touch it, so it still names `_atmosphere` at a path that no longer exists.

That turned out to be harmless, but only after testing rather than reasoning. I rehearsed it on a local icechunk repo seeded the same way production is, and on the live store after the move every reader path — default, `consolidated=True`, `consolidated=False` — returns `_surface` and only `_surface`. `IcechunkStore` does not serve that stored block to zarr's consolidated path, so no reader consumes the stale key. It is inert bytes that the next update's metadata copy overwrites. I left it alone rather than add a second, non-move write to production for something nothing reads.

Worth stating because the conclusion is version-dependent, not structural: it rests on `IcechunkStore` ignoring a stored consolidated block, which is behaviour of icechunk 2.2.0 / zarr 3.3.0, not a guarantee. A future version that honours it would turn this same one-commit rename into a store whose consolidated metadata points at a missing node. Anyone repeating this should re-run the three-reader check rather than inherit this result.

Two smaller qualifications, neither affecting the outcome:

- "the prod store factory reads credentials from a Kubernetes secret and the read-only kubeconfig available to me gets 401" is true of your access, not of the path itself — the factory opened the production repo and committed from a local checkout with `DYNAMICAL_ENV=prod`, no in-cluster job needed.
- "the window is minutes and the next fire is up to 23 hours away" holds only once deploy succeeds; the window really opens when the image reaches the cluster, not at merge. It was about 25 minutes here, all of it deploy.

Follow-ups 1 and 2 (dynamical-stac `cube:variables`, validation report) are still open and now user-visible: the published catalog names a variable that no longer resolves.

PR #1078, opened 2026-09-18
Add remove-chunks: dry-run-by-default deletion of chunks or arrays from a dataset's stores

aldenks:
## Where the tool lives, and why

Alden asked whether the chunk/array removal tool belongs in `src/scripts/icechunk_utils.py` as another subcommand. **No** — and it also should not stay a loose `src/scripts/remove_refs.py`. It is now `uv run main <dataset-id> remove-chunks`, a command on `DynamicalDataset` next to `backfill-kubernetes` and `dataset-urls`, with the logic in `reformatters/common/chunk_removal.py`.

- **Rejected: fold into `icechunk_utils.py`.** That script is repository-level maintenance (list/count/expire/gc on snapshot history), addressed by raw S3 URI or by *every* catalog repo, with three auth modes (`secret`/`env`/`anonymous`) and a fixed default secret name that the WeatherNext2 docs already have to override by hand. Removal is content-level and dataset-scoped: it needs the dataset's `append_dim`, the coordinate values, each array's shard grid and dimension names, and the store factory's version resolution to open exactly the store the backfill writes. Folding it in would give that script a second addressing mode (`--dataset`) that only one subcommand accepts, an `--anonymous` mode that cannot write, and a `--force` flag where Alden wants `--apply`. The two scripts would share a file, not a shape.
- **Rejected: keep a standalone script.** `remove_refs.py` reimplemented the dataset-registry lookup that the typer app already does, and `uv run main <id> --help` is where every other dataset operation is listed. Run under `DYNAMICAL_ENV=prod` like every other dataset operation, it opens the store through the store factory with the dataset's own k8s secret — the "consistent auth" Alden asked for is the store factory's auth.
- **Where store surgery lives now:** `icechunk_utils.py` for snapshot-history surgery on any repo; `remove-chunks` for content surgery on a dataset. Different axis, so two places is not two ways of doing one thing. The one overlap left in the tree is `src/scripts/weathernext2_holdback.py`, which has its own chunk-key enumerator and batched delete loop for a selection (a per-init lead cutoff by valid time) `chunk_keys` cannot express today; folding it in is a follow-up, and `chunk_keys`'s per-axis index lists are the shape to grow toward it.

## What changed vs the restored copy

- Dry run by default, `--apply` to delete. `--whole-array` is now explicit rather than "no selector means the whole array" (a forgotten `--at` no longer plans an array deletion).
- Deletes from the primary store **and every replica** (materialized datasets have Source Coop zarr v3 replicas; a primary-only delete would have left them inconsistent). One commit per array, each verified before the next. Store I/O runs on zarr's event loop rather than `asyncio.run`, since an fsspec replica store binds to the first loop that uses it. A whole-array delete re-consolidates a zarr v3 store's metadata: zarr drops a deleted direct member from it but not a nested (vertical-group) one. The prod S3 replica path itself has no local double, so it is not covered by a test.
- Keys are built from the array's shards-or-chunks grid and its `dimension_names`, so materialized (sharded) and ensemble arrays are handled: a selection must cover whole shards (asserted with the offending shard indices), `--lead-index` finds `lead_time` by name rather than assuming axis 1, and every trailing spatial/member/level shard is enumerated.
- Read-back verification after each commit checks every planned key on every store (the restored copy sampled 50).
- Fixed the dry-run summary that hard-coded "2 array(s)".

## Docs

`docs/backfill.md` "Remove data" bullet rewritten for the new command and the rule: removal is a manual one-off, never a step of a backfill or update, run between update fires (an update whose window still covers the removed positions re-ingests them); split a very large selection rather than holding millions of deletes in one commit. `docs/ops_card.md` gains a one-line troubleshooting pointer.

## Tests

`tests/common/chunk_removal_test.py`, 25 tests: 16 key-selection unit tests on real zarr arrays (forecast, analysis, ensemble, trailing pressure_level axis, sharded, misaligned-shard refusal); 6 against a local icechunk repo built with the `virtual_region_job_test` helpers (dry run leaves no snapshot, lead-index / at / whole-array removals with value read-back and commit message, refusals); 2 on a primary + zarr v3 replica for chunk removal and for whole-array removal of a root and a `pressure_level/` array with consolidated metadata re-checked; 1 typer `CliRunner` test for argument parsing and the dry-run default.

Checks run: `ruff format`, `ruff check`, `ty check`, fast suite (3080 passed, `-n 3`), slow tests in `dynamical_dataset_test.py` + `virtual_region_job_test.py` (79 passed). Reviewed by a Claude reviewer (opus, high); its findings — fsspec event-loop binding on replicas, nested-member consolidated metadata, per-array commits, logging the opened store, selector help wording — are applied in the second commit.


aldenks:
really not sure we want this to exist so easily. Maybe one guard: it could refuse to work on anything that's in the prod STAC?  pausing on it. maybe for now it lives as a closed PR that can be pulled from when needed.

PR #1081, opened 2026-09-18
Add memory to the virtual backfill sizing guidance

aldenks:
`docs/backfill.md`'s virtual sizing guidance capped refs per commit at half a million on commit-latency grounds and said nothing about memory — the cost that actually evicts pods. This replaces it with a section that links the mechanism the repo already documents and adds the one cost it doesn't.

## What was wrong with the old text, and with my first three drafts of the new one

The old cap predates icechunk's manifest binary search; commits of 5.7M and 13.2M refs now run routinely. That much was right from the start.

What took four drafts to get right is the frame. My first versions rebuilt the whole sizing story around memory, because memory was the failure I had just watched destroy 70% of a round. The repo owner's review put it plainly: the section over-indexed on our most recent failure mode and papered over commit latency and contention, whose driver is the total refs whose manifests must be downloaded, merged, compressed and re-uploaded on each commit — and a commit that loses the branch-tip race **redoes all of that**, which under enough concurrency puts a worker in a rebase-retry loop that never converges.

Going back to `virtual_datasets.md` with that in hand, **the mechanism was already documented there** — `(1 + rebase_attempts) × flush cost`, the gridlock failure mode, the ~10-worker parallelism ceiling, the `rebase_attempts` diagnostic, and the safe recovery. I had read the manifest-splitting section beneath it for terminology and not the backfill section itself. So the right section does not restate any of that; it points at it.

## What the section now says

**Two ref counts drive two costs, and conflating them was the last serious error.** Refs *rewritten* per commit — arrays × refs per active split — drives flush cost and therefore latency and rebase; it is set by the manifest split, not `jobs_per_pod`. Refs *added* per commit — `jobs_per_pod × refs per job` — drives memory, because the worker holds what it is adding, at about 1 GiB per million. The two come apart in practice: an HRRR virtual update rewrites 12M refs on a 3.7G pod without incident because it adds only thousands, while a backfill adding 19M refs on a 15G pod lost its round. A sizer who used one number for both would over-estimate memory four-fold on the 35-day and cut `jobs_per_pod` to compensate, quadrupling commits and manufacturing exactly the contention the section warns about.

**Memory's three terms**: the ~1 GiB/M slope on refs added, a rebase re-merge worth 0.1–0.3 GiB per concurrent sibling (worst observed 41% above base), and a moving base — one discontinuity of a few percent plus slow drift of about the same total, neither understood.

**The ceiling is the node, not the request** — no limit is set, so the failure is eviction; the per-pod ceiling formula, the fact that `nodepool.yaml` makes the binding class knowable in advance, and the rule never to express a threshold as a fixed number of gigabytes.

**Where the memory request comes from**: the update `ReformatCronJob`, so changing it is a code change merged to main, and trimming a cron job's request shrinks that dataset's future backfills.

**The order to size in**, five steps: cap refs added at ~60% of the request (a 7G request → ~3.9M), derive `jobs_per_pod`, check refs rewritten against the split guidance and fix the split rather than `jobs_per_pod` if flush cost binds, raise parallelism only while max `rebase_attempts` stays small, and check pod duration last.

## What is deliberately not here

The n=2,042 commit-time census, the four-row duration table, per-sibling coefficients, and most specific figures from the two GEFS 0.5° runs. The owner asked for durable mechanisms and the sizing order with numbers as illustrative ranges, and two review passes agreed the specifics were doing less work than their word count. What survives numerically is what a sizer needs: the memory slope, the headroom fraction, and one within-run contention measurement (commit time ~7s → ~53s across the first six commits of a round at constant refs) that shows the effect without the confound of comparing two different runs.

665 words. Reviewed four times in total; the last pass, on the rework, found the refs-added/refs-rewritten conflation and is what produced the current structure.


aldenks:
One correction after handover, from the backfill still running as this merges.

The section said the memory base takes **a step of roughly 3%, once in a run**. That is right about the discontinuity but incomplete: the base also *creeps*. The 16-day run has risen a further **1.3% over the eleven hours since its step**, in movements too small and too gradual to be steps — including one this afternoon that traced 6.326 → 6.349 → 6.365 GiB across three commits on a single node over six minutes, which is a ramp rather than a jump.

Same-node controlled, so it is real rather than node mix or scatter.

It does not change the guidance and I have said why in the text: the step is worth about twice the entire subsequent drift, so a sizing that covers the step covers the drift too. But "once in a run" would have read as "and then it is stable", which is not what the data shows. Now stated as step plus drift, with both magnitudes.

PR #1091, opened 2026-09-19
Bound manifest scan probe fan-out and let it resume from checkpoints

aldenks:
Found while validating `noaa-gefs-forecast-35-day-0-5-degree-virtual`, whose whole-archive manifest scan could not finish on a 15 GB host. Two independent causes, both fixed here.

## Unbounded probe fan-out

`filter_already_present` hands a whole region job's expected source files to `_exists_many`, which gathered every probe in one `asyncio.gather`. That per-job count is small for the stores this was tuned against (the HRRR virtual store is ~100 files per job), but it is **11,222** for this dataset — 31 members × 181 leads × 2 file types. With `probe_jobs` running 48 jobs concurrently that put ~538k requests in flight.

Measured on the 268-variable GEFS 35 day store: resident size climbed 2 → 11.4 GB in 10 minutes with **zero** jobs completed, heading for an OOM. Dropping to 4 concurrent jobs was both stable and *faster* — the default was thrashing, not merely using more memory.

`_exists_many` now issues probes in batches of 1,000, so a caller's memory no longer scales with the number of keys it passes. This is a latent problem for any caller with a large key list, not just this dataset.

## Scan could not resume

The scan then still ran for hours as a single call whose result existed only in memory. Two interruptions during this validation — an OOM and a container restart at 94% — each lost the whole thing.

`scan_manifest` now takes an optional `checkpoint_dir`, scanning the archive in 90-day slices and writing each as it completes, reusing any slice already present. Exposed as `--checkpoint-dir` on both `availability` and `run-all`. Slices are keyed by dataset id, window and variable filter so one directory can hold several datasets, and are written to a temporary file then renamed so an interrupted write is never reused.

Without `checkpoint_dir` behaviour is unchanged: one unsliced pass, nothing written.

## Also

`docs/validation.md` gains the `--checkpoint-dir` option and a note to set `MALLOC_ARENA_MAX=2` on large ensemble archives — the scan runs hundreds of probe threads and glibc's per-thread arenas hold freed memory rather than returning it, which roughly doubled resident size (9 GB vs 4.9 GB for the same work).

## Testing

- `test_exists_many_bounds_concurrent_probes` pins the batching; verified it fails on the pre-fix code and passes after.
- Checkpoint round-trip (including positions that were never probed, whose absence must survive), window merging, slice boundaries, and that the variable filter keys the file.
- `uv run ruff format && uv run ruff check && uv run ty check` clean; full fast suite 3060 passed.

https://claude.ai/code/session_01LvBsUGUhqcmFnoVGR2pHSh

---
_Generated by [Claude Code](https://claude.ai/code/session_01LvBsUGUhqcmFnoVGR2pHSh)_


aldenks:
superceeded and wrapped into 1093

PR #1093, opened 2026-09-19
Make a virtual whole-archive scan survivable on a large ensemble archive

aldenks:
Found while validating `noaa-gefs-forecast-16-day-0-5-degree-virtual` — 268 variables, 8,717 init times, 31 members, 105 leads, so 6,510 expected source files per region job. Builds on #1091, which is merged into this branch.

The validation run is still in progress; this PR is the tooling that came out of it. The dataset's own validation report is separate.

## The value time series read whole vertical profiles

This one is a bug, not a tuning knob, and it made a 268-variable virtual `run-all` impossible to finish on a 15 GB host.

`_sample_virtual_points` selected both run points with a single `point` indexer (`xr.DataArray([p1, p2], dims="point")` on latitude and longitude). That is vectorized indexing, which the zarr backend cannot push down — so it read the entire `ensemble_member × pressure_level` cross-product at every sampled position, defeating the scalar level, lead and member pins alongside it.

Measured on `pressure_level/cloud_mixing_ratio` over eight positions, same selection either way:

| indexing | peak RSS |
|---|---|
| scalars only, full spatial field | 0.43 GB |
| scalars + scalar lat/lon | 0.43 GB |
| scalars + the `point` indexer (before) | **6.95 GB** |

One such variable peaked at 7.6 GB alone in a fresh process, and `LOAD_CONCURRENCY` runs two at once. Two `run-all` attempts were OOM-killed at the same place in the variable sequence, right where the `pressure_level` group begins.

Reducing with the scalars first and taking each point as its own basic selection keeps every indexer basic. Verified identical values (`array_equal`, NaN-equal) at 0.42 GB against 6.98 GB, with the documented `point`-last dim order preserved.

## Decode scan checkpointing

On this store the decode scan is the long pole of a virtual `run-all`: 20 sampled region jobs, 4 concurrent, each checking tens of thousands of references and taking tens of minutes. The whole result lived only in memory, and that cost was paid twice during this validation — once to a container restart.

A sampled region job is now the unit a resumed scan skips. The key digests everything deciding which references a job decodes and whether they pass: window, variable filter, region sample size, and the resolved checker configuration. The shared checkpoint read/write helpers move to `scan_common` so both scans use one implementation, including the write-to-temp-then-rename.

Confirmed end to end on the real store: a 3.4-hour scan replayed as `Reusing 20 checkpointed region jobs, decoding the remaining 0` in three seconds.

**Known limitation, worth a follow-up.** The key includes the scan window's `end`, which is derived from the store's committed extent. On a live store that moves with every operational update — every 6 hours for this dataset — so the checkpoints self-invalidate on that cadence. I confirmed it by computing the key across candidate ends: the stored `49928390` corresponds to `end=2026-09-19T18:00`, and the store had already advanced. The manifest scan sidesteps this by slicing into fixed 90-day strides, so only its last window churns. Dropping `end` from the decode key looks right — a per-job checkpoint records whether *that* job's references decoded, its identity (region span + variable digest) is already in the filename, and on an append-only store region indices from the start are stable — but it would invalidate the checkpoints this validation is still using, so it is deliberately not in this PR.

## `--probe-workers` on the manifest scan

`probe_jobs` fixes its fan-out at 48 region jobs. Each in-flight job pins a different `init_time` manifest split, and a root split on this store holds roughly 4.7M refs against a ref cache sized (per `utils.py`'s own comment) to hold one, so manifests are refetched instead of reused.

Like for like, one whole 360-job checkpoint window each with nothing else running:

| `--probe-workers` | window | peak RSS |
|---|---|---|
| 48 (default) | 23.4 min | 7.2 GB |
| 8 | 20.1 min | 3.6 GB |

Memory is the reason to lower it; the time difference is small. The default is unchanged, so no other dataset's scan moves — an archive with few files per job (the HRRR virtual store is ~100) never hits this. An earlier commit on this branch quoted much larger gains, comparing a 200-job benchmark at 8 workers against a 360-job live window at 48; the follow-up commit replaces those with the table above.

## `batch_size` on `_exists_many`

The probe fan-out was fixed at the module default, so a call site with a different latency or memory profile than the write path could not tune how many probes it has in flight. Now a keyword argument defaulting to the same constant, leaving every existing caller unchanged.

## Not included

`CheckVirtualDecodeHealth` spends its time in `reference_exists`, which probes one key per call (`_exists_many(store, [key])`, a fresh event loop each time) — py-spy showed 12 of ~18 sampled threads there and 1 in actual decoding. Batching those probes looks like the real speedup, but it changes shared validation internals and getting it subtly wrong would turn a real decode failure into a false pass.

Separately, the manifest scan's per-variable pass probes the **middle** level of a variable's vertical dim and does not honour `--level`. On this store that makes `pressure_level/icing_probability` and `icing_severity` report 0 of 8,717 positions complete, because GEFS publishes icing only at 800–300 hPa and the middle of 31 pressure levels is 350 mb. Verified against the store: at 350 mb both are all-NaN, at 500 mb both carry 259,920 finite values over their full declared ranges. The data is fine; the availability figure is not. Fixing it needs the probe level in the manifest checkpoint key, which is a checkpoint format change and belongs in its own PR.

## Testing

- `uv run ruff format --check`, `uv run ruff check`, `uv run ty check` clean.
- `uv run pytest tests/scripts/validation tests/common/virtual_region_job_test.py -m "not slow"` — 76 passed.
- New: decode-scan checkpoint round-trip including a failing job, resume skipping recorded jobs, key separation across sampling parameters and variable filters, partial writes not reused, and an explicit `batch_size` bounding probes in flight.
- The memory and `--probe-workers` figures are measurements against the real store, not synthetic benchmarks.

https://claude.ai/code/session_01CBH7QdrnjsH7Aq3xwZv2Pd


src/reformatters/common/virtual_region_job.py
```diff
@@ -716,19 +716,34 @@ def _chunk_key_encoder(
...
+_PROBE_BATCH_SIZE = 1_000
+
+
 def _exists_many(
-    store: IcechunkStore, keys: Sequence[str], *, max_attempts: int = 8
+    store: IcechunkStore,
+    keys: Sequence[str],
+    *,
+    max_attempts: int = 8,
+    batch_size: int = _PROBE_BATCH_SIZE,
```
```suggestion
    batch_size: int = 1_000, # bound memory + manifest cache locality
```

src/reformatters/common/virtual_region_job.py
```diff
@@ -716,19 +716,34 @@ def _chunk_key_encoder(
     )
 
 
+# Bounds the probes in flight per call, so pending requests do not scale with the
+# number of keys passed (the result still does). A whole-archive scan probes every
+# expected source file of a region job at once -- tens of thousands for a large
+# ensemble archive, times the number of jobs scanned concurrently.
+_PROBE_BATCH_SIZE = 1_000
```
aldenks: Cut constant and comments

src/reformatters/common/virtual_region_job.py
aldenks: Applied in bb3de15d: the default is inlined as `batch_size: int = 1_000,  # bound memory + manifest cache locality`.

src/reformatters/common/virtual_region_job.py
aldenks: Applied in bb3de15d: `_PROBE_BATCH_SIZE` and its comment block are gone; the test uses the literal 1_000.

PR #1095, opened 2026-09-21
Mirror NOMADS HRRR GRIBs and indexes into R2

mrshll:
# Mirror NOMADS HRRR GRIBs and indexes into R2

First of two PRs split out of #1048 (Marsh, 2026-09-21: ship the mirror first and watch data land in the bucket, then the virtual product on top). This PR is the mirror alone; the product follows, stacked on this branch. Nothing here reads the mirror.

## What it does

One hourly single-pod cron, `noaa-hrrr-nomads-mirror-gribs` (`src/reformatters/noaa/hrrr/nomads_mirror.py`), registered beside the ECMWF archiver in `OPERATIONAL_ARCHIVERS`. It fires at :45, starts polling NOMADS at init + 49 min and stops one minute before the 59-minute pod deadline. For its init it copies the sfc/prs/nat f00–f18 GRIBs and NOAA's `.idx` files, 114 objects, into the R2 bucket `noaa-hrrr-nomads-mirror` under their NODD keys (`hrrr.YYYYMMDD/conus/hrrr.tHHz.wrfsfcf00.grib2`), through R2's signed S3 API with the secret `noaa-hrrr-nomads-mirror-storage-options-key`.

- **Copies run in parallel.** Each GRIB and its index copy as one task, and up to 4 tasks run at once while the job keeps polling. A pod slot freed by a finished copy is refilled from the latest listing, and no pair is tried twice per listing. Every listing and download shares the NOMADS rate limiter (100 requests per minute), and polls are 20 s apart. The pod's ephemeral storage is 8G, enough for four GRIBs on disk at once.

- A GRIB is copied only once it parses as whole GRIB2 messages that tile the file. Its task fetches the index first, then copies the GRIB, and writes the index only once it lists exactly the messages of the mirrored GRIB, re-copying the GRIB when they disagree. So an index in the mirror always describes the GRIB beside it. An empty or partly written index is retried on the next poll.
- A replaced or retried pod lists the bucket first and copies only what is missing.
- **Catch-up** (added 2026-09-21 at Marsh's request, as the ECMWF archiver and the DWD copy already do): once its own init is done, each fire walks the previous 24 inits, newest first, and copies whatever the mirror lacks that NOMADS lists, one NOMADS listing per incomplete init. It is best-effort and uses spare time only: it runs after the current init so it never delays fresh data, and it stops starting transfers 5 minutes before that polling deadline (6 before the pod's) so the copies in flight can finish. 24 is a bound on work, not NOMADS' retention (about two days); an outage longer than that loses its oldest hours even while NOMADS still has them. It trusts pairs already in the mirror, so it does not repair the truncated-pair case below.
- **The TTL is the bucket's lifecycle rule, not code.** Marsh set `expire-4d` and a 1-day multipart abort on 2026-09-17.

Shared code changed or first used here: `download.s3_store` takes an optional `endpoint` and credentials for a signed store on an S3-compatible service; `download.httpx_get` (the retrying httpx GET that `httpx_download_to_disk` uses, now public); `NoaaHrrrSourceFileCoord.relative_path()`, which is also the mirror key (`get_url` delegates to it), and `common/grib.py`'s `grib_message_offsets` and `grib2_message_length` (the latter also used by `NoaaVirtualRegionJob.grib_message_length_at`, replacing its own section 0 parse). NOMADS access reuses `noaa_utils.nomads_rate_limiter` and `NOMADS_RETRY_STATUS_CODES` with `httpx_download_to_disk`, as the HRRR, GFS and GEFS NOMADS fallbacks do.

## Known limits to watch during the mirror-only period

- **Green is not complete.** At its deadline the job starts no new transfer, waits for the copies in flight, logs a warning listing the files it did not mirror and exits 0, so the Sentry monitor stays green with files missing. Later fires retry them for 24 hours, but judge completeness from the bucket, not the monitor. A retried pod's "Mirrored N files" counts only what that pod copied.
- **A transfer started just before the deadline can outlive the pod.** The current init stops one minute before the pod's deadline, but one download can retry for up to 5 minutes. If Kubernetes kills the pod mid-copy, the fire ends as failed and the next fire's listing retries the file. The serial version had the same exposure, and the margin is unchanged.
- **Catch-up can be starved.** It has at most about 10 minutes per fire (f18 lands near init + 87 min; catch-up stops at init + 98 min). How much that recovers is unmeasured. A single connection moved a 150 MB file in 5 to 10 s in a smoke test before copies were parallel. A full init is tens of GB, and four copies now run at once, but parallel throughput has not been measured. If the current init is still missing any file, it polls to the deadline and that fire does no catch-up at all; if that happened every hour, nothing earlier would ever be recovered. When the walk runs out of time the log says `Deadline reached with N earlier inits not checked`; watch for that line during rollout.
- **The first fires after deploy fill an empty bucket.** Catch-up sees 24 missing inits and copies what it can per fire, newest first, four at a time, through the shared NOMADS rate limiter. The bucket therefore fills faster than one init per hour, and each of those fires logs one `114 files of <init> not mirrored` warning per incomplete init it checked; the key lists are at debug level.
- **R2 listing cost.** Each init checked lists its whole day prefix (up to 2,736 objects, about 3 list requests), so a healthy hour makes roughly 75 list requests, about 54,000 a month. Not optimised: object_store prefixes match whole path segments, so an hour-scoped prefix is not available without restructuring keys.
- **A truncated GRIB whose truncated index agrees with it is accepted and never revisited.** The product PR's completeness guard skips such a file, but nothing alerts on it while only the mirror runs; the check script below compares against NOMADS while NOMADS still has the file.

## Needs Marsh

Already done by Marsh: bucket, secret, lifecycle rules, public custom domain `noaa-hrrr-nomads-mirror.r2.dynamical.org` (answers 404 for a missing key today, as expected for an empty bucket).

1. **Merge = deploy** of the cron. Nothing else deploys with it.
2. **Watch it land.** After the deploy finishes and one full fire has run (fire at HH:45 mirrors init HH; by about HH+1:44 it is done), pick that init and run, from a checkout: `uv run python plans/1048_split_scratch/check_mirror_init.py 2026-09-22T15` (script below; it is not committed). While NOMADS still has the init (about two days), it checks that every mirrored GRIB's size and every mirrored index's content equal NOMADS', which is what catches a truncated pair, then decodes one message per file type from the mirror through an index-derived HTTP range. Expect `0 problems of 114 objects`. Run it again for the next init to see the mirror advance. The Cloudflare dashboard's R2 object browser shows the same objects with sizes and upload times; the public domain has no listing.
3. **TTL.** Note a few objects' upload times; about 4 days later they should be gone from the dashboard (lifecycle deletion is not exact to the hour, and the public domain may serve a cached copy a little longer).

<details><summary>check_mirror_init.py</summary>

```python
"""Usage: uv run python check_mirror_init.py 2026-09-21T14
For one init, while NOMADS still has it: every mirrored GRIB's size and every mirrored
index's content must equal NOMADS'; then decode one message per file type from the
mirror through an index-derived HTTP range."""
import sys

import gribberish
import httpx
import pandas as pd

MIRROR = "https://noaa-hrrr-nomads-mirror.r2.dynamical.org/"
NOMADS = "https://nomads.ncep.noaa.gov/pub/data/nccf/com/hrrr/prod/"
if len(sys.argv) > 2:
    MIRROR = sys.argv[2]

init = pd.Timestamp(sys.argv[1])
client = httpx.Client(timeout=60, follow_redirects=True)
problems = []
for file_type in ("sfc", "prs", "nat"):
    for lead in range(19):
        key = f"hrrr.{init:%Y%m%d}/conus/hrrr.t{init:%H}z.wrf{file_type}f{lead:02d}.grib2"
        mirrored, source = client.head(MIRROR + key), client.head(NOMADS + key)
        if mirrored.status_code != 200:
            problems.append(f"{key}: mirror {mirrored.status_code}")
        elif mirrored.headers["content-length"] != source.headers["content-length"]:
            problems.append(f"{key}: size differs from NOMADS")
        mirrored, source = client.get(MIRROR + key + ".idx"), client.get(NOMADS + key + ".idx")
        if mirrored.status_code != 200:
            problems.append(f"{key}.idx: mirror {mirrored.status_code}")
        elif mirrored.text != source.raise_for_status().text:
            problems.append(f"{key}.idx: differs from NOMADS")
print(f"{len(problems)} problems of 114 objects: {problems}")
for file_type in ("sfc", "prs", "nat"):
    key = f"hrrr.{init:%Y%m%d}/conus/hrrr.t{init:%H}z.wrf{file_type}f18.grib2"
    lines = client.get(MIRROR + key + ".idx").raise_for_status().text.splitlines()
    starts = [int(line.split(":")[1]) for line in lines]
    i = len(starts) // 2
    message = client.get(
        MIRROR + key, headers={"Range": f"bytes={starts[i]}-{starts[i + 1] - 1}"}
    ).raise_for_status().content
    values = gribberish.parse_grib_array(message, 0)
    print(file_type, "f18", lines[i].split(":")[3:5], values.shape, float(values.mean()))
```
</details>

## Verified here

- Full suite on this branch at 9ba9e55a: 3474 passed, 13 skipped. `ruff format`, `ruff check`, `ty check` clean.
- Rendered manifest: name `noaa-hrrr-nomads-mirror-gribs`, schedule `45 * * * *`, `concurrencyPolicy: Replace`, not suspended, pod deadline 3540 s, command `noaa-hrrr-nomads mirror-gribs`, secret volume mounted at `/secrets/noaa-hrrr-nomads-mirror-storage-options-key.json`. No product crons; no tracked file on this branch mentions the fast product.
- Source-to-local smoke test on 2026-09-21, before copies were parallel: `mirror_init_time` copied the 14Z init's sfc f00 and f18 GRIBs and indexes (150 MB, 160 MB) from NOMADS into a temporary `LocalStore` with nothing pending. This exercises the NOMADS listing, the whole-GRIB check and the index agreement check; it does not exercise R2 credentials or a full init's throughput inside the deadline.
- Catch-up smoke test, same day and also before the parallel copies: `mirror_earlier_init_times` with two inits back copied the 14Z and 13Z sfc f00 pairs from NOMADS into a `LocalStore`; a second call copied nothing and made no NOMADS request.
- Focused tests for the parallel copies each fail with one worker or with the old wait timeout. They cover two downloads in flight at once, polling that continues while a copy is in flight, one listing draining more pairs than the pool runs, a deadline that passes between a pair's index and GRIB fetches, a failed copy that admits no further work, and no busy wait once every pending pair is in flight.
- The check script above, pointed at NOMADS as the mirror, reports 0 problems and three decodes; pointed at the empty bucket it reports 114 problems.

Not verified, and only observable after deploy: R2 writes with the mounted secret, a full 114-object init inside the 59-minute deadline, lifecycle expiry.

## Note

The `generate-manual-workflows` prek hook's `files` pattern does not match `nomads_mirror.py`, so a future change to this cron's name would not regenerate the workflow choices by itself. The choice in this PR was generated by running the script. Not changed here.

### 2026-09-22: simplification pass

- [x] Dropped the `docs/ops_card.md` change (restored to main).
- [x] Inventoried existing NOMADS code for reuse. The mirror already used `nomads_rate_limiter`, `NOMADS_RETRY_STATUS_CODES`, `httpx_download_to_disk` and `NoaaHrrrSourceFileCoord.get_url(source="nomads")`; it now also uses `get_idx_url(source="nomads")` instead of appending `.idx` itself. Nothing else in the repo lists a NOMADS directory or copies whole NOMADS files.
- [x] Moved GRIB2 section 0 parsing into `common/grib.py` and shared it between the mirror and `NoaaVirtualRegionJob`.
- [x] Removed `parse_mirror_key` and `is_whole_grib2`, which nothing used, and `_index_matches`.
- Not reused, and why:
  - `ecmwf/archive_gribs/grib_inventory.iter_grib_messages` asserts on a malformed file, but the mirror needs a "not whole yet" answer while it polls.
  - `grib_decimal_scale_factors` scans for magic bytes and tolerates padding, so its contract is different.
  - The ECMWF and DWD archivers copy with rclone; the mirror checks each file before it writes with obstore.
- Split out as #1106: a shared `noaa_utils.nomads_download_to_disk` for the HRRR, GFS and GEFS NOMADS fallbacks. This mirror's `download_from_nomads` can switch to it once both PRs are on main.
- Codex review, pass 1:
  - Fixed: a partly written index made `parse_grib_index_lines` raise and end the whole fire. It is now retried, with a test.
  - Fixed: mirror membership is a set; the recorded sizes were never read.
  - Fixed: the deadline wording above.
  - Not taken: moving the product branch's `read_data_bytes` and `download_index` hooks into this PR. `feat/hrrr-nomads-product` will conflict in `grib_message_length_at`; resolve it as `header = self.read_data_bytes(coord, offset, offset + GRIB_SECTION_0_BYTES)` followed by `return grib2_message_length(header)`.

### 2026-09-22: Alden's review

- [x] 4074474160: `signed_s3_store` is folded into `s3_store`, which takes `endpoint`, `access_key_id`, `secret_access_key` and `virtual_hosted_style_request`. `mirror_store` maps the secret's icechunk options onto them, and the signed store gains the shared client options and retry config.
- [x] 4074486708: `_httpx_get_with_retry` is public as `httpx_get`, and `httpx_get_text` is gone. The NOMADS listing calls `httpx_get(...).text`.
- [x] 4074873664: `mirror_key` is removed; keys are `coord.relative_path()`. `feat/hrrr-nomads-product` must follow when it re-merges.
- [x] 4074890805: `MirrorResult` and the per-pair records are `FrozenBaseModel`s built by keyword.
- [x] 4074962866: copies run in parallel pairs while polling continues, as described above.
- Codex review of the rewrite, pass 1:
  - Fixed: work queued in the pool could start after the deadline. Admission is now bounded to the pool size, with a deadline check before every fetch in a pair.
  - Fixed: a failed copy let queued copies keep running. It now admits no further work.
  - Fixed: stale description claims.
  - Not taken: widening the one-minute margin before the pod deadline. See Known limits.
- Codex review, pass 2 (the last allowed for this change):
  - Fixed: the polling thread spun once every pending pair was in flight.
  - Fixed: added a test that one listing drains more pairs than the pool runs.
  - These last two fixes were not reviewed again.


src/reformatters/common/download.py
```diff
@@ -134,6 +134,24 @@ def gcs_store(bucket_url: str) -> obstore.store.GCSStore:
     return store
 
 
+def signed_s3_store(
```
aldenks: rather than add a new method, lets unify this with the existing s3_store.

src/reformatters/common/download.py
```diff
@@ -385,6 +403,18 @@ def httpx_download_to_disk(
     return local_path
 
 
+def httpx_get_text(
```
aldenks: doesn't need to be its own method. rather rename _httpx_get_with_retry -> httpx_get and use that at the current callers of this method

src/reformatters/noaa/hrrr/nomads_mirror.py
```diff
@@ -0,0 +1,378 @@
...
+# NOMADS keeps about two days of inits.
+CATCH_UP_INIT_TIMES: Final = 24
+# Catch-up stops starting copies this long before the fire's deadline, so one in
+# flight can finish.
+CATCH_UP_DEADLINE_MARGIN: Final = timedelta(minutes=5)
+
+
+def mirror_key(coord: NoaaHrrrSourceFileCoord) -> str:
+    """The mirror object key for coord's data file: identical to its NODD key."""
+    return coord.relative_path()
```
aldenks: is this method needed?

src/reformatters/noaa/hrrr/nomads_mirror.py
```diff
@@ -0,0 +1,378 @@
...
+        MIRROR_BUCKET, kubernetes.load_secret(MIRROR_SECRET_NAME)
+    )
+
+
+class MirrorResult(NamedTuple):
+    copied: list[str]
+    pending: list[str]
+
+
+class _MirrorFile(NamedTuple):
```
aldenks: prefer pydantic basemodel and kwarg based initialization

src/reformatters/noaa/hrrr/nomads_mirror.py
```diff
@@ -0,0 +1,378 @@
...
+    ):
+        polls += 1
+        poll_start = time.monotonic()
+        listed = list_directory(directory_url)
+        for file in [f for f in pending if f.name in listed]:
+            if pd.Timestamp.now("UTC") > deadline:
+                break
+            data_key = mirror_key(file.coord)
+            if file.is_index and data_key not in in_mirror:
+                continue
```
aldenks: I think this only starts copying the index _after_ the data file has been copied. I think we probably want to restructure this loop so that the downloads are 
- not blocking polls or other downloads 
- not serial

the rate limiters job is to do a shared rate limit, but we can/should download multiple files in parallel.

src/reformatters/common/download.py
mrshll: Done in cf16d214. `signed_s3_store` is gone, and `s3_store` takes optional `endpoint`, `access_key_id`, `secret_access_key` and `virtual_hosted_style_request` (still `functools.cache`d, and anonymous callers are unchanged). `mirror_store` maps the secret's icechunk options onto them, including `force_path_style` to path-style addressing, so the signed store also gets the shared client options and retry config.

src/reformatters/common/download.py
mrshll: Done in cf16d214. `_httpx_get_with_retry` is now the public `httpx_get`, `httpx_get_text` is removed, and the NOMADS listing calls `httpx_get(...).text`.

src/reformatters/noaa/hrrr/nomads_mirror.py
mrshll: No, removed in cf16d214. Keys are `coord.relative_path()` directly. The stacked product branch uses it too; I've told its owner to switch to `relative_path()` when it re-merges.

src/reformatters/noaa/hrrr/nomads_mirror.py
mrshll: Done in cf16d214. `MirrorResult` and the per-pair records (`_PairCopy`, `_PairResult`) are `FrozenBaseModel`s, built with keyword arguments.

src/reformatters/noaa/hrrr/nomads_mirror.py
mrshll: Restructured in cf16d214, then tightened in 127587c0 and 9ba9e55a.

- **Pairs:** each GRIB and its index copy as one task. The task fetches the index first, copies the GRIB, and writes the index only once its message offsets match the mirrored GRIB, re-copying the GRIB on disagreement. So the index no longer waits for the GRIB in a separate step, and an index in the mirror always describes the GRIB beside it.
- **Parallel, non-blocking:** up to 4 tasks run at once on a thread pool while the polling thread keeps listing every 20 s. A freed slot is refilled from the latest listing, trying each pair at most once per listing. Everything still goes through the shared `nomads_rate_limiter`. I raised the poll interval from 1 s because listings now share that 100-per-minute budget with parallel downloads.
- **Bounds:** admission is capped at the pool size, so nothing queues past the deadline. Every fetch in a task checks the deadline first. A failed copy admits no further work before its error surfaces. The polling thread waits on copies without spinning once every pending pair is in flight.
- **Resources:** the pod's ephemeral storage is 8G, enough for four GRIBs on disk.
- **Tests:** they show two downloads in flight at once, polling that continues during a slow download, and one listing draining more pairs than the pool runs. Each fails with a single worker.

src/reformatters/__main__.py
```diff
@@ -400,7 +401,10 @@ def startup(ctx: typer.Context) -> None:
 
 # Source archives that feed a dataset but have no store of their own. They deploy
 # their own cronjobs and are not datasets, so they carry no update/validate/backfill.
-OPERATIONAL_ARCHIVERS: Sequence[OperationalResources] = [EcmwfIfsEns46DayGribArchiver()]
+OPERATIONAL_ARCHIVERS: Sequence[OperationalResources] = [
+    EcmwfIfsEns46DayGribArchiver(),
+    NoaaHrrrNomadsMirror(),
+]
```
aldenks: doesn't need to be for this PR, but why are these different than a command on a DynamicalDataset subclass?  If they are 1:1 to a concrete dynamical dataset they just add a cli command to it (e.g. what icon-eu does), or if multiple use a single archive process we make it a superclass of those and its update/validate/backfill commands raise NotImplementedError. There's no requirement that operational kubernetes resources returns exactly 2 cronjobs, it can return 1 (an archiver) or 3 (archive, update, validate)

the code just below also begs the question of why its a new kind of thing:
```
for dataset in DYNAMICAL_DATASETS:
    app.add_typer(dataset.get_cli(), name=dataset.dataset_id)

for archiver in OPERATIONAL_ARCHIVERS:
    app.add_typer(archiver.get_cli(), name=archiver.dataset_id)

deploy_module.register_commands(app, DYNAMICAL_DATASETS, OPERATIONAL_ARCHIVERS)
```

mrshll: good call - this is a relic of shipping the copiers before the data products existed. will make in a new PR

PR #1108, opened 2026-09-23
Defer a listed NOMADS file it does not serve, and stream httpx downloads

mrshll:
The first production fire of the NOMADS HRRR mirror (#1095) needed 5 pods and 4 crashes to finish the 17Z init. Each crash was an uncaught 404 on a file the NOMADS listing had just shown, usually right after "peer closed connection" truncations. Peak pod memory was 5.2 to 7.9 GiB against a 2G request, because full-object httpx downloads held each whole body in memory.

## Changes
- **`nomads_mirror._fetch_if_served`.** When NOMADS lists a file but returns 404 for it, or every retry of the transfer fails, the pair stays pending. It is tried again after the next listing. Other statuses, R2 errors and bugs still raise.
- **`nomads_mirror.mirror_fire`.** This is the fire body the CLI runs. If any of the fire's own init is still not mirrored at the deadline, the fire now raises once catch-up has finished, so the cron monitor goes red. Without this, deferring could hide a persistent outage behind green check-ins. Normally the init completes about 20 minutes before the deadline (init+85 min against init+103 min). Earlier inits left incomplete by catch-up do not fail the fire.
- **`common/download.py`.** `httpx_get` and the full-object branch of `httpx_download_to_disk` now share one retry loop over `client.stream`. The body streams to disk inside that loop, so a truncated body is retried as a whole request and memory stays bounded. Byte-range downloads behave as before.

## Tests
- A listed data file or index that 404s, or keeps getting cut off, is copied on a later poll. There are four parametrized cases, all red without the source change. A 403 still fails the fire.
- A fire whose init is incomplete at the deadline raises after catch-up; an incomplete earlier init does not. A file that fails with `ConnectError` on every poll ends in a failed fire.
- A body cut off mid-stream is retried. A 256 MB streamed body peaks at under 16 MB of traced Python memory (red before this change). A 404 is raised after one request and leaves no temp file.
- The existing `httpx_get` tests now use `httpx.MockTransport`.
- `uv run ruff format && uv run ruff check && uv run ty check` pass. The full suite passes: 3514 passed, 13 skipped.

## Not in this PR
- The CronJob's `memory="2G"` request is unchanged. Re-measure peak memory after this lands.
- #1110 suspends the CronJob separately. This PR does not touch `suspend`.
- The mirror's deadline only gates when copies start; it does not bound the whole run.
- A consistent partial GRIB and index pair can still become permanent in the mirror.

These are tracked in #1109.

Merging deploys the change through Deploy Operational Updates. No cluster changes are needed.


mrshll:
Closing unmerged, as Marsh decided. The HRRR NOMADS mirror is moving to the Cloudflare Worker in dynamical-org/nomads-mirror, so the EKS mirror's fixes here (deferring listed-but-unserved files, failing a fire whose init is incomplete, streaming full-object downloads) won't be needed. The EKS cron stays suspended (#1110). The shared streaming change only materially helped the EKS mirror; the GFS, GEFS and HRRR NOMADS fallbacks download GRIBs through the byte-range path, which this PR didn't change. The branch fix/nomads-mirror-fetch-deferral is kept.

PR #1109, opened 2026-09-23
Evaluate running the NOMADS mirror copy on Modal (prototype)

mrshll:
> **Superseded on 2026-09-23.** Marsh chose a Cloudflare Worker in a separate repository for the NOMADS mirror. This PR stays open, unmerged, as the record of the Modal alternative and its research. Its review was cancelled before any findings, and no Modal work continues here. Do not merge.

**Do not merge.** This is the evaluation and prototype for moving the NOMADS HRRR mirror's hourly copy off AWS, where internet egress to R2 costs about $1.6k/month. It is stacked on #1108, which is the independently landable reliability hotfix. #1110 suspends the EKS cron separately; this branch does not touch `suspend`.

## Recommendation (provisional)

Run reformatters on Modal through one job-specific module. That is option 1. A separate library (option 2) is deferred: it moves the same bytes, so it costs the same egress, and it needs a larger migration.

Modal is not transfer-free. From **2026-10-01, Modal charges $0.04/GiB egress**, and uploads to external object stores count ([docs](https://modal.com/docs/guide/network-egress-billing)). The allowance is workspace-wide: 1 TiB on Starter, 10 TiB on Team.

Modeled cost per 30 days, at 25.891 GB per init and 720 inits:

| Path | Monthly cost |
|---|---|
| EKS today | ≈ $1,670 |
| Modal, marginal | ≈ $815 (egress $694 + us-east compute $76 + R2 $45) |
| Modal, if the Team allowance is unused | ≈ $405 |

The fallback is Cloudflare Workers, whose pricing has no bandwidth charge; that could run at roughly $50–60. It is a feasibility scenario only, and whether to pursue it is a scope decision.

## Option 1: what this branch adds
- **`nomads_mirror_modal.py`.** It declares `app`, one hourly Function (`49 * * * *`, us-east, 1 core, 2 GiB, `max_containers=1`, 58 min timeout, Sentry cron check-ins), and a separate `pilot_app` with its own secret. The image is `debian_slim(3.14).uv_sync(--no-dev)` plus the local `reformatters` source.
- **`nomads_mirror.py`.**
  - `mirror_store(options, bucket)` accepts explicit options.
  - `mirror_pilot` copies a chosen set of files from one init, in one listing, into a bucket other than the production mirror.
  - `mirror_fire` is shared with the Kubernetes CLI and comes from #1108.
- **Dependencies.** `modal` is a dev dependency only, so the EKS image does not change.
- **Unchanged.** `OperationalResources`, the dataset registry, `kubernetes.py`, the Dockerfile and the deploy workflows.
- **Run window.** A run works on the latest scheduled fire's init. That is safe because runs are serial and copies are idempotent.
- **Tests.** The pilot's local argument checks, the window and the app wiring run in a subprocess. pytest's `pythonpath=src/reformatters` makes `reformatters.google` shadow the `google` namespace that `modal` imports.

## Option 2: boundary assessment, not built
- **What it would move.** The mirror's GRIB and idx logic is shared with the virtual region jobs: `grib2_message_length` and the full `parse_grib_index_lines`. Its downloads and rate limiting are shared with every NOMADS fallback. The consumer imports `mirror_store` and `MIRROR_SECRET_NAME`.
- **What it would gain.** An image of about 60 MB instead of 1.1 GB, and an import of about 0.3 s instead of 4.8 s.
- **Packaging.** `Image.uv_sync` does not support uv workspaces.
- **Existing tools.** None of rclone, s5cmd, obstore or fsspec checks GRIB completeness or idx agreement.

## Before production (gates)
- **Partial pairs.** A consistent partial GRIB and idx pair can become permanent. We need a finality or revalidation rule; this gap predates this work.
- **Monitoring.** Keep the #1108 contract: a fire fails when its own init is incomplete at the deadline.
- **Evidence.** Modal throughput, peak memory and billed egress, measured by the pilot.
- **Cutover.**
  1. #1110 suspend is applied.
  2. Confirm the EKS Jobs and pods have drained.
  3. `modal deploy …::app`.
- **Rollback.**
  1. `modal app stop noaa-hrrr-nomads-mirror`.
  2. Confirm no containers or queued inputs remain.
  3. Land `suspend=False`.
- **Deploy.** A Modal step in CI runs only from main, with its own credentials. Removing the schedule must persist in code.

## Pilot (not run; needs Marsh)
The pilot reads one published init with one listing, has no catch-up, runs as one non-preemptible container, refuses the production bucket, and validates its inputs locally. The exact commands and budget are in the evaluation report.


mrshll:
Closing unmerged. The NOMADS mirror is moving to a Cloudflare Worker in a separate repository, which supersedes this Modal adapter. Modal's egress pricing ($0.04/GiB from 2026-10-01) is part of why. The branch `explore/nomads-modal`, its commits and the research in the description stay as the record of the Modal alternative. #1108, the reliability hotfix, is independent and unaffected.

